PR #133, opened 2013-12-04
Reset the database connection after MySQL bulk insert failure

### LINKED ISSUE #93: Database connectivity problems with MySQL some databases
On Ubuntu 13.04 (MySQL 5.5.31) and Windows 7 (MySQL 5.6) databases with multiple tables are failing to import. For BBS on Windows the Retriever just hangs saying "Connecting to Database".

Error messages on Linux are of the form:

```
Failed bulk insert ((2013, 'Lost connection to MySQL server during query')), inserting manually
INSERT INTO Gentry.sites (site_id, site_name, country, state_province, continent, abbreviation, lat, lon, min_elev, max_elev, precip) VALUES (1, 'Kitlope 1', 'Canada', 'British Columbia', 'North America', 'KITLOPE1', 53.07, -127.83, 20, 20, 2100);
(2013, 'Lost connection to MySQL server during query')
```

On Windows if the Retriever doesn't hang an error message like this occurs:

`error 10053 An established connection was aborted by the software in your host machine.`

ethanwhite:
After a bulk insert failure in MySQL (caused by servers not
allowing this by default) we fall back to a line by line insert.
However, using pymysql the failed insert causes the database
connection to get hung up.

This commit creates a new method to reset the connection and
calls it prior to starting the line by line insert.

Fixes #93.

### DIFF
--- a/engines/mysql.py
+++ b/engines/mysql.py
@@ -66,7 +66,7 @@ IGNORE """ + str(self.table.header_rows) + """ LINES
                 self.cursor.execute(statement)
             except Exception as e:
                 print "Failed bulk insert (%s), inserting manually" % e
-                self.reset_connection() # If the execute fails the database connection can get hung up
+                self.disconnect() # If the execute fails the database connection can get hung up
                 return Engine.insert_data_from_file(self, filename)
         else:
             return Engine.insert_data_from_file(self, filename)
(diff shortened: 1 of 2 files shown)


hkhatod:
So how do I fix again ? are are any speciifc steps I need to take in order to solve this?

ethanwhite:
This commit should have prevented the issue from happening. If you're seeing a related problem please open a new issue with some details and we'll be happy to take a look.

PR #227, opened 2014-10-16
Adding EA_Dell.script file

carol-rowe666:
Add database of 2352 thermal responses for 220 traits for microbes, plants, and animals compiled from 270 published sources. File from: http://esapubs.org/archive/ecol/E094/108/


ethanwhite:
This appears to be a line endings issue that for some reason isn't being properly handled by Python. If I run `dos2unix` on the raw data file (in `.retriever/raw_data/...`) everything works perfectly.

ethanwhite:
@carol-rowe666 we finally figure out what was going on with this data and will be merging your contribution in shortly via #479. Thanks again for the help and sorry it took so long to get this in.

Replaced by #479.

Issue #254, opened 2015-01-20
add eBird Observational Dataset


brymz:
This conversation started with Issue #90 (github.com/weecology/retriever/issues/90) and continues with the 'add eBird Reference Datasets' (Issue #252) and 'add eBird Stratified Random Design Data' (Issue #253) 

The eBird Observational Dataset is located for public access at  https://dataone.ornith.cornell.edu/metacat/d1/mn/v1/object/CLOEODDATA.05192014.1

The data downloads as CLOEODDATA.05192014.1.gz, though the website suggests it could also be named eBird_Observation_Dataset_2013.csv.gz.

I don't remember if we checked a line count for this one or the Basic Dataset, but my counter has it at 150,000,000.

The table format seems to be ready for the retriever. Column names are: 
BASISOFRECORD, INSTITUTIONCODE, COLLECTIONCODE, CATALOGNUMBER, OCCURRENCEID, RECORDEDBY, YEAR, MONTH, DAY, COUNTRY, STATEPROVINCE, COUNTY, DECIMALLATITUDE, DECIMALLONGITUDE, LOCALITY, KINGDOM, PHYLUM, CLASS, ORDER, FAMILY, GENUS, SPECIFICEPITHET, SCIENTIFICNAME, VERNACULARNAME, INDIVIDUALCOUNT

ethanwhite:
This is in progress in #286.

ethanwhite:
Add in #286.

PR #255, opened 2015-01-20
basic retreiver script for the eBird Observational Dataset at DataONE

brymz:
This is a basic retriever script for the eBird Observational Dataset in response to Issue #254. 

I think the basic script will be sufficient. Though, I am not sure the criteria for changing a column name and have not gone through the large file to see if there are any odd NULL values. 

Metadata is available at: https://cn.dataone.org/cn/v1/resolve/CLOEODMETADATA.05192014.1

### DIFF
--- a/scripts/eBirdObservation.script
+++ b/scripts/eBirdObservation.script
@@ -1,0 +1,8 @@
+# basic information about the script
+name: eBird Observational Dataset
+shortname: eBirdOD
+description:  2014. EOD - eBird Observation Dataset. CLOEODMETADATA.05192014.1
+tags: Taxon > Birds, Spatial Scale > Global, Data Type > Observational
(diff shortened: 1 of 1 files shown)


akshayah3:
@ethanwhite @brymz What else needs to be done in this PR?

ethanwhite:
Hi @akshayah3 - this PR isn't actually the right way to go. @brymz and I talked about this in person and we forgot to update the issue. I have a draft of the script in:
https://github.com/ethanwhite/retriever/blob/ebird-obs/scripts/eBird_observation.py

It needs to be tested and then any issues fixed. I'll just submitted it as a work in progress PR (#285) so that we can discuss it there, and I'm going to close this issue to avoid further confusion

PR #259, opened 2015-01-27
Add reset command

### LINKED ISSUE #258: Add --reset option to remove hidden files
We cache data and scripts in ~/.retriever. This is great most of the time, but sometimes you want to get rid of those files. We should add a:

`retriever --reset`

option to automatically get rid of this directory.

ethanwhite:
Adds a reset command that removes the database configuration settings,
scripts, and cached raw_data stored in HOME_DIR.

Fixes #258.

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -10,12 +10,13 @@ The main() function can be used for bootstrapping.
 import os
 import platform
 import sys
+import shutil
 # sys removes the setdefaultencoding method at startup; reload to get it back
 reload(sys)
 if hasattr(sys, 'setdefaultencoding'):
     # set default encoding to latin-1 to avoid ascii encoding issues
     sys.setdefaultencoding('latin-1')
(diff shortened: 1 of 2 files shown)


bendmorris:
Steam recently screwed this up very badly (https://github.com/ValveSoftware/steam-for-linux/issues/3671) so I'm trying to think about edge cases here which might result in accidentally deleted files.

Would it make sense to be explicit about what's deleted to avoid deleting any extra files that happen to be there? e.g. just the scripts and raw_data directories and connection config files? Or am I worrying about nothing? I think in the case of symlinks shutil.rmtree will throw an exception, so we should be good there.

ethanwhite:
I totally agree that an abundance of caution is good here. I'd be happy to just remove the scripts and raw_data directories as well as the connections file if you think that would add enough safety.

ethanwhite:
I just updated the PR with those protections and would be happy add any others that you think would make things safer.

PR #261, opened 2015-01-29
Updates Baldridge2012 script.

### LINKED ISSUE #260: Baldridge 2012 citations import
Title field of the citations table sometimes splits a single record into multiple records.

embaldridge:
Replaces citations table link with link to updated citations table, fixes #260 .


ethanwhite:
Thanks Elita! Nothing like having someone discover a dataset bug, fix the source of the bug, and submit a PR updating the script! That's an awesome contributor right there!

Issue #268, opened 2015-03-05
Error while installing Retriever on Linux when using a proxy


prabh27:
Error is as follows: 
Traceback (most recent call last):
  File "/usr/bin/retriever", line 9, in <module>
    load_entry_point('retriever==1.7.0', 'console_scripts', 'retriever')()
  File "/usr/lib/python2.7/dist-packages/retriever/__main__.py", line 31, in main
    check_for_updates(graphical=False if 'darwin' in platform.platform().lower() else True)
  File "/usr/lib/python2.7/dist-packages/retriever/lib/repository.py", line 74, in check_for_updates
    init.run()
  File "/usr/lib/python2.7/dist-packages/retriever/lib/repository.py", line 143, in run
    version_file = urllib.urlopen(REPOSITORY + "version.txt")
  File "/usr/lib/python2.7/urllib.py", line 87, in urlopen
    return opener.open(url)
  File "/usr/lib/python2.7/urllib.py", line 208, in open
    return getattr(self, name)(url)
  File "/usr/lib/python2.7/urllib.py", line 437, in open_https
    h.endheaders(data)
  File "/usr/lib/python2.7/httplib.py", line 969, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python2.7/httplib.py", line 829, in _send_output
    self.send(msg)
  File "/usr/lib/python2.7/httplib.py", line 791, in send
    self.connect()
  File "/usr/lib/python2.7/httplib.py", line 1176, in connect
    self.sock = ssl.wrap_socket(sock, self.key_file, self.cert_file)
  File "/usr/lib/python2.7/ssl.py", line 487, in wrap_socket
    ciphers=ciphers)
  File "/usr/lib/python2.7/ssl.py", line 243, in __init__
    self.do_handshake()
  File "/usr/lib/python2.7/ssl.py", line 405, in do_handshake
    self._sslobj.do_handshake()
IOError: [Errno socket error] [Errno 1] _ssl.c:510: error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol

ethanwhite:
Thanks for the bug report @prabh27. I can't replicate this at the moment. Can you give me some more information on your system configuration including OS and Python versions and the version of openssl since that appears to be what's throwing the error?

prabh27:
System Configuration: Linux 14.04
Python Version : 2.7.6
OpenSSL version : OpenSSL 1.0.1f 6 Jan 2014

[...]

PR #270, opened 2015-03-06
Update repository.py

prabh27:
Atleast one of the proxies would be set on the system. Check that proxy and set all the other proxies so that more proxy issues occur.

### DIFF
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -11,6 +11,17 @@ from threading import Thread
 from retriever import REPOSITORY, VERSION, MASTER_BRANCH, REPO_URL, SCRIPT_WRITE_PATH, HOME_DIR
 from retriever.lib.models import file_exists
 
+proxy = ["http_proxy","https_proxy","ftp_proxy","HTTP_PROXY","HTTPS_PROXY","FTP_PROXY"]
+
+flag = 0
+
+for i in range(len(proxy)):
+    if flag == 0:
+        if len(os.environ[proxy[i]]) != 0:
+            flag = 1
+            for j in range(len(proxy)):
+                os.environ[proxy[j]] = os.environ[proxy[i]]
+
 global abort, executable_name
 abort = False
 executable_name = "retriever"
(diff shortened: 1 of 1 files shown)


ethanwhite:
Thanks for the PR's @prabh27. I have a couple of general suggestions to get started and we can go forward from there:

1. Sets of changes that address the same issue are best submitted as a single PR, so the changes in #270 and #272 should be in a single PR. The PR title and description and associated commit messages should clearly state what the changes are accomplishing and why they were needed. For example, in this case I'd use a PR title like "Check for and use system proxies for downloading files" and an accompanying description like "In some cases when the user is using a proxy urlib.urlopen() will fail to successfully open https files. This prevents the retriever from accessing the scripts stored on GitHub and causes the installation to fail (see #268). This change checks for the existence of proxies and makes them available in a way that urllib.urlopen() can find them". This title and description would also be good examples of a first line of a commit message and a following paragraph with more detailed description. Being explicit like this will make it easier for future developers (or even you and I in two months when we don't remember this discussion as well) to understand what was done and why.
2. Since the changes in the two PRs are exact duplicates in different places, the code should either be moved somewhere were it can only be called once (like __init__.py), or placed in a function that can be imported and called in these two locations.
3. The code itself can be simplified and made more readable by:
    * Using Python's ability to loop over lists rather than indexes, `for proxy in possible_proxies:`
    * Using `os.getenv()` without the second argument. This will return `None` if the proxy doesn't exist, which means that lines 20 and 21 can be replaced with `if os.getenv(proxy[i]):` (or just `proxy` if you make the first change I'm recommending).
    * Using either `break` (if not in a function) or `return` (if in a function) at the end of the last `if` statement instead of using a flag
4. What is your feeling on the benefits of using `os.environ()` to set the environmental variables vs. identifying the proxy and then passing it to `urllib.urlopen()`?

prabh27:
@ethanwhite , Thanks for the reply. I have made the changes suggested by you. Regarding the 4th point, passing the proxy through urllib.urlopen() solved that problem but it got stuck in urllib.urlretrieve(). This was the reason why I used os.environ() and set all the proxy variables, so that no isssues of urllib occur.

ethanwhite:
Thanks for the clarification on the approach @prabh27. I'm going to go ahead and close this and #272 since we're now working on this in #277.

PR #273, opened 2015-03-08
Grant read/write/exec privileges to all users for retriever directories.

### LINKED ISSUE #206: Clean install from source leads to unwriteable .retriever directories
If you install the retriever from source and there are not an existing `~/.retriever/raw_data/` and `~/.retriever/scripts/` directories, these directories are created on install. Since the source install requires the use of `sudo` these directories are owned by `root` and when the retriever is run (typically using non-elevated privileges) it cannot write to those directories.

Removing the directories or changing their ownerships solves the problem, but this should be avoided.

bendmorris:
Fixes #206

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -27,6 +27,7 @@ for dir in (HOME_DIR, os.path.join(HOME_DIR, 'raw_data'), os.path.join(HOME_DIR,
     if not os.path.exists(dir):
         try:
             os.makedirs(dir)
+            os.chmod(dir, 0o777)
         except OSError:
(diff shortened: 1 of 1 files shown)


ethanwhite:
Two competing pull requests for the same issue. That's a new one for us :) At a first glance I like @lyttonhao's approach (#274) of explicitly changing ownership even though it's a bit more complicated. What do you think @bendmorris?

#274 needs some cleanup, but I thought we should agree on the right general approach before I recommend any changes to @lyttonhao.

ethanwhite:
Closing in favor of #274.

PR #274, opened 2015-03-08
Fix #206 change ownership of .retriever directories to the user who does the install

### LINKED ISSUE #206: Clean install from source leads to unwriteable .retriever directories

lyttonhao:
Fix #206. If it is in Unix platform and the commond is called by Sudo, change the ownership of .retriever.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -31,6 +31,7 @@ for dir in (HOME_DIR, os.path.join(HOME_DIR, 'raw_data'), os.path.join(HOME_DIR,
         try:
             os.makedirs(dir)
             if (not "win" in current_platform) and os.getenv("SUDO_USER"):
+                # change the ownership of dir directory to the user who is calling this sudo command 
                 pw = pwd.getpwnam( os.getenv("SUDO_USER") )
                 os.chown(dir, pw.pw_uid, pw.pw_gid)
         except OSError:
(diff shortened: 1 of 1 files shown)


ethanwhite:
Thanks for running this down @lyttonhao! I really like the general solution and have just a couple of suggestions to make things more readable:

* Move the import of pwd to the imports section at the top of the file
* We are already running platform.platform.lower() later in the file (Line 49), so let's move that line up so that we can use current_platform in the conditional instead of running platform.platform.lower() twice (mostly for readability)
* Lastly, I think for readability I'd move the call to `pwd.getpwname()` inside the `if` statement you added inside the loop. This means we call it 3 times instead of one, but it makes things more readable by keeping all of the sudo user logic in one place. It also means that we don't need to store `ifchown` so the single conditional can now become the more readable:

`if (current_platform != "win") and os.getenv("SUDO_USER"):`

lyttonhao:
Thanks for your detailed suggestions @ethanwhite. I have a question about the first point. Since pwd is not a module in windows, I think we need to put this import after the condition of current_platform.

ethanwhite:
Excellent point @lyttonhao, you're absolutely right, and nice work on the refactor. Can you add a one line comment right after Line 33 describing what this is accomplishing and then I'll get this merged.

lyttonhao:
Okay, I will update it soon. Thanks.

ethanwhite:
Thanks for all your hard work on this @lyttonhao.

PR #277, opened 2015-03-09
Check for and use system proxies for downloading files

prabh27:
In some cases when the user is using a proxy urlib.urlopen() will fail to successfully open https files. This prevents the retriever from accessing the scripts stored on GitHub and causes the installation to fail (see #268). This change checks for the existence of proxies and makes them available in a way that urllib.urlopen() can find them.


ethanwhite:
Thanks for this. It's definitely a big step in the right direction. A few more things to think about:

* On line 373 of engine.py it looks like you have a left over `proxies=proxies` from trying out the other approach. I think this is what's causing the failed tests (except the one noted earlier)
* I think we can clean up the function a little more, because if `os.getenv(proxy)` exists then `os.environ[proxy]` will have length > 0. Of course I might be missing something here. If I'm not then this should work just as well:

```python
def proxy_function():
    proxies = ["http_proxy","https_proxy","ftp_proxy","HTTP_PROXY","HTTPS_PROXY","FTP_PROXY"]
    for proxy in proxies:
        if os.getenv(proxy):
            for i in proxies: # Setting the other proxies of the system.
                os.environ[i] = os.environ[proxy]
                break
```

* How about putting this function in `__init__.py` instead of in it's own module.
* Does this function need to be called in the module where it is used or can it just be called once before the first `urllib.urlopen()` call. If it's the later then we could also just call this function in `__init__.py` as well.
* I'm sure that your editor just automatically did the whitespace cleanup for you, and we definitely need to get the whitespace cleaned up, but could you please submit this without the whitespace changes. The issue is that all of the whitespace changes make this commit difficult to read, so if we need to look back on it the history will be much more confusing this way.

Thanks for all your work on this.

prabh27:
Thanks for pointing out the error in engine.py. I forgot to remove that line. The line len(os.environ()>0) is important, as there might be a case, where a proxy exists in the system, but is set to null. That would result in setting up all the other proxies to null as well.
Yes, I'll add this function in __init__.py.
I hope the next code I submit doesn't have whitespace cleanup issues. 
Thanks for your guidance.

ethanwhite:
Thanks. I think we're getting close:

1. At this point I think there shouldn't be any changes to `respository.py` or `engine.py` (since we've moved things into `__init__.py`, but the whitespace and one other change are still showing up. If you're familiar with rebasing in git you could get rid of the earlier commits, but it's may be easier to create a fresh branch from master and make a single commit with your changes in to `__init__.py`. When you fix this I'd add the nice description from you PR to the body of the commit message (best practice for commit messages is a <50 char first line that provides a summary, followed by a blank line, followed by an expanded discussion if needed)
2. Right now it doesn't look like your function is actually being executed, it's just being defined. So you'll need to call the function in `__init__.py` as well. Let's also rename it to `set_proxy` while we're at it since we don't need that name for the module any more.
3. Regarding the case where a proxy is "set to null", what will `os.getenv()` and `os.environ[proxy]` return. If it's `None` then `if os.getenv()` will evaluate to `False` and the first `if` statement won't be entered. If it's a string value like `"null"`, then `len(os.environ[proxy]` will be 1 not 0. If there's a case I'm missing, then I think the code would be made clearer by explicitly testing for the null case rather than testing for a len() > 0.

prabh27:
I am sorry for some silly mistakes that I made. I am using PEP-8 Style whitespace cleanup. If you want me to use something else, please inform me. I changed the name of the function in __init__.py and also called it(sorry I forgot to call it last time).
len(os.environ) is required, because if you execute the command-> "export  HTTP_PROXY=""" and then check os.getenv("HTTP_PROXY"), it exists. And len(os.environ()) in this case will be 0. So, if I don't mention that, all the proxy variables will be set to empty.

ethanwhite:
No problem at all. Thanks for all of your hard work on this. At a quick glance the new PR looks good. I'll do some quick tests on it this afternoon and then hopefully get it merged.

PEP-8 whitespace is absolutely good, it's just that since we haven't always used it in the past cleaning it up at the same time as other changes makes the changes difficult to read. I'm actually going to add an issue right now to remind me to clean up the whitespace so that we don't run into this again.

ethanwhite:
Closed by #278.

PR #278, opened 2015-03-10
Check for and use system proxies for downloading files

prabh27:
In some cases when the user is using a proxy urlib.urlopen() will fail to successfully open https files. This prevents the retriever from accessing the scripts stored on GitHub and causes the installation to fail (see #268). This change checks for the existence of proxies and makes them available in a way that urllib.urlopen() can find them


ethanwhite:
Looks good to me. Thanks again for this @prabh27!

PR #280, opened 2015-03-10
Cleanup new set_proxy function

ethanwhite:
1. Move `set_proxy` to the main `__init__` file
2. Add a docstring
3. Switch to four space indentation

@prabh27 - I missed the fact that f55af10f1767d39fdba65fb4c17beee526f96748 was going into `lib/__init__.py` instead of the one in the root folder. While I was fixing that I did a little additional cleanup. Could you test this on your system with a proxy just to make sure everything still works. Thanks.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -93,6 +93,18 @@ def ENGINE_LIST():
     return engine_list
 
 
+def set_proxy():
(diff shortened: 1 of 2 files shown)


prabh27:
Hey, Sorry for the late reply. 
Everything works fine on my system. I have checked it. :)

ethanwhite:
Great. Thanks for testing!

Issue #58, opened 2012-06-27
Add XML engine


ethanwhite:
I think ecology is finally getting to the point of being read to move into an XML world. It would be nice to be ready.

akshayah3:
I think this would be very similar to the `JSON` engine. I have made a [Gist] (https://gist.github.com/akshayah3/f23d85dc726b85b0ee1c) with some changes to the json engine code. I have used `<params>` to mark the beginning of the xml file. And i have used the `xmlrpclib` module from the standard library to encode a xml file which is similar to `json.dump()`. Before I proceed any further, would appreciate a review.

ethanwhite:
Thanks for looking into this @akshayah3! I don't know anything about the details of `xmlrpclib`, but it looks like a good translation from the JSON engine on a first read through. Have you tested it out? Can `xmlrpclib` read in the resulting XML?

I'd recommend going ahead and submitting this as a WIP pull request and we can continue the discussion there.

Ultimately that similarity between this, the JSON engine and the csv engine means that we should factor out the common functions behavior into a shared parent class for text based file formats, but that doesn't need to happen now.

PR #281, opened 2015-03-12
Added xml engine

akshayah3:
@ethanwhite I havent tested this yet. Would like some suggestions on how to efficiently test this keeping all the standards in mind.
And yes `xmlrpclib` reads in xml using `xmlrpclib.loads()`

### DIFF
--- a/engines/xml.py
+++ b/engines/xml.py
@@ -1,0 +1,125 @@
+import os
+import xmlrpclib
+from retriever.lib.models import Engine, no_cleanup
+from retriever import DATA_DIR
+
+class DummyConnection:
+    def cursor(self):
+        pass
+    def commit(self):
+        pass
+    def rollback(self):
+        pass
+    def close(self):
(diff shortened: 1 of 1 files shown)


ethanwhite:
Terribly sorry for being slow getting to this. An initial read of the code looks good overall. In order to actually use the engine you'll also need to add it to the list at the top of `engines/__init__.py`.

When I do this and try to install a simple dataset I get an error:

```
ethan@oryx:~$ retriever install xml MoM2003
=> Installing Masses of Mammals (Ecological Archives 2003)
argument must be tuple or Fault instance
```

This occurs after the creation of the file but before any data actually gets inserted, so it is likely an issue in `insert_statement()`.

dlebauer:
(fixes #58)

ethanwhite:
Closing due to inactivity, but when someone comes back to tackle #58 this could be a useful starting point.

PR #284, opened 2015-03-30
Add Global Wood Density Database, Fixes #275

### LINKED ISSUE #275: Add Global Wood Density Database
Data: http://datadryad.org/resource/doi:10.5061/dryad.234

This is a very valuable plant trait database used by lots of ecologists that is trapped inside a multi-tabbed Excel file. Working with Excel files like this can be tricky. The best example in the Retriever at the moment is in https://github.com/weecology/retriever/blob/master/scripts/gentry.py. Note that the Gentry script is much more complicated that this one would need to be since the tables are all in one file and are all well structured.

panks:
Have a look.
Added script file for Global Wood Density Database. Some values is reference sheet had special characters so added check in **lib/excel.py**. Also **download_file()** was leaving the xls unreadable after replacement, so had to add it to list along with .zip and .exe in **lib/engine.py**.

### DIFF
--- a/scripts/gwdd.py
+++ b/scripts/gwdd.py
@@ -1,0 +1,119 @@
+#retriever
+"""Retriever script for Zanne et al. Global wood density database.
+
+"""
+
+import os
+import sys
+import xlrd
+from retriever.lib.templates import Script
+from retriever.lib.models import Table
+from retriever.lib.excel import Excel
+
+
+
+class main(Script):
+    def __init__(self, **kwargs):
+        Script.__init__(self, **kwargs)
+        self.name = "Zanne et al. Global wood density database."
+        self.shortname = "GWDD"
+        self.urls = {"GWDD": "http://datadryad.org/bitstream/handle/10255/dryad.235/GlobalWoodDensityDatabase.xls?sequence=1"}
+        self.tags = ["Taxon > Plants", "Spatial Scale > Global",
+                     "Data Type > Observational"]
+        self.ref = "http://datadryad.org/resource/doi:10.5061/dryad.234"
+        self.addendum = """Citation for the database: Zanne, A.E., Lopez-Gonzalez, G.*, Coomes, D.A., Ilic, J., Jansen, S., Lewis, S.L., Miller, R.B., Swenson, N.G., Wiemann, M.C., and Chave, J. 2009. Global wood density database. Dryad. Identifier: http://hdl.handle.net/10255/dryad.235. 
+        *Correspondence for updates to the database: G.Lopez-Gonzalez@leeds.ac.uk
+        For descriptions of the database, see Chave et al. 2009. Towards a worldwide wood economics spectrum. Ecology Letters. Identifier: http://hdl.handle.net/10255/dryad.234
+
+        Below we list the rules of use for the Global wood density database. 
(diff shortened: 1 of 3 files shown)


scripts/gwdd.py
```diff
@@ -0,0 +1,130 @@
...
+                       ("Wood_Density"          ,   ("double", )),
+                       ("Region"                ,   ("char", )  ),
+                       ("Reference_Number"      ,   ("int",)    )]
+        table.pk = 'Number'
+        table.contain_pk = True
+
+        gwdd = []
+        for line in lines:
+            gwdd_data = [line["Number"],
+                            line["Family"],
```
ethanwhite: To match our general style all of the items in this list should be aligned with the opening delimiter.

scripts/gwdd.py
```diff
@@ -0,0 +1,130 @@
...
+                    def format_value(s):
+                        s = Excel.cell_value(s)
+                        return str(s).title().replace("\\", "/").replace('"', '')
+                    for num, label in enumerate(["Number", "Family", "Binomial", "Wood_Density",
+                                "Region", "Reference_Number"]):
+                        this_line[label] = format_value(row[num])
+                    lines.append(this_line)
+                except:
+                    raise
+                    pass
```
ethanwhite: What is the try/except accomplishing here? At a first glance it looks like if the try doesn't work it will just raise the exception anyway, so it may not be necessary.

scripts/gwdd.py
```diff
@@ -0,0 +1,130 @@
...
+                    def format_value(s):
+                        s = Excel.cell_value(s)
+                        return str(s).title().replace("\\", "/").replace('"', '')
+                    for num, label in enumerate(["Reference_Number", "Reference"]):
+                        this_line[label] = format_value(row[num])
+                    lines.append(this_line)
+                except:
+                    raise
+                    pass
+
```
ethanwhite: Same comment on the try/except as above.

scripts/gwdd.py
```diff
@@ -0,0 +1,130 @@
...
+        table.columns=[("Reference_Number"  ,   ("int",)    ),
+                       ("Reference"         ,   ("char", )  )]
+        table.pk = 'Reference_Number'
+        table.contain_pk = True
+
+        gwdd = []
+        for line in lines:
+            gwdd_ref = [line["Reference_Number"],
+                            line["Reference"]
+                            ]
```
ethanwhite: Same comment on aligned with the opening delimiter.

ethanwhite:
Thanks @panks! In general this looks like it's in really good shape.

The biggest thing is that when I test it the last column in both tables is missing.

Also, I think this should actually be split into three commits: one for the additional clean line endings expection, one for the unicode check, and one for the addition of script.

panks:
Oh! I just checked instead **table.contains_pk** I typed **tabel.contain_pk** which is setting **column** value to 0 instead of -1 in Table class, due to which last row is missing. I'll fix that and make the other changes you mentioned.

panks:
I checked, now everything working fine. Have a look.

ethanwhite:
Everything looks great except a little confusion that resulted from how GitHub displayed my inline comment about items in a list being aligned with the opening delimiter. I was referring to the lines starting at 71 and 107. With the correct general format being:

```
gwdd_ref = [line["Reference_Number"],
            line["Reference"]]
```
So, the lists on lines 60 and 100 should go back to the way they were before and the lists starting at 71 and 107 should be updated.

Once that's cleaned up this is ready to go in.

panks:
Done!

ethanwhite:
Awesome. Great work @panks! In it goes.

PR #286, opened 2015-03-30
Added eBird observation dataset

akshayah3:
@ethanwhite The only thing new in the code is the `Table` class imported at the top.


ethanwhite:
Thanks. Have you run this and checked to see if it worked? If not, can you run it through at least MySQL and PostgreSQL as those are the most likely places that something funny could happen.

akshayah3:
@ethanwhite I am having some problems integrating this newly added script to the retriever. It is not getting added to the existing list of scripts even after updating the retriever. Could you tell me how to add a new script to the retriever?

ethanwhite:
@akshayah3 you have two choices. What I typically do for testing new scripts is just to run the retriever from inside the root directory for the source code. The retriever checks for a `./scripts` directory, which is there since you're in the root of the repo, and finds the script. Alternatively, you can copy the script into the `~/.retriever/scripts/` directory.

akshayah3:
@ethanwhite I have made a small change. Could you check the new code?

ethanwhite:
The changes look fine. Is this working for you? When I run `retriever ls` it doesn't see the dataset (which means I can't do the install), but I can't see why from a first look.

akshayah3:
@ethanwhite Unfortunately that is what I have been trying to debug for some time now.

ethanwhite:
Ah, OK, let me take a closer look.

ethanwhite:
Sorry I didn't see this earlier. You need to add `#retriever` as the first line of the file so that the retriever knows this is one of its scripts.

akshayah3:
@ethanwhite Thanks for the suggestion! It works for me now.

ethanwhite:
When I run this I get the following error.

```
ethan@simon:~/$ retriever install postgres eBirdOD -u XXXXX -p XXXXX

=> Installing eBird Observation Dataset
Creating database eBirdOD...
Creating table eBirdOD.main...
Couldn't create table (syntax error at or near ","
LINE 1: ...PROVINCE varchar, COUNTY varchar, DECIMALLATITUDE , DECIMALL...
                                                             ^
). Trying to continue anyway.
INSERT INTO eBirdOD.main (BASISOFRECORD, INSTITUTIONCODE, COLLECTIONCODE, CATALOGNUMBER, OCCURRENCEID, RECORDEDBY, YEAR, MONTH, DAY, COUNTRY, STATEPROVINCE, COUNTY, DECIMALLATITUDE, DECIMALLONGITUDE, LOCALITY, KINGDOM, PHYLUM, CLASS, ORDER, FAMILY, GENUS, SPECIFICEPITHET, SCIENTIFICNAME, VERNACULARNAME, INDIVIDUALCOUNT) VALUES ('HumanObservation', 'CLO', 'EBIRD_CAN', 'OBS142125343', 'URN:catalog:CLO:EBIRD_CAN:OBS142125343', 'obsr199897', 2012, 2, 20, 'Canada', 'Nova Scotia', 'Lunenburg', null, null, 'Mahone Bay', 'Animalia', 'Chordata', 'Aves', 'Charadriiformes', 'Alcidae', 'Cepphus', 'grylle', 'Cepphus grylle', 'Black Guillemot', null);
syntax error at or near "ORDER"
LINE 1: ...CIMALLONGITUDE, LOCALITY, KINGDOM, PHYLUM, CLASS, ORDER, FAM...
```

scripts/eBird_observation.py
```diff
@@ -0,0 +1,52 @@
...
+                       ("OCCURRENCEID",("char", )),
+                       ("RECORDEDBY",("char", )),
+                       ("YEAR",("int", )),
+                       ("MONTH",("int", )),
+                       ("DAY",("int", )),
+                       ("COUNTRY",("char", )),
+                       ("STATEPROVINCE",("char", )),
+                       ("COUNTY",("char", )),
+                       ("DECIMALLATITUDE",("real", )),
+                       ("DECIMALLONGITUDE",("real", )),
```
ethanwhite: `"real"` should be `"double"`' on 34 and 35.

scripts/eBird_observation.py
```diff
@@ -0,0 +1,52 @@
...
+                       ("LOCALITY",("char", )),
+                       ("KINGDOM",("char", )),
+                       ("PHYLUM",("char", )),
+                       ("CLASS",("char", )),
+                       ("ORDER",("char", )),
```
ethanwhite: `ORDER` is a reserved keyword in most SQL dialects. This should be replaced with something like `TAXONOMICORDER`.

ethanwhite:
I think I found the two issues that were preventing this from executing. The issue with the keyword `order` should have been handled automatically, so I've opened a new issue to look into that. Now we just need to see if all of the data gets in cleanly.

This PR has been great at turning up other issues that we need to deal with :)

ethanwhite:
This appears to be working for me after the in-line fixes I recommended above.

@akshayah3 - once those changes are made I think we can get this in. Thanks!

akshayah3:
@ethanwhite Made the changes. Hope it looks good now.

ethanwhite:
Looks great! Thanks!

PR #287, opened 2015-03-31
Check input path on download. Fixes #205

### LINKED ISSUE #205: download command should probably fail when specified path does not exist
A datsaet can be downloaded to a specific path with the function `download` while specifying the -p argument. For example `retriever download MCDB -p my_path` will download each of the MCDB files and then copy them to the directory my_path but if my_path does not exist a file called my_path is created and the files overwrite that file as each is copied from the download directory. It may be best if the retriever fails with a warning that the path does not exist.

panks:
Have a look, is this good enough, or we should add some other checks?
shutil.copy() can fail in two situation:
1- No write access to location
2- File already exists - In this case it already gives a error message saying it's the same file.

So should we catch the exception and prints it, or print our own error string, as I have done in this patch (it's cleaner but the user wouldn't know the failure is due to which of the cases above)?

### DIFF
--- a/engines/download_only.py
+++ b/engines/download_only.py
@@ -51,7 +51,18 @@ class engine(Engine):
                     print("Keeping existing copy.")
                 else:
                     print("Copying %s from %s" % (file_name_nopath, file_path))
-                    shutil.copy(file_name, self.opts['path'])
+                    if os.path.isdir(self.opts['path']):
+                        try:
+                            shutil.copy(file_name, self.opts['path'])
+                        except:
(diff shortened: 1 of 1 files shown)


ethanwhite:
Looks great. Can you expand the commit message to explain the problem this is solving and move the "Fixes 205" to the bottom line of that expanded commit. After that I think this is good to go.

panks:
Done!

ethanwhite:
Thanks @panks! Looks good to me.

PR #288, opened 2015-04-01
Added a few docstings and refactored a small code

akshayah3:
@ethanwhite Do have a look.


ethanwhite:
Nice work @akshayah3! Thanks!

PR #289, opened 2015-04-01
Refactored the entire Json format_insert_value()

akshayah3:
@ethanwhite What do you think about this?

### DIFF
--- a/engines/jsonengine.py
+++ b/engines/jsonengine.py
@@ -87,40 +87,7 @@ class engine(Engine):
            in json
 
         """
-        #TODO There is a lot of duplicated code with engine.format_insert_value
-        #Refactoring so that this code could be properly shared would be preferable
-        datatype = datatype.split('-')[-1]
-        strvalue = str(value).strip()
-
-        # Remove any quotes already surrounding the string
-        quotes = ["'", '"']
-        if len(strvalue) > 1 and strvalue[0] == strvalue[-1] and strvalue[0] in quotes:
-            strvalue = strvalue[1:-1]
-        nulls = ("null", "none")
-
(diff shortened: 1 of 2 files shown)


akshayah3:
I was going through the code base, we can also use `correct_invalid_value` to replace `nulls` with `None`

ethanwhite:
Thanks for taking a try at this @akshayah3. I don't think we want to introduce engine specific changes to engine.py as this makes things less modular and more difficult to maintain.

There some refactoring to be done in this general area that I think we talked about in another issue, which addresses the fact that the json and csv engines (as well as any other text file based engines like the XML engine you started) have a lot of overlap that could be reduced by creating a parent engine for text files and having all of the text file engines inherit from it.

If I've missed something about your justification for these changes, please let me know. @bendmorris do you have any thoughts on this?

akshayah3:
@ethanwhite Sure. Your justifications are valid indeed. I had a go at this because it was mentioned in the inline comments that some refactoring might be needed here. Do you have any ideas as to how do we refactor this code in general, If yes I will continue with that in this PR, otherwise I will close this.

bendmorris:
:+1: agree with your reasoning @ethanwhite

ethanwhite:
Sorry I've been slow on this @akshayah3 - I do think there is some refactoring to be done, but I think it should be in a separate PR. In short, all of the text file based engines have a lot of overlap that could be reduced by creating a parent engine for text files and having all of the text file engines inherit from it. Happy to discuss this further if it's something you want to tackle. Just go ahead and open a new issue or PR.

akshayah3:
Sure! Will open an issue

PR #296, opened 2015-04-03
Clean column names and Check for already existing file in download_only

panks:


panks:
Extra test cases are failing as MoM dataset script contains column name **sporder** and in `clean_column_name()` after replacement it's becoming **spsporder**, so we can either update the script file or the hash value in test file.

ethanwhite:
Hi @panks - terribly sorry to let this languish for so long. Part of this is now in 12f27072968168a90213dea9cdefb630bb6da1ee and the rest is in #408. Thanks again for all of your work on this. I'm going to go ahead and close this PR since the other two took/take care of it.

Issue #269, opened 2015-03-06
Regression test failing for Del Moral dataset on MySQL


ethanwhite:
A regression test has just started failing on Travis for the Del Moral dataset on MySQL. E.g.;
https://travis-ci.org/weecology/retriever/builds/53258092

The tests are still passing locally for me and the PRs I merged today were all passing Travis when submitted, so I restarted some older builds that have been passing for extended periods of time and they are also now failing in the same way. I'm guessing that Travis has changed it's MySQL version, or something similar, resulting in a change in the hash.

ethanwhite:
Interesting Travis is using the same SQL version (except for the Ubuntu part)

![screenshot from 2015-03-21 21 11 22](https://cloud.githubusercontent.com/assets/744427/6767858/155dff36-d00f-11e4-927a-d3ca7bc0d1e2.png)

as my local system

```
ethan@oryx:~$ mysql -u travis -Bse "SELECT @@version"
5.5.41-0ubuntu0.14.10.1
```

This test is passing fine on my local system.

ethanwhite:
The test is failing because the bulk insert for the species table that is attempted by the MySQL engine is working on Travis, but not on my local machine. This means that the clean up steps in Engine.add_to_table get skipped for the species table because it is imported via mysql.insert_data_from_file. Since there are extra spaces in some of the family names, these don't get removed on Travis, but they do on my dev machine. This means that the dumps differ in the presence of the spaces and the test fails. Travis must have tweaked this configuration setting recently, which is what caused the test to start to fail.

ethanwhite:
The fix for this should be to simply add

`*do_not_bulk_insert: True`

To the .script file under the `species` table.

This doesn't currently work. It appears that when the retriever is installed the `__init__` in `Script` is run twice (at least for DelMoral, probably for everything). The first time it is run it gets the passed in attributed in the `tables` dictionary correctly. The second time the attribute is missing and `tables` is overwritten without the necessary attribute.

@bendmorris - any thoughts about where this double calling of Script/BasicTextTemplates `__init__` is happening?

ethanwhite:
Never mind. The script was getting installed twice because it was in both `~/.retriever/scripts/` and in the repository. Fix coming soon (hopefully).

ethanwhite:
Fixed in a530e15c72e02db1b93210fe63fe0dc9b31d613f.

PR #300, opened 2015-04-12
Add script for Palmer et al 2007

dmcglinn:
This commit adds the Oosting forest plant dataset from Palmer et al. 2007 which is an ecological archives dataset: http://esapubs.org/archive/ecol/E088/162/

### DIFF
--- a/scripts/EA_palmer2007.script
+++ b/scripts/EA_palmer2007.script
@@ -6,6 +6,7 @@ ref: http://esapubs.org/archive/ecol/E088/162/
 
 table: pres, http://esapubs.org/archive/ecol/E088/162/Oosting_pres2.txt
 table: species, http://esapubs.org/archive/ecol/E088/162/species_codes.txt
+replace: check, checkrec
 table: trees, http://esapubs.org/archive/ecol/E088/162/Oosting_Trees_1998.txt
 *nulls: '.', ''
 table: env, http://esapubs.org/archive/ecol/E088/162/Oosting_env.txt
--- a/scripts/EA_palmer2007.script
+++ b/scripts/EA_palmer2007.script
@@ -1,0 +1,10 @@
+name: North Carolina forest plants (Ecological Archives 2007)
+shortname: Palmer2007
+description: Michael W. Palmer, Robert K. Peet, Rebecca A. Reed, Weimin Xi, and Peter S. White. 2007. A multiscale study of vascular plants in a North Carolina Piedmont forest. Ecology 88:2674.
+tags: Taxon > Plants, Spatial Scale > Local, Data Type > Time Series, Data Type > Observational
+ref: http://esapubs.org/archive/ecol/E088/162/
+
+table: pres, http://esapubs.org/archive/ecol/E088/162/Oosting_pres2.txt
+table: species, http://esapubs.org/archive/ecol/E088/162/species_codes.txt
+table: trees, http://esapubs.org/archive/ecol/E088/162/Oosting_Trees_1998.txt
+table: environment, http://esapubs.org/archive/ecol/E088/162/Oosting_env.txt
--- a/scripts/EA_palmer2007.script
+++ b/scripts/EA_palmer2007.script
@@ -7,4 +7,5 @@ ref: http://esapubs.org/archive/ecol/E088/162/
 table: pres, http://esapubs.org/archive/ecol/E088/162/Oosting_pres2.txt
 table: species, http://esapubs.org/archive/ecol/E088/162/species_codes.txt
 table: trees, http://esapubs.org/archive/ecol/E088/162/Oosting_Trees_1998.txt
-table: environment, http://esapubs.org/archive/ecol/E088/162/Oosting_env.txt
+*nulls: '.', ''
+table: env, http://esapubs.org/archive/ecol/E088/162/Oosting_env.txt


dmcglinn:
From what I can tell it looks like the Travis CI build failed because of tests on three datasets that are unrelated to my changes.

ethanwhite:
Yeah, the Travis failure is discussed in #269. We've run into this before where some subtle difference in the version of the database management system between our systems and the CI causes a mismatch in the hashes. I'm trying to figure out a more stable way to structure those tests. Just. Need. More. Time. :)

Have you tested this locally? When I run the PR into SQLite I get an error:

```
ethan@oryx:~/$ retriever install sqlite Palmer2007 -f palmer2007.sqlite
=> Installing North Carolina forest plants (Ecological Archives 2007)
Creating table Palmer2007_environment...
Inserting rows to Palmer2007_environment: 289 / 289
Creating table Palmer2007_pres...
Inserting rows to Palmer2007_pres: 21253 / 21253
Creating table Palmer2007_species...
Inserting rows to Palmer2007_species: 224 / 224
Creating table Palmer2007_trees...
Couldn't create table (near "TEXT": syntax error). Trying to continue anyway.
INSERT INTO Palmer2007_trees (id, code, fx, fy, d90, c90, d98, c98, f1, f2, f3, f4, azm, tx, ty, nx, ny, cl, check, notes) VALUES (574, 'POST', '0', 0, '.', '.', '.', '.', '', '', '', '.', '.', '0', '2400', '0', '0', '.', null, null);
near "check": syntax error
```

Are you seeing the same thing?

dmcglinn:
I see the same error for sql as well. I'll try to see which rows in which table are causing the problem.

dmcglinn:
The error is cause because there are both empty cells and cells with "." that are intended to be null values. Here is a reference to R code that I developed to deal with the problems. 

https://github.com/weecology/mete-spatial/blob/master/scripts/spat_empir_data_filtering.R#L152

I suppose it is necessary to convert this R code to python?

ethanwhite:
Nope, you just need to tell the retriever what the null values are. See the Null Values section of:
http://www.ecodataretriever.org/scripting.html

dmcglinn:
sweet I'll take a look!

dmcglinn:
I added dots ('.') and blanks ('') as null values see ea49302, but I'm still getting errors when installing to sqlite. Specifically the output messages from retriver are as follows:
```
$ retriever install sqlite Palmer2007
=> Installing North Carolina forest plants (Ecological Archives 2007)
Downloading Oosting_pres2.txt...
Creating table Palmer2007_pres...
Inserting rows to Palmer2007_pres: 21253 / 21253
Downloading species_codes.txt...
Creating table Palmer2007_species...
Inserting rows to Palmer2007_species: 224 / 224
Creating table Palmer2007_trees...
Couldn't create table (near "TEXT": syntax error). Trying to continue anyway.
INSERT INTO Palmer2007_trees (id, code, fx, fy, d90, c90, d98, c98, f1, f2, f3,
f4, azm, tx, ty, nx, ny, cl, check, notes) VALUES (574, 'POST', 0, 0, null, null, null, null, null, null, null, null, null, 0, 2400, 0, 0, null, null, null);
near "check": syntax error
Done!
```
I'm not sure what's causing the problems but maybe something in the column names that sqlite doesn't like? Also the dataset does have a stray comma in a column by itself row 12467 in column labeled d98. I tried to list comma as a null value as well but this caused problems I think because the notes column has commas throughout it.

ethanwhite:
Turns out that `check`, which is one of the column titles, is a reserved keyword in SQL. You can replace the name by adding

`replace: check, someothernameforcheck`

immediately above `table: trees, http://esapubs.org/archive/ecol/E088/162/Oosting_Trees_1998.txt`

dmcglinn:
Sweet thanks @ethanwhite! If there's no easy way to fix the stray comma on row 12467 I think this is ready to be merged.

ethanwhite:
Is that comma on 12467 supposed to indicate a null? If so, then you could just add `,` to the list of nulls and it would get rid of it.

dmcglinn:
I tried that but it threw errors I think because there are commas embedded
in the text in the last column which contains notes.

dmcglinn:
here's the error message:
```
Daniel@MCGLINN-DELL ~/Documents/retriever/scripts (EA_palmer2007)
$ retriever install csv Palmer2007
  File "c:\Anaconda\Scripts\retriever-script.py", line 9, in <module>
    load_entry_point('retriever==1.6', 'console_scripts', 'retriever')()
  File "c:\Anaconda\lib\site-packages\retriever-1.6-py2.7.egg\retriever\__main__.py", line 40, in main
    script_list = SCRIPT_LIST()
  File "c:\Anaconda\lib\site-packages\retriever-1.6-py2.7.egg\retriever\__init__.py", line 88, in SCRIPT_LIST
    return [module.SCRIPT for module in MODULE_LIST(force_compile)]
  File "c:\Anaconda\lib\site-packages\retriever-1.6-py2.7.egg\retriever\__init__.py", line 67, in MODULE_LIST
    compile_script(join(search_path, script_name))
  File "c:\Anaconda\lib\site-packages\retriever-1.6-py2.7.egg\retriever\lib\compile.py", line 43, in compile_script
    nulls = [eval(v) for v in [v.strip() for v in value.split(',')]]
  File "<string>", line 1
    '
    ^
SyntaxError: EOL while scanning string literal
```

ethanwhite:
Yep, you're totally right. I'd been assuming they were enclosed, but they aren't.

Thanks for this contribution @dmcglinn! It's going in.

PR #304, opened 2015-04-17
WIP: Remove unused code form lib/tools.py

ethanwhite:
Mostly testing code from 2010 that we don't use any more.


ethanwhite:
The remote tests are currently showing meaningful failures. This PR should not be merged at the moment.

ethanwhite:
Tests are passing now, but testing output shows repeated instances of:

`Failed to load script: gentry (/home/travis/.retriever/scripts/)`

which shouldn't be happening.

ethanwhite:
I had missed an import of `TestScript`, which is now fixed. It looks to me like the:

`Failed to load script: gentry (/home/travis/.retriever/scripts/)`

is resulting from actually testing the old script rather than the new one due to a script order issue.

@henrykironde if you agree I think we can merge this, but I'm also happy to wait until we work out that ordering of script priority issue before merging to make sure that extra output goes away.

PR #307, opened 2015-04-18
On Windows have CLI default to storing data in directory retriever is run from

### LINKED ISSUE #250: Improve Windows installer version handling of default paths
The Windows Installer version of the retriever currently defaults to installing files from file based engines onto the Desktop. This was done because if the GUI is launched from the start menu it's working directory is a system path. This has caused confusion https://github.com/ropensci/ecoretriever/issues/43 and should be improved.

The simplest solution is probably to check to see if the retriever is running as the CLI, and if it is to not override the default path.

ethanwhite:
To avoid issues with the Installer installed version of the Retriever confusing folks
by defaulting to storing their data in the system folder where the Retriever is installed,
we had defaulted to storing all data on Windows to the Desktop. This is undesirable for the
CLI and for things that wrap the CLI like the ecoretriever R package. This keeps the Desktop
as the default when using the GUI on Windows, but returns the CLI behavior to the standard
of defaulting the to current working directory.

Fixes #250.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -51,10 +51,14 @@ DATA_SEARCH_PATHS =     [
                          ]
 DATA_WRITE_PATH =       DATA_SEARCH_PATHS[-1]
 
-# create a default data directory for Windows since the Windows installer places
-# the executable in a place where users won't expect the data to be stored.
-user_desktop = os.path.join(os.path.expanduser('~'), 'Desktop')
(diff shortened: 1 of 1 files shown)


ethanwhite:
@dmcglinn - this provides a central fix to https://github.com/ropensci/ecoretriever/issues/43

Can you take a peak and give it a quick test if you have a few minutes. Thanks.

PR #311, opened 2015-04-24
Add url field to sample script

### LINKED ISSUE #301: Should "ref" or "url" be a default tag
The new script generation approach creates a generic file that lacks the ref or url tag. I think this tag should be included on the generic script because all of the datasets are coming from someplace online. Examining some of the script files suggests to me that the tags "ref" and "url" can be used interchangeably. Essentially this tag ensures that when browsing for datasets using the retriever you also see a hyperlink to the dataset.

ethanwhite:
This encourages the inclusion of a link to the source of the data.

Fixes #301.


dmcglinn:
This looks great. I'm going to merge it.

PR #313, opened 2015-05-06
add default script directory to dev docs

davharris:
BTW, how do users find these docs? I didn't see a way to access it from the command line or from the GUI.


ethanwhite:
The docs system is currently broken and more generally in a state of disrepair. It's on the list to update to a newer documentation system and get everything up on Read The Docs. You're just finding all of the embarrassing bits of the Retriever this week :sweat_smile:.

Fortunately we're hiring a dev soon so we should start to have more hours to clean things up.

PR #319, opened 2015-05-09
fix string formatting in engine.py

jamesaq:
Uses string formatting `format()` instead of `replace()`


ethanwhite:
Thanks @jamesaq!

Issue #332, opened 2015-08-16
Home range size database


embaldridge:
Dryad Description:
Database of mean species masses and corresponding empirically measured home range sizes for 569 vertebrate species from across the globe, including birds, mammals, reptiles, and fishes.

When using this data, please cite the original publication:
Tamburello N, Côté IM, Dulvy NK (2015) Energy and the scaling of animal space use. The American Naturalist 186(2): 196-211. http://dx.doi.org/10.1086/682070

Additionally, please cite the Dryad data package:
Tamburello N, Côté IM, Dulvy NK(2015) Data from: Energy and the scaling of animal space use. Dryad Digital Repository. http://dx.doi.org/10.5061/dryad.q5j65 http://datadryad.org/resource/doi:10.5061/dryad.q5j65/1

ethanwhite:
Closed by #333.

PR #333, opened 2015-08-24
Adds script for Database of Vertebrate Home Range Sizes.

embaldridge:
Creates script for importing Database of Vertebrate Home Range Sizes by Tamburello et al. 2015 from Dryad.  Solves #332 .


ethanwhite:
Thanks for the contribution @embaldridge!

This wouldn't run on my machine, which is weird since it seems to be passing on Travis, but the issue was easy to fix. It looks like the you to change `NA` to `'NA'` when declaring the nulls.

embaldridge:
@ethanwhite Oops, should work now, assuming the build still passes.

embaldridge:
Oops, should work now.

embaldridge:
Should be good to go now, sorry about that.

ethanwhite:
Looks great! Thanks.

PR #338, opened 2015-09-16
Allow downloading files without names in urls

### LINKED ISSUE #337: Support downloading of files via web services
Many web services return files whose names are not directly encoded in the url. For example, we can download data from the PRISM web service using urls like:
http://services.nacse.org/prism/data/public/4km/tmin/200904

which returns a file named:
`PRISM_tmin_stable_4kmM2_200904_bil.zip`

Our current system doesn't play well with this because we are typically getting determining the filename from the url by assuming that the end of the url is the filename. E.g.,
https://github.com/weecology/retriever/blob/master/lib/engine.py#L383
https://github.com/weecology/retriever/blob/master/lib/engine.py#L553

This is currently blocking incorporating the PRISM data into the retriever.

ethanwhite:
When using web services to retrieve files the file names are often not included in the url.
The retriever uses urls to create file names for raw data storage and access so this change
adds the ability to add a prefix to the "file name" that is being parsed from the url. In many webservices
this "file name" may just be the last parameter set and so require the prefix to be informative.

In some cases this results in an archive without the appropriate file extension, so this also makes
sure that the line endings are never cleaned in files downloaded using `download_files_from_archive`.

Fixes #337 (at least for the short-term).

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -377,11 +377,15 @@ class Engine():
             file.close()
 
     def download_files_from_archive(self, url, filenames, filetype="zip",
-                                    keep_in_dir=False, archive_prefix=''):
-        """Downloads one or more files from an archive into the raw data
-        directory."""
+                                    keep_in_dir=False, archivename=None):
+        """Downloads files from an archive into the raw data directory.
+
+        """
         downloaded = False
-        archivename = self.format_filename(filename_from_url(url, prefix=archive_prefix))
+        if archivename:
+            archivename = self.format_filename(archivename)
+        else:
+            archivename = self.format_filename(filename_from_url(url))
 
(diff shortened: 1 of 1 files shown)


ethanwhite:
@henrykironde can you give this a look over when you get the chance.

ethanwhite:
@henrykironde - if you're happy with this you can go ahead and push the `Merge pull request` button.

ethanwhite:
I've rerolled this following the merge of #340 so that it can now be merged. If Travis is happy then we should go ahead and merge this (since the PRISM script currently relies on it). We can then come back and reconsider the design of both this and #340.

ethanwhite:
Closing in favor of #341.

PR #339, opened 2015-09-16
Add script for downloading the modern PRISM climate data

ethanwhite:
The modern PRISM data provides monthly raster precip and temperature measures
for the United States from 1981 to present.

Currently all files from all of the archives are moved into a single directory. At some point it may be better to override final_cleanup and add each archives files to a directory of the matching name.

Relies on #338.


ethanwhite:
@henrykironde - if you're happy with this you can go ahead and push the `Merge pull request` button.

PR #340, opened 2015-09-20
Allow keeping downloaded files in subdirectories

ethanwhite:
Adds options for keeping files downloaded files in subdirectories for:

1. Files extracted from archives in the raw data directories
2. Downloaded files when they are transferred to their final destination


ethanwhite:
This feels a little awkward to me as it's written, but it's the best I could come up with. If there's a better way to approach this problem I'd love recommendations.

PR #341, opened 2015-09-24
Allow downloading archives without names in urls

ethanwhite:
When using web services to retrieve files the file names are sometimes not included in the url.
The retriever uses urls to create file names for raw data storage and access so this change
adds the ability to add a prefix to the "file name" that is being parsed from the url. In some webservices
this "file name" may just be the last parameter set and so require the prefix to be informative.

To solve this for archives this adds an optional archivename argument to download_files_from_archive
that allows a database script to provide the name for the file to be saved.

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -377,11 +377,15 @@ class Engine():
             file.close()
 
     def download_files_from_archive(self, url, filenames, filetype="zip",
-                                    keep_in_dir=False):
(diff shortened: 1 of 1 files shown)


ethanwhite:
This is a simpler and more direct version of #338 targeted at just solving the problems for archives, using existing functionality. The larger problem in #337 can then be addressed later.

PR #343, opened 2015-09-24
Add Slack integration to Travis CI

ethanwhite:


ethanwhite:
Looks like this is working since this just showed up in Slack.

screenshot from 2015-09-24 14 55 39

PR #350, opened 2015-10-13
Forkedretriever

henrykironde:
more examples on using the retriever command line options

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,26 +83,26 @@ optional arguments:
 ```
 These examples are using Breeding Bird Survey data (BBS) 
 
-######Using download 
+Using download 
 
   `retriever download -h    (gives you help options)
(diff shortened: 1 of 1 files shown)


ethanwhite:
This is very useful addition @henrykironde. Thanks.

I'd suggest two minor improvements:

1. I'd suggest moving the `download` section below the `install` section. More people will be using `install` than `download` and I think having the most used stuff come first might avoid some confusion.
2. Since this history has gotten a little messy let's go ahead and squash all of this into a single commit. I'm happy to walk you through the process if you're not familiar with it.

Other than that it looks great and will definitely help folks figure out how to make the most of the retreiver.

ethanwhite:
Closing in favor of #351.

PR #359, opened 2015-11-11
Henrye

henrykironde:
changing --subdir to be a flag and taking no arguments


ethanwhite:
Hi @henrykironde - everything in here looks great and works perfectly. The one issue is that the message for the first commit doesn't reflect the changes, which are the changes you made improve behavior on Windows. Let's get that commit message updated and then this is ready to go.

ethanwhite:
Thanks @henrykironde.

PR #367, opened 2015-12-07
add the code coverage metrics and travis legacy infrastructure (CL).

henrykironde:


.travis.yml
```diff
@@ -1,22 +1,41 @@
...
 language: python
 python:
   - "2.7"
 # command to install wxPython using apt
-before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq python-wxgtk2.8
+# before_install:
+#   - sudo apt-get update -qq
+#   - sudo apt-get install -qq python-wxgtk2.8
```
ethanwhite: Let's delete these 3 lines instead of commenting them out, assuming I'm understanding correctly that we don't need them now.

.travis.yml
```diff
@@ -1,22 +1,41 @@
...
+addons:
+  apt:
+    packages:
+      # for wxPython:
+      - python-wxgtk2.8
+      - python-wxtools
+      - wx2.8-doc
+      - wx2.8-examples
+      - wx2.8-headers
+      - wx2.8-i18n
```
ethanwhite: Do we need to install wx for testing purposes now that the retriever only loads it for the GUI following c73c39c? I suspect we can remove the wx imports entirely.

henrykironde: Hi ethan,

Thanks for deleting of the commented lines.
For now we can also delete wx and may be, include it when we get to
improving testing for app modules to improve the coverage.

Regards

Henry S.

ethanwhite:
Sounds good. If you'll go ahead and update the PR by deleting the commented lines and the wx stuff (make the changes locally and then do another `git commit --amend` and `git push -f`) and leave a comment to let me know it's done we can get this merged in.

PR #369, opened 2015-12-10
add code coverage, Depsy, continuous integration 

henrykironde:


test/test_regression_cli.py
```diff
@@ -44,40 +59,43 @@ def _test_factory(test_method, name, *args):
...
-        current_md5 = getmd5('output_file')
+        os.system("echo .dump | sqlite3 output_database > output_file")
+        current_md5 = getmd5(unixfileformat("output_file")) 
         assert current_md5 == known_md5
 
-
 class CSVRegression(TestCase):
     def check_csv_regression(self, dataset, known_md5):
         """Check for regression for a particular dataset imported to csv"""
+        os.chdir(os.getcwd())
```
ethanwhite: Is this necessary? It seems like it would just look up the current working directory and the set the current working directory to that value, which seems the same as not running this line in the first place. If we do need this line we should probably add a comment to it since it's difficult (for me at least) to understand what it's accomplishing.

test/test_regression_cli.py
```diff
@@ -44,40 +59,43 @@ def _test_factory(test_method, name, *args):
...
         os.system("retriever install csv %s -t output_file_{table}" % dataset)
         os.system("cat output_file_* > output_file")
         current_md5 = getmd5('output_file')
         assert current_md5 == known_md5
 
-
 class MySQLRegression(TestCase):
     def check_mysql_regression(self, dataset, known_md5):
         """Check for regression for a particular dataset imported to sqlite"""
+        os.chdir(os.getcwd())
```
ethanwhite: Same comment as above.

test/test_regression_cli.py
```diff
@@ -44,40 +59,43 @@ def _test_factory(test_method, name, *args):
...
         os.system("mysqldump testdb -u travis --compact --compatible=no_table_options --no-create-db --no-create-info --result-file=output_file")
-        current_md5 = getmd5('output_file')
+        current_md5 = getmd5(unixfileformat("output_file")) 
         assert current_md5 == known_md5
 
-
 class PostgreSQLRegression(TestCase):
     def check_postgres_regression(self, dataset, known_md5):
         """Check for regression for a particular dataset imported to sqlite"""
+        os.chdir(os.getcwd())
```
ethanwhite: Same comment as above.

ethanwhite:
Thanks @henrykironde. This looks great. Merging.

PR #371, opened 2015-12-16
Vertnet data 

### LINKED ISSUE #357: Add vertnet data
http://vertnet.org/resources/datatoolscode.html

henrykironde:
fixes #357


ethanwhite:
Looks great! On a machine configured for bulk import it only takes about 10 minutes to install (not including the download).

PR #372, opened 2015-12-23
explicit gui retriever

henrykironde:
Switch to an explicit retriever-gui for running the GUI

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -25,107 +25,98 @@ from retriever.lib.get_opts import parser
 
 def main():
     """This function launches the EcoData Retriever."""
-    if len(sys.argv) == 1 or (len(sys.argv) > 1 and sys.argv[1] == 'gui'):
-        # if no command line args are passed, launch GUI
+    if len(sys.argv) == 1:
+        sys.argv.append('-h')
 
-        check_for_updates(graphical=False if current_platform == 'darwin' else True)
-        lists = get_lists()
+    script_list = SCRIPT_LIST()
+    args = parser.parse_args()
 
-        from retriever.app.main import launch_app
-        launch_app(lists)
+    if args.quiet:
+        sys.stdout = open(os.devnull, 'w')
 
-    else:
-        # otherwise, parse them
+    if args.command == 'help':
+        parser.parse_args(['-h'])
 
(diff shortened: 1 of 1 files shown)


ethanwhite:
Looks good overall. A couple of questions/suggestions:

1. If you haven't already we should build `retriever.exe` using `python setup.py py2exe` to make sure that it still runs properly (if executed it should launch the GUI).
2. Instead of printing out "arguments required, use retriever -h for help" when no arguments are provided, we should just display the help (the results of retriever -h). This is standard command line behavior (e.g., this is what `git` and `grep` do).
3. Is there an execution path that results in `args.command == 'gui'` evaluating to True when the command line call isn't `retriever gui`? I'm just wondering about the need for lines 73-77.

__main__.py
```diff
@@ -25,107 +25,102 @@
...
+    elif args.command == 'gui':
         check_for_updates(graphical=False if current_platform == 'darwin' else True)
+        lists = get_lists() 
+        from retriever.app.main import launch_app
+        launch_app(lists)            
+
         lists = get_lists()
 
         from retriever.app.main import launch_app
         launch_app(lists)
```
ethanwhite: Looks like lines 62-64 are direct copies of 66-69.

ethanwhite:
I noted what looks like some accidental code duplication. Once that's fixed we should be able to get this merged. Thanks!

ethanwhite:
Oh, looks like one more thing after a quick test. When I just run `retriever` I get:

```
ethan@stice:~/RetrieverCode ((285ed9a...))$ retriever 
usage: retriever [-h] [-v] [-q]
                 {download,install,update,gui,new,ls,citation,reset,help} ...
retriever: error: too few arguments
```

That similar to what `grep` does, but we don't want the `retriever: error: too few arguments` and we should probably add the equivalent of grep's `Try 'grep --help' for more information.`

ethanwhite:
Looks great! As soon as you've tested `retriever.exe` we can get this merged.

ethanwhite:
Note on current status: this is causing some issues with our binaries, so this is currently on hold until for 2.0.

ethanwhite:
Closing for the time being since this is on hold until 2.0.

PR #373, opened 2015-12-23
add Download regression test

henrykironde:


test/test_regression_cli.py
```diff
@@ -94,8 +116,18 @@ def check_postgres_regression(self, dataset, known_md5):
         current_md5 = getmd5(unixfileformat("output_file")) 
         assert current_md5 == known_md5
 
+
+class DownloadRegression(TestCase):
+    def check_download_regression(self, dataset, known_md5):
+        """Check for regression for a particular dataset downloaded only"""
+        os.system("retriever download %s -p raw_data/%s" %(dataset,dataset))
```
ethanwhite: We should use the newer `.format` instead of `%` to help set us up for the shift to Python 3 at some point. This line would become:

`"retriever download {0} -p raw_data/{0}".format(dataset)`

The indexing here (i.e., the use of `{0}` instead of `{}`) lets us reuse the first value inside `.format()` instead of listing it twice. Make a similar change to the next line.

test/test_regression_cli.py
```diff
@@ -36,6 +46,18 @@ def getmd5(filename):
         sum.update(line)
     return sum.hexdigest()
 
+def getmd5d(directoryename):
```
ethanwhite: Let's expand this function name a little (on the first read my brain kept thinking this was the same as `getmd5`). Even `getmd5dir` would be fine.

Also, here and below it should be `directoryname` instead of `directoryename`.

ethanwhite:
Looks great!

Since the relevant commit is on built on the earlier gui commit, I'll need to wait until we get #372 tested and merged before merging this. In the future if you go back to master before creating a new branch, each addition will be independent of the others and make merging things easier. So, the general workflow would be:

1. Checkout master
2. Create a new branch
3. Add a new feature or fix a bug
4. Add, commit, push, PR
5. Checkout master
6. Create a new branch
7. Add a new feature or fix a bug
8. Add, commit, push, PR

ethanwhite:
It occurred to me that given the changes in the other branch that this won't actually merge cleanly anyway, so I cherry picked this commit and the others that were ready to go into master.

ethanwhite:
Merged in ac09a9e2c4d1b3d2a3ddfe9bc288fb40761b702e.

PR #378, opened 2016-01-19
Check for version 1.6 before importing DownloadOnlyTemplate in MammalSuperTree

### LINKED ISSUE #208: v1.6.0 will break if download scripts are added to version.txt in master
We have now done this twice (see #180 and #199).

In v1.6.0 `__init__.py` line 16: `MASTER = True`. This results in the retriever always checking `master` for `version.txt` and discovering scripts that it doesn't know how to handle. In the future, the retriever will handle this gracefully thanks to #204, but it's unclear how we should go about introducing the download only functionality since it will break a number of existing installations.

ethanwhite:
If v1.6 is running use a dummy script to avoid importing a template that
doesn't exist in that version.

This will allow us to keep updated versions of all scripts available without breaking
existing installations of v1.6. This resulted because MASTER was set to TRUE accidentally
in the v1.6 release. See https://github.com/weecology/retriever/issues/208 for details.

Fixes #208.

### DIFF
--- a/scripts/MammalSuperTree.py
+++ b/scripts/MammalSuperTree.py
@@ -1,9 +1,22 @@
 #retriever
-from retriever.lib.templates import DownloadOnlyTemplate
-
-SCRIPT = DownloadOnlyTemplate(name="Mammal Super Tree",
-                              shortname='mammsupertree',
-                              ref='http://doi.org/10.1111/j.1461-0248.2009.01307.x',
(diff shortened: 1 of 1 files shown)


ethanwhite:
@henrykironde - you should test the version of MammalSuperTree.py in this PR against v1.6 of the retriever by installing v1.6 and then downloading this version of the script into `.retriever/scripts`. If `retriever ls` doesn't error we should be good to go.

ethanwhite:
Just pushed a fix for an issue identified by @henrykironde - the check should be against 'v1.6' and 'v1.6.0', not against '1.6' and '1.6.0'.

henrykironde:
Tested on version 1.6 and 1.7

PR #379, opened 2016-01-19
Error if installing, not just downloading, download only script from template

ethanwhite:


ethanwhite:
We currently do this check in the .py download only scripts, but not in the those using the DownloadOnlyTemplate.

henrykironde:
Looks fine, I will go ahead and Merge

PR #381, opened 2016-01-26
add Bioclim missing hdr files.  fixes #380

### LINKED ISSUE #380: Bioclim data missing hdr files when downloaded to path
retriever download Bioclim -p

henrykironde:


ethanwhite:
Good catch and fix @henrykironde!

I think we can make this code a little cleaner.

For starters we don't need the index, so we can drop the use of `enumerate` and `idx`. I also think that `ext` will be clearer than `word` for describing the file extension.

In addition I tend to think that nested for loops are a bit difficult to read in list comprehensions, so I'd recommend either writing that out as an actual nested for loop or using itertools to get the combinations for you using something like:

```
file_names = ["bil{0}.{1}".format(file_num, ext) for file_num, ext in
              itertools.product(range(1,20), ['bil', 'hdr'])]
```

ethanwhite:
Looks great. Thanks @henrykironde.

PR #382, opened 2016-01-26
Bump version to 1.8.0-dev

ethanwhite:
Now that #378 is in we should be able to go ahead and update both the version number and `version.txt`. Updating the version number should fix the annoying behavior where when master is installed running the retriever from outside of the repository directory results in the use of the previous versions scripts.

In general we should be the version to `1.x.0-dev` immediately following every release.

While #378 should have fixed the issues that could cause this PR to break previous releases, we should immediately test both 1.6, 1.7, and master after this PR is merged.


henrykironde:
I will go ahead and test

PR #385, opened 2016-01-29
replace icon file with a multi-layer .ico file

henrykironde:
a multi-layer icon can be obtained using the tools referenced
ref: use Greenfish Icon Editor or imagemagick (convert)
arrangement of sizes matters.
in this commit the size starts from biggest to smallest.

displaying but still blurred


ethanwhite:
Moving in the right direction at least :smile:. Thanks!

Issue #392, opened 2016-02-04
Build failure on mac


davharris:
When I clone b6b0f74003f333a39030cfe0cca24ffdf1da37b6 and run `bash build_mac`, I get the following results

```
arch: posix_spawnp: python: Bad CPU type in executable
arch: posix_spawnp: python: Bad CPU type in executable
mv: rename dist to mac: No such file or directory
rm: mac/*.egg: No such file or directory
ditto: can't get real path for source 'mac/retriever.app'
mv: rename mac/retriever2.app to mac/retriever.app: No such file or directory
build_mac: line 11: cd: mac: No such file or directory
	zip warning: name not matched: retriever.app

zip error: Nothing to do! (try: zip -r retriever.zip . -i retriever.app)
```

Some quick Googling suggests that it could be a 32-bit versus 64-bit issue dating back ~~10~~ 4 years, but I haven't found much that's concrete.

ethanwhite:
Fixed in #393.

PR #393, opened 2016-02-04
update mac arch

davharris:
Partial solution to #392 

This still gets a lot of errors, but it compiles. Currently installing dependencies to see how many of the errors/warnings I can make disappear.

### DIFF
--- a/build_mac
+++ b/build_mac
@@ -1,11 +1,11 @@
 export DYLD_LIBRARY_PATH=/usr/local/mysql/lib/
 sudo rm -rf mac
-sudo arch -x86_64 python setup.py install
-sudo arch -x86_64 python setup.py py2app
+sudo  setup.py install
+sudo python setup.py py2app
 sudo mv dist mac
 sudo rm mac/*.egg
 sudo rm -rf mac/scripts
-sudo ditto --rsrc --arch x86_64 mac/retriever.app mac/retriever2.app
+sudo ditto --rsrc x86_64 mac/retriever.app mac/retriever2.app
 sudo rm -rf mac/retriever.app
 sudo mv mac/retriever2.app mac/retriever.app
 cd mac
--- a/build_mac
+++ b/build_mac
@@ -1,6 +1,6 @@
 export DYLD_LIBRARY_PATH=/usr/local/mysql/lib/
 sudo rm -rf mac
-sudo  setup.py install
+sudo python setup.py install
 sudo python setup.py py2app
 sudo mv dist mac
 sudo rm mac/*.egg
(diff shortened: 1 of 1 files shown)


davharris:
Update: Running `pip install py2app` seems to grab some necessary dependencies that aren't available if wxpython isn't already installed.  In particular, I seemed to be missing modulegraph and macholib.

davharris:
That seems to solve it.  Not sure why py2app didn't have all its dependencies, but adding them allows it to compile once the --arch flags have been reset.

By the way, how do I uninstall retriever to make sure that fresh installs work?

ethanwhite:
Thanks Dave. A little history on this:

1\. The i386 arch flag was originally added to force the use of 32-bit architecture so that the .app would run successfully on older macs still using 32-bit architectures (https://github.com/weecology/retriever/commit/bb35adc26beb986e433175a935c463b39fe2a4b8). When we did this in 2012 there will still lots of 32-bit only macs out there. I'm sure this has improved over the last 3.5 years.

2a. In order to be able to change the architecture you need a 64/32-bit version of Python. This is what you get if you install from Python's website, but not what you get from certain other approaches, so depending on where your Python came from this may explain the error.

2b. On a more general note, getting an install on OS X that will lead to a .app file that will sucessfully work on other computers is very finicky. The instructions for doing so are [on the wiki](https://github.com/weecology/retriever/wiki/Mac-OSX-Build).

3\. I'm totally open to dropping the i386 arch if we think that there basically aren't any 32-bit Macs out there anymore. It would be interesting to know what the last 32-bit only OS X release was. If we go this route we can just drop the `--arch` flag and argument entirely because OS X defaults to 64-bit.

davharris:
Okay, thanks for the update.  I got Python from Anaconda. I think they no longer bundle the 32-bit version with their 64-bit versions. Since Anaconda is generally what you recommend to students, I think this means that 32-bit Python is going to be pretty rare for your user base.

[If I'm reading things correctly](http://apple.stackexchange.com/a/99644/15086), Apple only sold 32-bit x86 Macs between 2005 and 2006 (slightly longer for the mac mini). Anything in the last 9-10 years would be 64-bit, and anything older than about 12 years would be PowerPC.

I suppose we could use an `if` statement to choose the -arch flag.

ethanwhite:
Thanks for the info @davharris! Given the dates you mention I'm happy to switch to 64-bit builds of the `.app` file if @henrykironde doesn't have any concerns. An `if` statement won't really work here because this relates to building the distributed binary which will either be in 64-bit or 32-bit and won't be able to adapt to the OS.

If it sounds good to @henrykironde I'd suggest that @davharris update this PR by removing the `--arch` flags and arguments.

ethanwhite:
Thanks for this @davharris. We're trying to run down all of the issues with the current Mac build before merging anything, but just wanted to let you know we're working on it and that we really appreciate the help.

henrykironde:
Great work @davharris for this PR and thanks @MarconiS for the tests

PR #408, opened 2016-02-12
Check if a file to be downloaded already exists the destination location

ethanwhite:
When using the `download` engine and installing the data somewhere other than the working directory,
the retriever didn't check to see if the file was already there. This change checks the destination and
if the file is already there it does not download/move the file and reports that it already exists.

This also required changing the `download` engines regression tests so that they were still testing
the main behavior of the code.


ethanwhite:
No rush on this one. I had just finished it so I'm putting it in before I forget.

Congrats again on getting your first release out the door!

PR #423, opened 2016-02-26
adding available dataset documentation fixes #420

### LINKED ISSUE #420: Add list of all available datasets to the documentation
It would be nice to have a page of the documentation that is an auto-generated list of all of the currently available datasets. Each item would have the text from the `name` attribute and be linked to the `url` or `ref` attribute if available.

henrykironde:


docs/conf.py
```diff
@@ -1,5 +1,34 @@
...
+==================
+
+
+"""
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
+dataset_file.write(dataset_file_tittle)
+for script in script_list:
+    s_ref = ""
```
ethanwhite: I don't think you need this given the else on lines 26/27.

docs/conf.py
```diff
@@ -1,5 +1,34 @@
 import sys, os
 from retriever import VERSION,COPYRIGHT
+from retriever.lib.repository import check_for_updates
+from retriever import SCRIPT_LIST
+
+# Create the .rst file for the available datasets
+num_counter = 1
+dataset_file = open("datasets.rst", "w")
+dataset_file_tittle = """
```
ethanwhite: "tittle" -> "title"

docs/conf.py
```diff
@@ -1,5 +1,34 @@
...
+==================
+Datasets Available
+==================
+
+
+"""
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
+dataset_file.write(dataset_file_tittle)
```
ethanwhite: "tittle" -> "title"

ethanwhite:
Instead of incrementing a counter like `num_counter` this is a create place to use [`enumerate`](https://docs.python.org/2/library/functions.html#enumerate). So instead of 

```
num_counter = 1
for script in script_list:
    ...
    num_counter += 1
```

you can use:

```
for num_counter, script in enumerate(script_list):
    ...
```

Since it's not a "counter" any more I'd probably call it something else, maybe `script_num`?

Also, `s_ref` is a little hard to understand as a variable name in this context. I'd use something like `reference_link` or `ref_url` or even just `url`.

Other than that, everything looks great.

docs/conf.py
```diff
@@ -1,5 +1,36 @@
...
+Datasets Available
+==================
+
+
+"""
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
```
ethanwhite: Duplicates of 15 and 16.

docs/conf.py
```diff
@@ -1,5 +1,36 @@
-import sys, os
 from retriever import VERSION,COPYRIGHT
+from retriever.lib.repository import check_for_updates
+from retriever import SCRIPT_LIST
+
+# Create the .rst file for the available datasets
+script_num = 1
+datasetfile = open("datasets.rst", "w")
+datasetfile_tittle = """
```
ethanwhite: This should be "datasetfile_title"

docs/conf.py
```diff
@@ -1,5 +1,36 @@
...
+
+"""
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
+# write the title of dataset rst file
+datasetfile.write(datasetfile_tittle)
```
ethanwhite: Change to "datasetfile_title" to match above spelling fix.

docs/conf.py
```diff
@@ -1,5 +1,36 @@
...
+
+# get info from the scripts
+for script_num, script in enumerate(script_list, start=1):
+    reference_link = ""
+    if script.ref and script.ref is not "":
```
ethanwhite: We don't need the second part of the conditional because `""` evaluates to `False`. So the `if script.ref` will take care of this. Sorry I didn't notice this last time.

docs/conf.py
```diff
@@ -1,5 +1,36 @@
...
+
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
+# write the title of dataset rst file
+datasetfile.write(datasetfile_tittle)
+
+# get info from the scripts
+for script_num, script in enumerate(script_list, start=1):
+    reference_link = ""
```
ethanwhite: Now that I'm reading this again I think it's better to have this in an `else` at the end and not here at the top.

ethanwhite:
Great changes overall. I caught a couple of things that I missed or didn't get quite right in the first review.

The other issue with the current PR is that it appears to had removed `docs/MacOsxBuild.rst`, which isn't part of these changes unless I'm missing something.

ethanwhite:
Just merged the updated PR #427. Closing.

Issue #312, opened 2015-05-06
optimizing add_to_table?


davharris:
I don't speak much Python and I'm not terribly familiar with this code, but it looks like lines are fed one-at-a-time to the database, with [lots of checks on each line](https://github.com/weecology/retriever/blob/58678486a8d0f8b3909ec0cad0e14dfb0095f7ef/lib/engine.py#L86-L130).

Without having profiled anything myself, it seems like it could be much faster if the database got larger chunks and if some lines whose results don't change were moved outside the loop.  For most of the data sets, this probably won't make a practical difference, but it adds up when there are millions of rows (e.g. BBS50).

What do you think?

davharris:
I did a bit more reading, and I seem to have confirmed that retriever's insert operation is substantially slower than it could be.

The graph in [this stack overflow question](https://stackoverflow.com/questions/15778716/sqlite-insert-speed-slows-as-number-of-records-increases-due-to-an-index) says that sqlite can insert close to 100k rows per second.  I seem to be getting closer to 1k rows per second on my laptop with retriever.

So it seems like a bit of work on the engine could save ~60 minutes when downloading BBS50 (and even more on some of the larger databases).  Speeding things up could also make it easier to run integration tests and catch bugs more easily.

ethanwhite:
Thanks for looking into this @davharris. Running big datasets line by line is definitely a little on the slow side and if we can do something to speed that up it would be great. The challenge is to do this in a way that works for all datasets including the ones that need to be cleaned since the code can't know which lines it needs to cleanup in advance. But, yes, definitely if we could make some improvements here that would be great.

For some of the other database management systems we do take advantage of their bulk import features where available and appropriate for the dataset (i.e., something that doesn't need cleanup within individual files). So, if you're looking for speed with things like BBS50 I'd recommend considering PostgreSQL as it will take <30 seconds not including the download. MySQL can be similarly fast, though getting it configured to allow bulk imports can take a little more work.

ethanwhite:
FYI, as of bc182e42d6e73c1a20a71ce6ad83604c4b463084 bulk inserting now works with SQLite and install BBS50 into SQLite now takes <60 seconds not including the download.

The changes in bc182e42d6e73c1a20a71ce6ad83604c4b463084 also point the route forward for optimizing `add_to_table` more generally.

henrykironde:
Thanks to @Deborah-Digges and @davharris 
We believe we have achieved the milestone for this issue and we shall go ahead and close.

[...]

PR #424, opened 2016-02-27
Add high speed bulk importing for SQLite

ethanwhite:
For files that do not need to be cleaned this adds high speed bulk importing
equivalent to that already available in MySQL and PostgreSQL.

This is accomplished by splitting the file up into intermediate sized chunks
(large enough to speed things up, small enough to consistently fit in memory)
and then inserting them using `executemany`.

This results in over a 10x speed up for BBS, and it's probably >20x since BBS
has several tables that can't be bulk inserted.

### DIFF
--- a/engines/sqlite.py
+++ b/engines/sqlite.py
@@ -35,6 +35,59 @@ class engine(Engine):
         """Escapes single quotes in the line"""
         return line.replace("'", "''")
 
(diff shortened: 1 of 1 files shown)


ethanwhite:
This approach might also allow us to speed up the standard inserts as well. See #312.

PR #426, opened 2016-02-27
Shorten and simplify ecoretriever docs

ethanwhite:
These don't need to be as comprehensive here so I've pulled some extra
detail and edited them for length a bit.


ethanwhite:
Note, I don't really know `rst` so definitely give this a read to make sure I didn't break links, etc.

PR #427, opened 2016-02-29
add available list of datasets in the documentation.

### LINKED ISSUE #420: Add list of all available datasets to the documentation

henrykironde:
fixes #420

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,33 @@
-import sys, os
 from retriever import VERSION,COPYRIGHT
+from retriever.lib.repository import check_for_updates
+from retriever import SCRIPT_LIST
+
+# Create the .rst file for the available datasets
+datasetfile = open("datasets.rst", "w")
+datasetfile_title = """
+==================
+Datasets Available
+==================
+
+
+"""
+check_for_updates(graphical=False)
+script_list = SCRIPT_LIST()
+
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -1,5 +1,33 @@
-import sys, os
 from retriever import VERSION,COPYRIGHT
+from retriever.lib.repository import check_for_updates
+from retriever import SCRIPT_LIST
+
+# Create the .rst file for the available datasets
+script_num = 1
```
ethanwhite: This line doesn't get used so we can remove it.

ethanwhite:
Almost there. I caught one extra line that we don't need (see in-line comment). The one other thing is that I think including the blank case for `reference_link` is clearer as an `else`, to make clear that this is what we fall back to if nothing else is available. I.e.,

```
for script_num, script in enumerate(script_list, start=1):
    if script.ref.strip():
        reference_link = script.ref
    elif bool(script.urls.values()):
        reference_link = script.urls.values()[0].rpartition('/')[0]
    else:
        reference_link = ""
```

ethanwhite:
Looks great! Thanks.

PR #429, opened 2016-02-29
add LEDA data. fixes #384

### LINKED ISSUE #384: Add LEDA data
http://www.uni-oldenburg.de/en/biology/landeco/research/projects/leda/

henrykironde:


ethanwhite:
You were right. That's a tricky one 😄. Let's definitely chat tomorrow and figure out if there's anyway to simplify it. The one thing I noticed on the first read through is that the conditional on Line 58 could use some parentheses to make sure the logic of how to combine the conditions is clear. If we flip the conditions around we should also be able to get rid of the `if X: pass else:` part of the code and just have an `if not X:` instead.

PR #430, opened 2016-02-29
add DELL data

henrykironde:
This references f925823d740958b562a92bc17f4724883ffcd318 by @carol-rowe666, 
It was redeveloped because the source file has some null byte values. In case, the file is fixed by the author, we can roll back to this simple script.

```
name:Thermal dependence of biological traits. - Dell, et al., 2013
shortname:ThermalTraitDependence 
description: Anthony I. Dell. 2013. The thermal dependence of biological traits. Ecology 94:1205.
# tables
table: TempTrait, http://esapubs.org/archive/ecol/E094/108/TempTrait_001.txt
*nulls: 'NA'
```

### DIFF
--- a/scripts/EA_Dell.py
+++ b/scripts/EA_Dell.py
@@ -1,0 +1,36 @@
+#retriever
+"""EcoData Retriever script DELL dataset"""
+from retriever.lib.models import Table, Cleanup, correct_invalid_value
+from retriever.lib.templates import Script
+
+
+class main(Script):
+    def __init__(self, **kwargs):
+        Script.__init__(self, **kwargs)
+        self.name = "Thermal dependence of biological traits, 2013",
+        self.shortname = "DELL"
+        self.ref = "http://esapubs.org/archive/ecol/E094/108/"
+        self.urls = {'TempTrait': 'http://esapubs.org/archive/ecol/E094/108/TempTrait_001.txt'}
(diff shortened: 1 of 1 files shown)


ethanwhite:
We may want to think about implementing the suggested use of [`tempfile`](https://docs.python.org/2/library/tempfile.html) from #75 in this case, but I'm also OK with just going with this as is and adding a note about this script to #75 to cleanup later. Everything else looks good.

Let me know if you want me to merge or if you want to take a look at `tempfile` first.

henrykironde:
Hi Ethan,

I will take a look, let's Merge later.

On Monday, February 29, 2016, Ethan White <notifications@github.com> wrote:

> We may want to think about implementing the suggested use of tempfile
> <https://docs.python.org/2/library/tempfile.html> from #75
> <https://github.com/weecology/retriever/issues/75> in this case, but I'm
> also OK with just going with this as is and adding a note about this script
> to #75 <https://github.com/weecology/retriever/issues/75> to cleanup
> later. Everything else looks good.
>
> Let me know if you want me to merge or if you want to take a look at
> tempfile first.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/weecology/retriever/pull/430#issuecomment-190500081>.
>


-- 

Regards

Henry S.

ethanwhite:
Also, to give the original contributor credit for getting this started, let's have this PR include her commit first and then yours as a rewrite of it.

henrykironde:
@ethanwhite, I think we should merge this. Changing to tempfile will affect many scripts, not only this script and this will be introduced as we move to json package and python3.

ethanwhite:
@henrykironde sounds good.

Replaced with #479 to integrate history from #227.

PR #435, opened 2016-03-01
Convert numeric nulls when also converting string nulls

ethanwhite:
correct_invalid_value was failing to convert numeric nulls when string nulls
were also present to be converted. So, if a table contained both 'NA' and -999
as null values the -999 would not be properly converted.

This appears to only effect EA_barnes2008.py among the current scripts.


henrykironde:
looks fine

PR #442, opened 2016-03-08
Use status bar for retriever update

### LINKED ISSUE #397: Use status bar for script download progress
Instead of printing out the name of each script being downloaded, use a status bar to indicate the progress of downloading all scripts.

ghoshbishakh:
Fixes #397
Use a progress bar while scripts are being downloaded for retriever update

### DIFF
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -20,12 +20,7 @@ def download_from_repository(filepath, newpath, repo=REPOSITORY):
     """Downloads the latest version of a file from the repository."""
     try:
         filename = filepath.split('/')[-1]
-
-        def reporthook(a, b, c):
-            print "%3.1f-%s\n" % (min(100, float(a * b) / c * 100), filename),
-            sys.stdout.flush()
-
-        urllib.urlretrieve(repo + filepath, newpath, reporthook=reporthook)
+        urllib.urlretrieve(repo + filepath, newpath)
     except:
         raise
         pass
@@ -81,7 +76,29 @@ def check_for_updates(graphical=False):
     if graphical:
         splash.Hide()
         sys.stdout = sys.__stdout__
-    print "The retriever is up-to-date"
+    print "\nThe retriever is up-to-date"
+
+
+def update_progressbar(progress):
+    """Show progressbar
+
+    Takes a number between 0 and 1 to indicate progress from 0 to 100%.
+
+    """
+    # Try to set the bar_length according to the console size
+    try:
+        rows, columns = os.popen('stty size', 'r').read().split()
+        bar_length = int(columns) - 35
+        if(not (bar_length > 1)):
+            bar_length = 20
+    except:
+        # Default value if determination of console size fails
+        bar_length = 20
+    block = int(round(bar_length*progress))
+    text = "\rDownload Progress: [{0}] {1:.2f}%".format(
+        "#"*block + "-"*(bar_length-block), progress*100)
+    sys.stdout.write(text)
+    sys.stdout.flush()
 
 
 class InitThread(Thread):
@@ -158,14 +175,18 @@ class InitThread(Thread):
             # read scripts from the repository and the checksums from the
             # version.txt
             scripts = []
+            print "Downloading scripts..."
             for line in version_file:
                 scripts.append(line.strip('\n').split(','))
+
+            total_script_count = len(scripts)
 
             # create script directory if not available
             if not os.path.isdir(SCRIPT_WRITE_PATH):
                 os.makedirs(SCRIPT_WRITE_PATH)
 
-            for script in scripts:
+            update_progressbar(0.0/float(total_script_count))
+            for index, script in enumerate(scripts):
                 script_name = script[0]
                 if len(script) > 1:
                     script_version = script[1]
@@ -175,7 +196,6 @@ class InitThread(Thread):
                 path_script_name = os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name))
 
                 if not file_exists(path_script_name):
-                    print "Downloading script: " + script_name
                     download_from_repository("scripts/" + script_name,
                                              os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
 
@@ -201,6 +221,7 @@ class InitThread(Thread):
                     except Exception as e:
                         print e
                         pass
+                update_progressbar(float(index + 1)/float(total_script_count))
         except:
             raise
             return


lib/repository.py
```diff
@@ -81,7 +82,39 @@ def flush(self):
...
+        progress = 0
+        status = "error: progress var must be float\r\n"
+    if progress < 0:
+        progress = 0
+        status = "Halt...\r\n"
+    if progress >= 1:
+        progress = 1
+        status = "Done...\r\n"
+    block = int(round(barWidth*progress))
+    text = "\r[{0}] {1:.2f}%".format("#"*block + "-"*(barWidth-block),
```
henrykironde: @ghoshbishakh great work. just a few things to change.
I think we can add Downlaod progress ` text = "\rDownload Progress: [{0}] `

lib/repository.py
```diff
@@ -81,7 +82,39 @@ def flush(self):
...
         splash.Hide()
         sys.stdout = sys.__stdout__
-    print "The retriever is up-to-date"
+    print "\nThe retriever is up-to-date"
+
+
+def update_progressbar(progress):
+    """Shows progressbar given progress in float
+    Example: 0.5 for 50%
+    """
```
henrykironde: In the doc string we can remove this `Example:` 0.5 for `50%`
or we wan instead explain how this comes up
 
 """Shows progressbar given progress in float

accepts a a number  between 0 and 1 and it's converted to a float.
a number under 0 represents a 'halt'.
a number at 1 or greater represents 100%
"""

ghoshbishakh:
@henrykironde thank you for taking a look. :smile:  I am updating my PR.

ghoshbishakh:
@henrykironde  I hope that is better

lib/repository.py
```diff
@@ -81,7 +82,39 @@ def flush(self):
...
+    try:
+        rows, columns = os.popen('stty size', 'r').read().split()
+        barWidth = int(columns) - 20
+        if(not (barWidth > 1)):
+            barWidth = 20
+    except:
+        pass
+    status = ""
+    if isinstance(progress, int):
+        progress = float(progress)
```
henrykironde: how about we change the logic and assume that given a number just convert to float, and this case we may not need to check if its int.
so before line 102, we can just say ` progress = float(progress)`

lib/repository.py
```diff
@@ -175,7 +212,7 @@ def run(self):
                 path_script_name = os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name))
 
                 if not file_exists(path_script_name):
-                    print "Downloading script: " + script_name
+                    # print "Downloading script: " + script_name
```
henrykironde: lets remove this `# print "Downloading script: " + script_name`

ghoshbishakh:
good call :smile:  will update.

henrykironde:
Modify this to change the **length** of the progress bar
barWidth = 20

if we go with bar_length or barLenth , people reading the code later may understand easily since the comment is talking about length

ghoshbishakh:
ok sure but actually I am dynamically setting the length of the progressbar later according to the console width. So I think its better to comment that as default bar width?

henrykironde:
ok that works

lib/repository.py
```diff
@@ -81,7 +82,41 @@ def flush(self):
     if graphical:
         splash.Hide()
         sys.stdout = sys.__stdout__
-    print "The retriever is up-to-date"
+    print "\nThe retriever is up-to-date"
+
+
+def update_progressbar(progress):
+    """Shows progressbar given progress in float
+    accepts a a number between 0 and 1 and it's converted to a float.
```
henrykironde: space after the first line

ghoshbishakh: sorry I did not get that

ghoshbishakh: ah in the docstring ok!

ghoshbishakh:
Did I miss anything?

lib/repository.py
```diff
@@ -81,7 +82,39 @@ def flush(self):
...
+    bar_length = 20
+    # Try to set the bar_length according to the console size
+    try:
+        rows, columns = os.popen('stty size', 'r').read().split()
+        bar_length = int(columns) - 35
+        if(not (bar_length > 1)):
+            bar_length = 20
+    except:
+        pass
+    status = ""
```
henrykironde: if we are creating `status`, then I guess we may include it in line   ` text = "\rDownload Progress:....` 
or we can remove the creation of `status`

lib/repository.py
```diff
@@ -81,7 +82,39 @@ def flush(self):
...
+
+    accepts a a number between 0 and 1 and it's converted to a float.
+    a number under 0 represents a 'halt'.
+    a number at 1 or greater represents 100%
+    """
+    # Modify this to change the default length of the progress bar
+    bar_length = 20
+    # Try to set the bar_length according to the console size
+    try:
+        rows, columns = os.popen('stty size', 'r').read().split()
```
henrykironde: do we need `row`?

ghoshbishakh: @henrykironde actually that returns a tuple thats why.
I can take index 2 though like [1].
Should I do that?

ghoshbishakh:
@henrykironde cat you take a look at it now?

henrykironde:
ok

ethanwhite:
Thanks for this @ghoshbishakh. Overall it looks great.

I have a couple of thoughts/recommendations/questions:

1. I think we can remove the `reporthook` function entirely and then get rid of the associated optional argument in the call to `urllib.urlretriever`.

2. The docstring for `update_progressbar` could cleaned up a little (for part an explanation of 
part of the cleanup see number 4 below). E.g.,

    ```
    """Show progressbar

    Takes a number between 0 and 1 to indicate progress from 0 to 100%.

    """
    ```

3. Is the initial declaration of `bar_length` there for cases when determining the
size of the terminal doesn't work properly? If so then it would be clearer to
include this in the `except`. If not, then could you explain a little more about
what it's accomplishing.

4. Since `update_progressbar` shouldn't be getting values < 0 or > 1 I think we can
probably get rid of lines 105-112, and 116-117 unless I'm
missing something. If you were thinking about checking for bad values we could use
an `assert` statement and throw an exception if the value isn't between 0 and 1.

5. I don't understand why it's necessary to call `update_progressbar` at both the
beginning and end of the main `for` loop. Could you explain a little be more about
why we need both of them.

ghoshbishakh:
@ethanwhite thanks for the suggestions :smile: 

1.  I am removing the reporthook

2.  Updated docstring

3.  Yes it was there for cases where determination of console size fails. I have put it in the except.

4.  I was thinking that later may be we could use statuses like halt for any exception or like that. But yes its better to keep it simple

5. Actually the 0% is not displayed if we don't call it first once. So I will call it outside the for loop first then so that it isn't repeated uselessly?

ghoshbishakh:
@ethanwhite I have pushed with the required changes

henrykironde:
great work @ghoshbishakh

ghoshbishakh:
@henrykironde  thanks! :smile: 
Can you guide me a bit here:
https://github.com/numfocus/gsoc/issues/105

PR #444, opened 2016-03-15
Parsed script files to json

goelakash:
@henrykironde @ethanwhite 
I modified the `compile.py` parser to save the `.script` files to `json` in a temp folder for review purposes.

Some ideas to follow `datapackage.json` standard might be-

1. 'ref' -> 'homepage'
2. 'urls' -> 'resources' (with name and url of tables as sub-keys)

PR #445, opened 2016-03-18
Function to parse script files to JSON

goelakash:
#444 
Also, I don't suppose I could explicitly mention that we are moving to JSON until a tool is created that allows the user to create a JSON based script. Because the current scripts doc also doesn't mention anywhere that the files are being compiled in python readable format.
@henrykironde @ethanwhite

### DIFF
--- a/lib/compile.py
+++ b/lib/compile.py
@@ -1,3 +1,5 @@
+import json
+
 script_templates = {
     "default": """#retriever
 from retriever.lib.templates import BasicTextTemplate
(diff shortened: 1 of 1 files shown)


goelakash:
#444

goelakash:
#444

henrykironde:
@goelakash  thanks for the PR, I am going to delete/close this PR unless you suggest otherwise. 

1) could you add this file to the [branch goelakash:datapackage.json-format](https://github.com/weecology/retriever/pull/444)

2) Push changes to the old branch to update the old PR, you may delete the old file or leave it.

3) Right now we should not worry about the errors because we are developing the design and concepts. The Docs will definitely be the last pieces to put together 

4) At this point I suggest that you also focus on writing the proposal, looking at the way forward as mentioned in the [project idea](https://github.com/numfocus/gsoc/blob/master/2016/ideas-list-ecodata-retriever.md) and in the [PR comments](https://github.com/weecology/retriever/pull/444) by @ethanwhite

goelakash:
I realised that this function should be in a separate file, as it doesn't compile the JSON. I added those changes in the other branch. Closing this.

PR #446, opened 2016-03-21
add py.test and remove nose test fixes #404

### LINKED ISSUE #404: Switch to pytest for testing
nose is no longer being developed and the consensus replacement is [pytest](http://pytest.org/latest/).

henrykironde:
no environment variables explicitly set (tox.ini). [ref](https://github.com/codecov/example-python)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,14 @@ install:
   - pip install -r requirements.txt --use-mirrors
   - python setup.py -q install
   - pip install codecov
-  - pip install nose-cov
+  - pip install pytest-cov
+  - pip install pytest-xdist      
+
 
 # command to run tests
-script: nosetests --with-cov
+
+script:
+  - py.test --cov=retriever 
 
 # database setup
 before_script:


ethanwhite:
So does this pass mean we're all set for merge. That was more straight forward than I'd been assuming.

henrykironde:
@ethanwhite, we cant merge. 
The tests work fine however the coverage report is what we are not getting right. Codecov is not performing as expected.
I am assuming two things,

1. The path setting is not right
2. or we are right and pytest only reports the files that have been hit (which case I assume is a false hypothesis) 

This is the [report](https://codecov.io/github/weecology/retriever?pr=446) to our current PR.
 and basically only the lib folder is shown which is not the case with the [old reports](https://codecov.io/github/weecology/retriever?pr=445) that show the whole repo tree

ethanwhite:
@henrykironde - got it. Let me know when you think everything is working and I'll take a look.

henrykironde:
pytest runs well. 
The cov report here is a little different from the cov report when using nose.
With pytest , they do not report files that are not hit by the tests and hence some files don't show up.

ethanwhite:
Is the addition to the slack secure line intentional? Looks like it might be an accidental copy/paste.

ethanwhite:
Thanks for switching that back @henrykironde. Let's go ahead and officially make the move to pytest.

henrykironde:
@ethanwhite copy and paste never a good way :+1:

PR #449, opened 2016-03-22
Futurize all .py files

goelakash:
See this page for additional details: [Automatic conversion](http://python-future.org/automatic_conversion.html)

Pending Python 3 tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 sudo: false
 language: python
 python:
-  - "2.7"
+  - "2.7.10"
 addons:
   postgresql: "9.1"
 
 services:
   - postgresql
 install:
-  - pip install -r requirements.txt --use-mirrors
+  - pip install -r requirements.txt
   - python setup.py -q install
   - pip install codecov
   - pip install nose-cov
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 sudo: false
 language: python
 python:
-  - "2.7.10"
+  - "2.7"
 addons:
   postgresql: "9.1"
 
 services:
   - postgresql
 install:
-  - pip install -r requirements.txt
+  - pip install -r requirements.txt --use-mirrors
   - python setup.py -q install
   - pip install codecov
   - pip install nose-cov
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ retriever.egg-info
 apidocs
 .DS_Store
 .cache/*
+*.py.bak
+test/raw_data/*
+.coverage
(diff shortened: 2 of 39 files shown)


goelakash:
@henrykironde @ethanwhite Only `lib`/`*.py` files remain. I am running into some HTTP errors, probably because of `urllib` inconsistencies.

henrykironde:
that's because of urllib.request vs  urllib

henrykironde:
We tend to keep the number of commit small, so that the history is clean.
So for each issue, I would suggest, create it's own branch.

For the case of testing using Travis, if you sign into Travis, you can add your master Repo so that every time you commit, you commit to your master for testing purposes

ethanwhite:
@henrykironde makes a good point about minimizing the total number of commits. Since I think of the entire shift to Python 3 as one big thing I'd probably just keep developing however works best (including lots of small commits if you want) and then squash all of the commits together using `rebase` at the end. Happy to help walk you through the squash if that's something you don't know how to do yet.

Also, @goelakash, it's great that you started on this already and I'd recommend adding a note about this PR to your proposal.

goelakash:
@henrykironde Okay, I'll make a travis account for my repo.

@ethanwhite Thanks, I added this PR's link to the proposal. I generally do `git reset --soft HEAD~X && git commit`.

ethanwhite:
> @ethanwhite Thanks, I added this PR's link to the proposal. I generally do git reset --soft HEAD~X && git commit. 

That's fine too for cases where you want to do a full squash.

.travis.yml
```diff
@@ -1,14 +1,14 @@
 sudo: false
 language: python
 python:
-  - "2.7"
+  - "2.7.10"
```
henrykironde: do we need to specify the release?

.travis.yml
```diff
@@ -1,14 +1,14 @@
...
 addons:
   postgresql: "9.1"
 
 services:
   - postgresql
 install:
-  - pip install -r requirements.txt --use-mirrors
+  - pip install -r requirements.txt
```
henrykironde: is there a reason not to use mirrors
(I am not sure if it's still supported or not, but I know some times it helps when installing the requirements )

goelakash: Yeah, these changes can be removed. I just added them because I didn't have a 2.7.9 environment for testing. The build works without these also.

ethanwhite:
@goelakash - looks like we'd lost track of this. Sorry about that. GitHub doesn't notify folks after updated commits have been pushed (like you're last three), so a good thing to do is add a new comment that mentions that you've made the requested changes.

Was there a reason to close this PR other than our lack of activity?

goelakash:
This branch had become incompatible with the master, and I have a similar branch on my fork which is synced with master. So I thought best to close this one.

ethanwhite:
Ah great. Thanks for clarifying.

PR #451, opened 2016-03-24
Add citation for FIA

ethanwhite:
After a question from a user about how to cite FIA I contacted Carol Perry,
one of the FIA folks, and got the official citation folks should be using.

### DIFF
--- a/scripts/fia.py
+++ b/scripts/fia.py
@@ -23,6 +23,7 @@ class main(Script):
                      'species': 'http://apps.fs.fed.us/fiadb-downloads/REF_SPECIES.CSV'}
         self.tags = ["Taxon > Plants", "Spatial Scale > Continental",
                      "Data Type > Observational"]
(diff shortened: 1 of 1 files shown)


henrykironde:
Thanks for the PR @ethanwhite. this is how it looks, I am Curious about the "DATEOFDOWNLOAD"

Citation:   DATEOFDOWNLOAD. Forest Inventory and Analysis Database, St. Paul, MN: U.S. Department of Agriculture, Forest Service, Northern Research Station. [Available only on internet: http://apps.fs.fed.us/fiadb-downloads/datamart.html]

ethanwhite:
Yeah, it's a little confusing. The idea is that the user is supposed to insert the date that they downloaded the data. If you have any ideas about how to communicate that more clearly I'm happy to make changes.

henrykironde:
This communicates it well, was just making sure if it the DATEOFDOWNLOAD or DATEOFUPDATE.
But since it is the date the data was downloaded then we are fine.

PR #457, opened 2016-04-19
minor error message fix

davharris:
in the error message about `The dataset * isn't currently available in the Retriever` the message should propose `retriever ls` not `retriever -ls` (i.e. `l` and `s` aren't flags)


ethanwhite:
Thanks @davharris!

PR #459, opened 2016-04-19
BBS file capitalization (fix #548)

davharris:
fix #548

PR #462, opened 2016-04-20
Change number of header rows in BBS species table to match new data structure

ethanwhite:


ethanwhite:
In combination with #459 this yields what looks like a clean install for me.

henrykironde:
@ethanwhite, I am guessing the same applies to BBS50

ethanwhite:
Yep, sorry I missed that. New PR #471.

PR #466, opened 2016-04-20
add execution for setting auto-comit off when bulk insert

henrykironde:
add execution for setting auto-comit off when bulk insert
remove print exception when bulk insert fails for mysql

To improve the likely hood of bulk insert for mysql and postgres

mysql configuration
===================

set off some of the parameters (mysql_set_autocommit_off)
SET autocommit=0;
SET UNIQUE_CHECKS=0;
SET FOREIGN_KEY_CHECKS=0;
SET sql_log_bin=0;

Once the bulk has been  has been executed
we set the configurations back (mysql_set_autocommit_on)

Postgres
========

BEGIN initiates a transaction block and all statements after a BEGIN command will be executed
in a single transaction until an explicit COMMIT or ROLLBACK is given.

self.execute("BEGIN")
self.execute(statement)
self.execute("COMMIT")

### DIFF
--- a/engines/mysql.py
+++ b/engines/mysql.py
@@ -45,6 +45,10 @@ class engine(Engine):
     def insert_data_from_file(self, filename):
         """Calls MySQL "LOAD DATA LOCAL INFILE" statement to perform a bulk
         insert."""
+
+        mysql_set_autocommit_off = """SET autocommit=0; SET UNIQUE_CHECKS=0; SET FOREIGN_KEY_CHECKS=0; SET sql_log_bin=0;"""
+        mysql_set_autocommit_on = """SET GLOBAL innodb_flush_log_at_trx_commit=1; COMMIT; SET autocommit=1; SET unique_checks=1; SET foreign_key_checks=1;"""
+        
         self.get_cursor()
         ct = len([True for c in self.table.columns if c[1][0][:3] == "ct-"]) != 0
         if (self.table.cleanup.function == no_cleanup and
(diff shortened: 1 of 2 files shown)


engines/mysql.py
```diff
@@ -64,9 +64,20 @@ def insert_data_from_file(self, filename):
 IGNORE """ + str(self.table.header_rows) + """ LINES
 (""" + columns + ")"
             try:
+                self.cursor.execute("""
+                  SET autocommit=0;
```
henrykironde: so since we always have problems with bulk insert, in cases where the user has not set his database system for auto commit off, we can actually set it off here.
we can also set some other parameters off the may not let us do the bulk insert.

engines/postgres.py
```diff
@@ -85,7 +84,9 @@ def insert_data_from_file(self, filename):
 WITH DELIMITER ','
 CSV HEADER"""
             try:
+                self.execute("BEGIN")
```
henrykironde: This is the same idea

henrykironde:
@ethanwhite, let me know what you thing

ethanwhite:
Ah, thanks. So basically these settings increase the likelihood that bulk insert will actually work. Is that the right idea? If so, can you update the commit message to include a bit more of this explanation? Thanks.

engines/mysql.py
```diff
@@ -64,10 +68,13 @@ def insert_data_from_file(self, filename):
...
 (""" + columns + ")"
             try:
+                self.cursor.execute(mysql_set_autocommit_off)
                 self.cursor.execute(statement)
+
+                self.cursor.execute(mysql_set_autocommit_on)
             except Exception as e:
-                print "Failed bulk insert (%s), inserting manually" % e
                 self.disconnect()  # If the execute fails the database connection can get hung up
+                self.cursor.execute(mysql_set_autocommit_on)
```
henrykironde: in case of any exception, remember to set the configurations back ``self.cursor.execute(mysql_set_autocommit_on)``

PR #479, opened 2016-04-25
Add data on trait responses to temperature

ethanwhite:
Adds data from Dell et al. 2013 on the thermal dependence of biological traits.

This PR combines #227 and #430 to maintain the history of the script development
as well as the final solution.


ethanwhite:
Mistake while cleaning up history. Closing.

henrykironde:
reverting since Travis is failing.

PR #480, opened 2016-04-25
Revert "Add data on trait responses to temperature"

henrykironde:
Reverts weecology/retriever#479


ethanwhite:
OK, let's see if this works.

PR #483, opened 2016-04-28
Cleanup of temp files (MoM2003 and a few others)

goelakash:


ethanwhite:
Thanks @goelakash, this looks good. Two things I'd recommend:

1. Since `crosstab_crosstab.txt` and  `simple_csv_simple_csv.txt` are generated in `test_integration.py` we should clean them up there rather than in `test_regression_cli.py`.

2. I would also clean up the commit message just a little. Here's a nice [post on how to write good commit messages](http://chris.beams.io/posts/git-commit/). In this case I'd use something like:

```
Cleanup temporary files after running tests

Delete raw data files that are downloaded, and fake data files
that are generated, after tests are complete.
```

goelakash:
@ethanwhite How does this look?

ethanwhite:
A big step in the right direction. Thanks!

Moving the code for getting rid of the text files let me see that there's a slightly better way to do it. The loop starting on Line 34 in `test_integration.py`, let's us automatically clean up all of those kinds of files that not only exist now, but will exist in the future, so try moving the `os.remove()` inside the for loop and automatically assembly those file names so that each trip through the loop removes that test's file.

Also, it's typically considered good form to avoid punctuation in the first row of the commit message, so I'd go ahead and get rid of the `:`.

ethanwhite:
Thanks @goelakash. Looks great!

Remember in the future to add a comment after you've updated the files to let other folks know about it (Github doesn't send notifications for new/updated commits).

goelakash:
@ethanwhite Okay, will do.

PR #484, opened 2016-05-05
Updated postgresql to 9.2 in .travis.yaml

goelakash:
@henrykironde @ethanwhite


henrykironde:
Since we are moving forward and PostgreSQL-9.2 is more current than PostgreSQL-9.1, we have decided to change to 9.2.

PR #485, opened 2016-05-09
Improve testing

henrykironde:
add csv export module to retriever

from mysql, sqlite and postgres, we can export the data into csv file
using

$ retriever export -h
usage: retriever-script.py export [-h] [--sorted] {mysql,postgres,sqlite} ...

positional arguments:
  {mysql,postgres,sqlite}
                        engine-specific help
    mysql               MySQL
    postgres            PostgreSQL
    sqlite              SQLite

optional arguments:
  -h, --help            show this help message and exit
  --sorted              export csv sorted

with the --sorted flag set we can export csv files sorted
this will help in setting up tests regardless of the server versions

we have also updated the md5s.


lib/tools.py
```diff
@@ -160,3 +260,130 @@ def reset_retriever(scope):
...
+    try:
+        fp = open(input_file, 'r')
+        json.loads(fp.read())
+    except ValueError:
+        return False
+    return True
+
+
+def json2csv(input_file, output_file=None, header_values=None):
+    """Convert Json file to CSV
```
henrykironde: we are yet to add tests for these functions

lib/tools.py
```diff
@@ -43,6 +43,106 @@ def getmd5(lines):
...
+        infile.close()
+        with open(inputfile, 'wb') as output:
+            for line in content.splitlines():
+                output.write(line + '\n')
+        output.close()
+    except IOError as e:
+        print "I/O error({0}): {1} ".format(e.errno, e.strerror)
+    return inputfile
+
+def getmd5_u(file_path):
```
henrykironde: old way method, getmd5.
supports the csv and dowmload engine

engines/__init__.py
```diff
@@ -5,7 +5,7 @@
     "postgres",
     "sqlite",
     "msaccess",
-    "csv",
+    "csvengine",
```
henrykironde: changing csv.py to csvengine. this may be better.

__main__.py
```diff
@@ -117,15 +117,22 @@ def main():
...
-                    dataset.download(engine, debug=debug)
-                    dataset.engine.final_cleanup()
-                except KeyboardInterrupt:
-                    pass
-                except Exception as e:
-                    print e
-                    if debug: raise
+                if args.command == 'export':
+                    for items in dataset.urls.keys():
+                        if hasattr(args, 'sorted') and args.sorted:
```
henrykironde: use engine object

ethanwhite:
Following up on our in-person conversation:

* Since this is basically for internal use we decided to remove it from the
  command line API
* The current system fetches all data into RAM, which could make it difficult to
  use on large datasets for checking for changes (in fact I just crashed my 16
  GB machine exporting vernet). I think the solution you came up with of fetching
  and writing in chunks sounds great.

Other comments:

* Instead of the two nearly identical functions `getmd5_u` and `get_path_md5`,
  we should have a single function `getmd5` function that takes an optional
  argument with a default for 'rU' vs 'rb'. I know this is probably temporary, but
  using the optional argument will be cleaner and we can remove it later if we
  don't need it any more.
* `to_csv` is an existing name for methods in other modules (e.g., Pandas) used
  for exporting to csv, so let's use that instead of `export_2csv` for
  consistency.
* there's quite a bit of overlap in the `export_2csv` functions. Is it possible to
  generalize this in some way?

Overall I'm very excited to be moving towards this approach to testing.

henrykironde:
closed because moved to #511

Issue #210, opened 2014-09-24
Content and formatting in the output of `retriever ls`


cboettig:
As discussed in https://github.com/ropensci/ecoretriever/issues/36#issuecomment-56735242
- [x] It's not trivial to map the available datasets listed to the more
  meaningful descriptions on the website,
  http://ecodataretriever.org/available-data.html. 
- [ ] The naming convention for datasets is unclear, and not obviously scalable as an
  identifier.

dmcglinn:
I agree this is a bit of a problem. It seems like the mapping of names from `ls datasets` to those descriptions at http://www.ecodataretriever.org/available-data.html could be accomplished with a table at http://www.ecodataretriever.org/available-data.html with two columns: 1) the long hyperlinked descriptive name of the dataset, and 2) the shortname that retriever users will refer to the dataset by (e.g., BBS). 

With respect to scalability of names it appears that many of the datasets names are of the format `LastnameYEAR` where Lastname refers to the lastname of the first author of the dataset. This system could be codified as a rule for published datasets. For unpublished datasets I don't think a simple system will be that easy to define. 

Additionally it appears that the output of `retriever ls` does not automatically detect all of the available datasets. For example the newly created script `EA_palmer2007.script` which generates the dataset named `Palmer2007` is not listed by `retriever ls`.

ethanwhite:
I also agree that this is something that needs improving, just haven't had time to work on it.

> Additionally it appears that the output of retriever ls does not automatically detect all of the available datasets. For example the newly created script EA_palmer2007.script which generates the dataset named Palmer2007 is not listed by retriever ls. 

This is probably because you're running the most recent release rather than the current version of `master`. The current release will only download scripts that existed as of that release. The truth is the entire relationships between scripts and releases needs to be thought about from both a technical and user perspective and I'm hoping that once I get a software engineer hired we'll be able to tackle that side of things.

shreyneil:
I think this issue was solved using  https://github.com/weecology/retriever/pull/488 . Please review and close this issue.

ethanwhite:
I agree that this has generally been addressed. Thanks for pointing this out @shreyneil!

We've fully addressed the first point in that they are explicitly linked at https://retriever.readthedocs.io/en/latest/datasets_list.html and through the verbose presentation that @shreyneil points to.

We've also improved the ability to digest and work with this metadata in Python and R.

[...]

PR #488, opened 2016-05-13
Verbose mode for dataset display using 'retriever ls'

goelakash:
#210 
Screenshot:

1. without `-l` flag
selection_009

2.  with `-l` flag for verbose output
selection_010

@ethanwhite @henrykironde

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -88,7 +88,6 @@ def main():
             return
 
         if args.command == 'ls' or args.dataset is None:
-            import lscolumns
 
             # If scripts have never been downloaded there is nothing to list
             if not script_list:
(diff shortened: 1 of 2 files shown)


henrykironde:
This is Great @goelakash, 
I want us to envision a case where we have a lot of data sets, how can we effectively display the information. One way could be sorting alphabetically based on some variables, ie. tags (list based on some grouping like plants ) or citation ....

We could present this PR changes as the default display of results and then if a person gives an optional flag
for example`` retriever ls -l `` could display the data based on some criteria.

ethanwhite:
Thanks @goelakash! I like the idea of keeping the current compressed output with `retriever ls` and this longer form with more detail with `retriever ls -l` (very analogous to the behavior of `ls` itself).

I also like the idea of being able to filter based on tags, e.g., `retriever ls fish` and `retriever ls -l fish` just showing the outputs for datasets containing the `fish` tag. (this part could definitely be a separate PR if you preferred).

henrykironde:
looks good, just let me know if you plan to add anything

goelakash:
@henrykironde  Added extension to search with keywords.

PR #489, opened 2016-05-16
Forcing value to type str for keys starting with '*'

goelakash:
Using str(e) unconditionally.
@henrykironde @ethanwhite

### DIFF
--- a/lib/compile.py
+++ b/lib/compile.py
@@ -89,7 +89,7 @@ def compile_script(script_file):
                     except:
                         e = str(value)
 
-                    tables[last_table][key] = str(e) if e.__class__ != str else "'" + e + "'"
+                    tables[last_table][key] = "'" + str(e) + "'"
             else:
                 # general script attributes
                 values[key] = '"' + value + '"'


henrykironde:
for the record, could you give some scenarios/reasons for the change.

goelakash:
While porting code to Python3, we need to import `builtins.str`, which is of different type than python2 `str`. This if else condition was using `eval(value)`, which was causing problems in doing this.

Either way, forcing `str` to `str` doesn't cause a problem I think.

henrykironde:
Okey, I think since this will be introduced when we shifting to python 3, I will keep it open and merge it when we are merging all changes related to the python 3 shift :+1: .  Hope the wait does not affect our progress forward.

henrykironde:
`` if e.__class__ != str else "'" + e + "'"  `` interesting code. :+1:

goelakash:
Delimiters are always strings, I just checked. 
Why this code was introduced, I don't know, since we could easily flag those scripts as invalid which didn't have string delimiters.

henrykironde:
merging because the added line removes redundancy in code

PR #490, opened 2016-05-17
Print exception message if script not loaded on installation

goelakash:
Prints description of error when `__init__.py` fails to compile script

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -94,8 +94,8 @@ def MODULE_LIST(force_compile=False):
                     # make sure it works and then add it
                     new_module.SCRIPT.download
                     modules.append(new_module)
-            except:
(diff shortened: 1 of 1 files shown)


henrykironde:
Thanks @goelakash. Since it fails on codecov, we could bring this in later.
I suggest combining these changes to the json package branch you working on.

henrykironde:
could you also include the output before and output after the changes.

goelakash:
@henrykironde This doesn't change the output currently as all the scripts compile successfully. I just thought about adding an exception description for debugging script compilation in the future.

henrykironde:
Its okey, I will go ahead and close this. We shall include it later as we work on the compile module or something related

PR #492, opened 2016-05-23
Tests: Add test for the exception condition of the function correct_invalid_value

Deborah-Digges:
Adding a test case for the following function:

```
def correct_invalid_value(value, args):
```

in `cleanup.py`

### DIFF
--- a/test/test_retriever.py
+++ b/test/test_retriever.py
@@ -151,3 +151,6 @@ def test_getmd5():
     """Test md5 sum calculation"""
     lines = ['a,b,c\n', '1,2,3\n', '4,5,6\n']
     assert getmd5(lines) == '0bec5bf6f93c547bc9c6774acaf85e1a'
+
+def test_invalid_value_number_exception():
+    assert correct_invalid_value(-999, {}) == -999


test/test_retriever.py
```diff
@@ -151,3 +151,6 @@ def test_getmd5():
     """Test md5 sum calculation"""
     lines = ['a,b,c\n', '1,2,3\n', '4,5,6\n']
     assert getmd5(lines) == '0bec5bf6f93c547bc9c6774acaf85e1a'
+
+def test_invalid_value_number_exception():
+    assert correct_invalid_value(-999, {}) == -999
```
henrykironde: looks good, could you comment some other values we could give as the second args right now its takes {}

Deborah-Digges:
Hi @henrykironde. Thanks for reviewing this. The second argument is supposed to take a dictionary with a key `nulls` that specifies the values that must be replaced by None. In the absence of this key, the function throws an exception and causes the value to be returned as is. This is the condition that this test executes

henrykironde:
great, could we include one case where there are keys present?

Deborah-Digges:
Those are already present [here](https://github.com/Deborah-Digges/retriever/blob/95a8dec99b4818bf4197961b3b75fdaee4fe0ec4/test/test_retriever.py#L54).

henrykironde:
okey i see what you did, I have just realised that 

```
def test_correct_invalid_value_string():
    assert correct_invalid_value('NA', {'nulls': ['NA', '-999']}) == None

def test_correct_invalid_value_number():
    assert correct_invalid_value(-999, {'nulls': ['NA', '-999']}) == None
```
could you move them close and  name it some thing like  
`def test_correct_invalid_value_exception`

henrykironde:
very good

Deborah-Digges:
sure 👍 Will make those changes

Deborah-Digges:
Sorry, I ended up making a new [PR](https://github.com/weecology/retriever/pull/493)

henrykironde:
I will go ahead and close this

PR #493, opened 2016-05-23
Update test_retriever.py

Deborah-Digges:


test/test_retriever.py
```diff
@@ -57,6 +57,9 @@ def test_correct_invalid_value_string():
 def test_correct_invalid_value_number():
     assert correct_invalid_value(-999, {'nulls': ['NA', '-999']}) == None
     
+def test_correct_invalid_value_exception():
```
henrykironde: :+1:

Deborah-Digges: Thank you 👍

PR #494, opened 2016-05-24
Just an update to the main branch

goelakash:
@ethanwhite In the future, should I directly commit to this branch, or fork another branch from this and commit to that?


henrykironde:
Hi @goelakash, I discussed the PR with @ethanwhite and in the future dno need to update unless there is a strong need for that. Git takes care of the update  incase  there are no conflicts otherwise we could manually merge using command line

goelakash:
Okay. There were some compile.py related updates that I needed so that's why I updated this branch.

Issue #487, opened 2016-05-12
Transition to Python 3 and change scripts format to JSON


henrykironde:
References: #167 and #326 

The current list of objectives includes:

- [x] Update .script to .json files
- [x] Update code for Python2/3 compatibility
- [x] add CLI options for creating a metadata .json file for a new dataset, and edit old ones.

__Start date - 23rd may__

Week 1
---------
Convert .script files to json

Week 2
---------
Add JSON compilation support.

Week 3 & 4
--------------
Start developing the CLI tool. 

Week 5 & 6
--------------
Work on adding Python 3 support

Midterm evaluation Deliverables
-----------------------------------------
 - datapackage.json scripts
 - CLI tool for datapackage.json metadata file creation
 - Python 3 support

Week 6 & 7
--------------
Writing tests for the test-suite and improve test-suite coverage

Week 8
---------
Triaging bugs and problems associated to consolidation of cross-tabulated data.

Week 9
---------
Writing documentation

Final evaluation Deliverables
------------------------------------
- Additional tests
- Documentation for datapackage.json CLI and manual editing
- bug-patches

Post- GSoC
---------------
Add new features and work on other bugs and issues.

ethanwhite:
@henrykironde & @goelakash - this is a great super issue for covering all of the things that @goelakash is working on this summer as part of GSoC. @goelakash - I'm guessing that @henrykironde's already talked with you a bit about the ideas behind these sorts of issues, but in the spirit of GSoC giving you exposure to who folks work on bigger projects I thought I'd take a couple of minutes to give my thinking behind this sort of thing and then make a suggestion or two for making this one even better.

When working on major changes like the two sets of things you're working on this summer it's important to keep track of all of the different bits and pieces to make sure that everything gets done and that you're managing your time well (e.g., not spending most of your time on a small piece of the project and not getting around to the more core elements). There are two general approaches to this that we use in our work: super issues and milestones.

Super issues, like this one, collect all of the different pieces of the bigger task together in a single issue. You can use the `- [ ] item 1` and `- [x] item 2` markdown notation to produce lists like:

- [ ] item 1
- [x] item 2

that let you check off items right in the issue. Each item is then often associated with a smaller sub-issue where ideas/discussion/tracking of that specific task can happen. These sub-issues are often associated with the items, e.g., 

- [ ] item 1 #22 
- [x] item 2 #26 

[...]

PR #497, opened 2016-05-29
Python3 compatibility changes

goelakash:
Working on #487 
@henrykironde @ethanwhite

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ retriever.egg-info
 apidocs
 .DS_Store
 .cache/*
+*.bak
--- a/engines/jsonengine.py
+++ b/engines/jsonengine.py
@@ -1,4 +1,7 @@
 """Engine for writing data to a JSON file"""
+from builtins import zip
+from builtins import object
+from builtins import range
 
 import os
 import json
@@ -7,7 +10,7 @@ from retriever.lib.models import Engine
 from retriever import DATA_DIR
 from collections import OrderedDict
 
-class DummyConnection:
+class DummyConnection(object):
     def cursor(self):
         pass
 
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -1,3 +1,9 @@
+from __future__ import print_function
+from builtins import object
+from future import standard_library
+standard_library.install_aliases()
+from builtins import range
+from builtins import input
 import sys
 import os
 import getpass
@@ -13,7 +19,7 @@ from retriever.lib.cleanup import no_cleanup
 from retriever.lib.warning import Warning
 
 
-class Engine():
+class Engine(object):
     """A generic database system. Specific database platforms will inherit
     from this class."""
     name = ""
@@ -108,11 +114,11 @@ class Engine():
                     insert_stmt = self.insert_statement(cleanvalues)
                 except:
                     if self.debug:
-                        print types
+                        print(types)
                     if self.debug:
-                        print linevalues
+                        print(linevalues)
                     if self.debug:
-                        print cleanvalues
+                        print(cleanvalues)
                     raise
 
                 try:
@@ -130,10 +136,9 @@ class Engine():
                 try:
                     self.execute(insert_stmt, commit=False)
                 except:
-                    print insert_stmt
+                    print(insert_stmt)
                     raise
 
-        print
         self.connection.commit()
 
     def auto_create_table(self, table, url=None, filename=None, pk=None):
@@ -153,7 +158,7 @@ class Engine():
                    (open, (file_path, "rb"))))
         lines = gen_from_source(source)
 
-        header = lines.next()
+        header = next(lines)
         lines.close()
 
         source = (skip_rows,
@@ -189,7 +194,7 @@ class Engine():
             lines_to_scan = []
             n = 0
             while n < lines:
-                lines_to_scan.append(source.next())
+                lines_to_scan.append(next(source))
                 n += 1
         else:
             lines_to_scan = source
@@ -224,7 +229,7 @@ class Engine():
                                     value = float(value)
                                     if "e" in str(value) or ("." in str(value) and
                                                              len(str(value).split(".")[1]) > 10):
-                                        column_types[i] = ["decimal","30,20"]
+                                        column_types[i] = ["decimal", "30,20"]
                                 except:
                                     column_types[i] = ['char', max_lengths[i]]
                             if column_types[i][0] == 'char':
@@ -266,7 +271,7 @@ class Engine():
         elif thistype[0:3] == "ct-":
             thistype = thistype[3:]
 
-        if thistype in self.datatypes.keys():
+        if thistype in list(self.datatypes.keys()):
             thistype = self.datatypes[thistype]
 
             if isinstance(thistype, tuple):
@@ -290,11 +295,11 @@ class Engine():
         engine.db"""
         db_name = self.database_name()
         if db_name:
-            print "Creating database " + db_name + "..."
+            print("Creating database " + db_name + "...")
             # Create the database
             create_stmt = self.create_db_statement()
             if self.debug:
-                print create_stmt
+                print(create_stmt)
             try:
                 self.execute(create_stmt)
             except Exception as e:
@@ -302,7 +307,7 @@ class Engine():
                     self.connection.rollback()
                 except:
                     pass
-                print "Couldn't create database (%s). Trying to continue anyway." % e
+                print("Couldn't create database (%s). Trying to continue anyway." % e)
 
     def create_db_statement(self):
         """Returns a SQL statement to create a database."""
@@ -319,7 +324,7 @@ class Engine():
     def create_table(self):
         """Creates a new database table based on settings supplied in Table
         object engine.table."""
-        print "Creating table " + self.table_name() + "..."
+        print("Creating table " + self.table_name() + "...")
 
         # Try to drop the table if it exists; this may cause an exception if it
         # doesn't exist, so ignore exceptions
@@ -330,7 +335,7 @@ class Engine():
 
         create_stmt = self.create_table_statement()
         if self.debug:
-            print create_stmt
+            print(create_stmt)
         try:
             self.execute(create_stmt)
         except Exception as e:
@@ -338,7 +343,7 @@ class Engine():
                 self.connection.rollback()
             except:
                 pass
-            print "Couldn't create table (%s). Trying to continue anyway." % e
+            print("Couldn't create table (%s). Trying to continue anyway." % e)
 
     def create_table_statement(self):
         """Returns a SQL statement to create a table."""
@@ -353,7 +358,7 @@ class Engine():
                     types.append(self.convert_data_type(column[1]))
 
         if self.debug:
-            print columns
+            print(columns)
 
         column_strings = []
         for c, t in zip(columns, types):
@@ -384,8 +389,8 @@ class Engine():
         if not self.find_file(filename):
             path = self.format_filename(filename)
             self.create_raw_data_dir()
-            print "Downloading " + filename + "..."
-            file = urllib.urlopen(url)
+            print("Downloading " + filename + "...")
+            file = urllib.request.urlopen(url)
             local_file = open(path, 'wb')
             if clean_line_endings and (filename.split('.')[-1].lower() not in ["exe", "zip", "xls"]):
                 local_file.write(file.read().replace("\r\n", "\n").replace("\r", "\n"))
@@ -469,13 +474,13 @@ class Engine():
             script.shortname,
             key
             )
-            for key in script.urls.keys() if key])
+            for key in list(script.urls.keys()) if key])
 
     def final_cleanup(self):
         """Close the database connection."""
 
         if self.warnings:
-            print '\n'.join(str(w) for w in self.warnings)
+            print('\n'.join(str(w) for w in self.warnings))
 
         self.disconnect()
 
@@ -552,16 +557,16 @@ class Engine():
         """Manually get user input for connection information when script is
         run from terminal."""
         for opt in self.required_opts:
-            if not (opt[0] in self.opts.keys()):
+            if not (opt[0] in list(self.opts.keys())):
                 if opt[0] == "password":
-                    print opt[1]
+                    print(opt[1])
                     self.opts[opt[0]] = getpass.getpass(" ")
                 else:
                     prompt = opt[1]
                     if opt[2]:
                         prompt += " or press Enter for the default, %s" % opt[2]
                     prompt += ': '
-                    self.opts[opt[0]] = raw_input(prompt)
+                    self.opts[opt[0]] = input(prompt)
             if self.opts[opt[0]] in ["", "default"]:
                 self.opts[opt[0]] = opt[2]
 
@@ -596,7 +601,7 @@ class Engine():
         else:
             # Save a copy of the file locally, then load from that file
             self.create_raw_data_dir()
-            print "Saving a copy of " + filename + "..."
+            print("Saving a copy of " + filename + "...")
             self.download_file(url, filename)
             self.insert_data_from_file(self.find_file(filename))
 
@@ -618,7 +623,7 @@ class Engine():
             n += 1
         insert_stmt %= tuple([str(value) for value in values])
         if self.debug:
-            print insert_stmt
+            print(insert_stmt)
         return insert_stmt
 
     def table_exists(self, dbname, tablename):
@@ -645,7 +650,7 @@ def skip_rows(rows, source):
     """Skip over the header lines by reading them before processing."""
     lines = gen_from_source(source)
     for i in range(rows):
-        lines.next()
+        next(lines)
     return lines
 
 
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -1,4 +1,5 @@
 """Checks the repository for updates."""
+from __future__ import print_function
 
 
 import os
@@ -47,7 +48,7 @@ def check_for_updates():
     init = InitThread()
     init.run()
 
-    print "\nThe retriever is up-to-date"
+    print("\nThe retriever is up-to-date")
 
 
 def update_progressbar(progress):
@@ -96,7 +97,7 @@ class InitThread(Thread):
             # read scripts from the repository and the checksums from the
             # version.txt
             scripts = []
-            print "Downloading scripts..."
+            print("Downloading scripts...")
             for line in version_file:
                 scripts.append(line.strip('\n').split(','))
 
@@ -136,11 +137,11 @@ class InitThread(Thread):
 
                 if need_to_download:
                     try:
-                        os.remove(os.path.normpath(os.path.join(HOME_DIR,"scripts", script_name)))
+                        os.remove(os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name)))
                         download_from_repository("scripts/" + script_name,
                                                  os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
                     except Exception as e:
-                        print e
+                        print(e)
                         pass
                 update_progressbar(float(index + 1)/float(total_script_count))
         except:
--- a/lib/table.py
+++ b/lib/table.py
@@ -1,10 +1,15 @@
+from future import standard_library
+standard_library.install_aliases()
+from builtins import next
+from builtins import object
 import csv
-import StringIO
+import io
+from functools import reduce
 
 from retriever.lib.cleanup import *
 
 
-class Table:
+class Table(object):
     """Information about a database table."""
 
     def __init__(self, name, **kwargs):
@@ -22,7 +27,7 @@ class Table:
         self.escape_single_quotes = True
         self.escape_double_quotes = True
         self.cleaned_columns = False
-        for key, item in kwargs.items():
+        for key, item in list(kwargs.items()):
             setattr(self, key, item[0] if isinstance(item, tuple) else item)
 
     def auto_get_columns(self, header):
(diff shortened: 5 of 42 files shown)


henrykironde:
@goelakash, What is the status of the PR?

goelakash:
It adds py3 compatibility to all files in `app` folder and most of the files in `lib` folder, except a few. 

This is will need much more work before complete compatibility is achieved. But as @ethanwhite mentioned in #487 the python3 changes could be periodically merged as it maintains both python2 and 3 compatibility.

henrykironde:
That's fine @goelakash, Thanks for the good work. I actually wanted to know if you wanted us to consider it for merging. we shall go ahead and have a look at it.
 we shall go ahead and have a look at it.

lib/engine.py
```diff
@@ -130,10 +131,10 @@ def source_gen():
                 try:
                     self.execute(insert_stmt, commit=False)
                 except:
-                    print insert_stmt
+                    print(insert_stmt)
                     raise
 
-        print
+        print()
```
henrykironde: we can remove this print

lib/engine.py
```diff
@@ -290,19 +291,19 @@ def create_db(self):
...
+                print(create_stmt)
             try:
                 self.execute(create_stmt)
             except Exception as e:
                 try:
                     self.connection.rollback()
                 except:
                     pass
-                print "Couldn't create database (%s). Trying to continue anyway." % e
+                print("Couldn't create database (%s). Trying to continue anyway." % e)
```
henrykironde: use { }".format instead if %

scripts/bbs50stop.py
```diff
@@ -177,14 +183,14 @@ def regioncodes_cleanup(value, engine):
             for part in range(1,11):
```
henrykironde: i think we can change this too  `for part in range(1,11):`

henrykironde:
@goelakash,still under review, however, I would suggest you grep all the prints that you may have missed and wrap them too.  
- also change all of them to .{ }".format where need be.
- squash the commits, I think we do not need that  first commit related to `add_unicode.py`
@ethanwhite will also take a look and if all review are cleared accordingly, this will get merged, Overall great :+1:  :) 
last. since we plan to remove the Gui, I think do not put more effort on the app directory. And also the `test/test_regression_cli.py` will change completely

scripts/bbs50stop.py
goelakash: To what should I change it to?

scripts/npn.py
```diff
@@ -51,7 +54,7 @@ def download(self, engine=None, debug=False):
             engine.create_table()
 
             base_url = 'http://www.usanpn.org/getObs/observations/'
-            years = range(2009, 2013)
+            years = list(range(2009, 2013))
```
henrykironde: @goelakash, so we have wrapped range(2009, 2013) -> list(range(2009, 2013))
and if this change is relevant to py3 we can go ahead and wrap other similar occurrences of range.
What do you think?

goelakash: @henrykironde Yes, consider these python 2 and 3 snippets:
```
Python 2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> x = range(0,10)
>>> x
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

```
Python 3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> type(range(0,1))
<class 'range'>
>>> x = range(0,1)
>>> x
range(0, 1)
>>> x = range(0,10)
>>> x
range(0, 10)
>>> list(x)
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

goelakash:
@henrykironde All these changes were done using `futurize` python package. I had to do incremental changes to ensure that all the tests were passing.

lib/engine.py
goelakash: There's a lot of these old-style format in the code, and changing that will be a lot of manual changes . Since both are supported in python 3, I think we could create a issue and delay it until important changes are done?

goelakash:
@henrykironde Can this me merged?

goelakash:
@henrykironde The last two commits pass on my machine but fail on travis. Any clues? (I did a `python setup.py install` before running `py.test`)

henrykironde:
@goelakash, yah the changes are really okey, so after you have finished  with the possible changes we are going to merge it. And like you said, we do not have to make all changes at once but we can actually try to cover most. If we can't have all the prints changed and the range () ->list(range). Its also okey. I think we can merge it in tomorrow. The great news is that we have finished removing the dependency on the server and I guess tomorrow we can  have that merged too.

lib/engine.py
ethanwhite: I agree with @goelakash that it's not necessary to make those changes in existing lines that we're touching to make Python 3 compatibility changes. It would be fine to make those changes at the same time, but I also don't think it hurts anything to wait. Any new lines that get added should use `"{}".format` should use.

ethanwhite:
I can't currently build this on my machine due to the `builtins` import:

```
ethan@stice:~/RetrieverCode ((e2c4ab0...))$ pip install . --upgrade
Processing /home/ethan/Dropbox/Informatics/Retriever/src
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-CZcPpw-build/setup.py", line 25, in <module>
        from __init__ import VERSION
      File "/tmp/pip-CZcPpw-build/__init__.py", line 8, in <module>
        from builtins import str
    ImportError: No module named builtins
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-CZcPpw-build/
```

I'm not sure is this is related to the Travis failure, but if someone could let me know what I need to do to get it working I'd appreciate it (I already tried install `builtins` via `pip`).

ethanwhite:
Ah, never mind. The module is `future`. I've got it installed and everything is building.

app/download_manager.py
```diff
@@ -97,7 +99,7 @@ def progress(s):
                     s = s.split(' / ')
                     total = float(s[1])
                     current = float(s[0].split(': ')[1])
-                    progress = int((current / total) * 1000)
+                    progress = int((old_div(current, total)) * 1000)
```
ethanwhite: Do we need integer division here? At first glance I suspect that standard division is fine and that this is just an automatic change. If we don't need it we should keep the original form for simplicity and drop the extra import.

lib/engine.py
```diff
@@ -1,10 +1,14 @@
...
+standard_library.install_aliases()
+from builtins import range
 import sys
 import os
 import getpass
 import zipfile
 import gzip
 import tarfile
-import urllib
+import urllib.request, urllib.parse, urllib.error
```
ethanwhite: I'm a little confused by this line (probably because I don't know much about Python 2/3 compatibility :smile:). We don't seem to use the last two imports and when we use `urllib.request` we call it as `urllib.request.urlopen`, which seems to work fine for me using just `import urllib`.

@goelakash - can you explain what's going on here to me? Thanks.

lib/repository.py
```diff
@@ -1,9 +1,15 @@
...
+from future import standard_library
+standard_library.install_aliases()
+from past.utils import old_div
+from builtins import range
 
 
 import os
 import sys
-import urllib
+import urllib.request, urllib.parse, urllib.error
```
ethanwhite: Same question as above about this import.

lib/repository.py
```diff
@@ -185,7 +191,7 @@ def run(self):
             if not os.path.isdir(SCRIPT_WRITE_PATH):
                 os.makedirs(SCRIPT_WRITE_PATH)
 
-            update_progressbar(0.0/float(total_script_count))
+            update_progressbar(old_div(0.0,float(total_script_count)))
```
ethanwhite: Same point as above about whether or not we need `old_div`.

lib/repository.py
```diff
@@ -219,9 +225,9 @@ def run(self):
                         download_from_repository("scripts/" + script_name,
                                                  os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
                     except Exception as e:
-                        print e
+                        print(e)
                         pass
-                update_progressbar(float(index + 1)/float(total_script_count))
+                update_progressbar(old_div(float(index + 1),float(total_script_count)))
```
ethanwhite: Again, do we need `old_div`.

lib/table.py
```diff
@@ -1,10 +1,15 @@
+from future import standard_library
+standard_library.install_aliases()
+from builtins import next
+from builtins import object
 import csv
-import StringIO
+import io
 
 from retriever.lib.cleanup import *
+from functools import reduce
```
ethanwhite: Imports should be grouped and ordered following https://www.python.org/dev/peps/pep-0008/#imports, so `functools` should be imported before the local retriever imports.

scripts/bbs.py
```diff
@@ -2,9 +2,12 @@
 """Retriever script for Breeding Bird Survey
 
 """
+from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
 
 import os
-import urllib
+import urllib.request, urllib.parse, urllib.error
```
ethanwhite: Same question about import.

scripts/bbs50stop.py
```diff
@@ -2,9 +2,13 @@
...
 
 """
+from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
+from builtins import range
 
 import os
-import urllib
+import urllib.request, urllib.parse, urllib.error
```
ethanwhite: Same question about import.

scripts/fia.py
```diff
@@ -2,9 +2,12 @@
 """Retriever script for Forest Inventory and Analysis
 
 """
+from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
 
 import os
-import urllib
+import urllib.request, urllib.parse, urllib.error
```
ethanwhite: Same question about the import.

scripts/npn.py
```diff
@@ -1,9 +1,12 @@
 #retriever
 """Retriever script for National Phenology Network data
 """
+from future import standard_library
+standard_library.install_aliases()
+from builtins import range
 
 import os
-import urllib
+import urllib.request, urllib.parse, urllib.error
```
ethanwhite: Same question about the import.

ethanwhite:
Thanks for the great work on this @goelakash. I've made a a few in-line comments that are a combination of suggestions and things I don't understand.

This is also passing on my local machine, but we do need to figure out why it is failing on Travis and get that fixed before we merge.

lib/repository.py
goelakash: Python 2 has `urllib.urlopen()` and Python  3 has `urllib.request.urlopen()`.
The lines
```
```
gets and alternative version of `urllib` from the future package. This allows the syntax to be `urllib.request.urlopen()` everywhere.

Although, I agree that we don't need `urllib.parse` and `urllib.error`. I'll remove the unused imports everywhere.

scripts/prism_4km_monthly_modern.py
```diff
@@ -1,9 +1,12 @@
 #retriever
 
 """Retriever script for direct download of PRISM climate data"""
+from future import standard_library
+standard_library.install_aliases()
+from builtins import range
 
 from retriever.lib.templates import Script
-import urlparse
+import urllib.parse
```
goelakash: Python 2 has `urlparse` and Python 3 has `urllib.parse`. Similar to that I explained above.

goelakash:
@ethanwhite I edited the imports and removed `old_div` where it wasn't needed.

lib/repository.py
```diff
@@ -191,7 +191,7 @@ def run(self):
             if not os.path.isdir(SCRIPT_WRITE_PATH):
                 os.makedirs(SCRIPT_WRITE_PATH)
 
-            update_progressbar(old_div(0.0,float(total_script_count)))
+            update_progressbar(0.0/float(total_script_count))
```
ethanwhite: Add whitespace around `/`

lib/repository.py
```diff
@@ -227,7 +227,7 @@ def run(self):
                     except Exception as e:
                         print(e)
                         pass
-                update_progressbar(old_div(float(index + 1),float(total_script_count)))
+                update_progressbar(float(index + 1),float(total_script_count))
```
ethanwhite: Should this be: `float(index + 1) / float(total_script_count)`?

ethanwhite:
Thanks @goelakash. I noted two small issues to clean up and then I'm happy with all of the things I noted.

Any ideas on what is going on with the Travis failures?

henrykironde:
could you run `retriever reset all ` and try the tests again.

app/app.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, lists):
             engine = wizard.CONNECTION.engine
             options = wizard.CONNECTION.option
             opts = dict()
-            for key in options.keys():
+            for key in list(options.keys()):
```
QuLogic: The `list` is not necessary here (and below).

QuLogic: In fact, neither is `keys()`.

app/connect_wizard.py
```diff
@@ -155,7 +156,7 @@ def Draw(self, evt):
                 self.option = dict()
                 saved_opts = get_saved_connection(self.engine.name)
                 for opt in self.engine.required_opts:
-                    if opt[0] in saved_opts.keys():
+                    if opt[0] in list(saved_opts.keys()):
```
QuLogic: Here as well, neither `list` nor `keys` are necessary.

ethanwhite:
Thanks for the review @QuLogic!

ethanwhite:
I merged #499 today without thinking about this PR so there are going to be conflicts with the Python 3 compatibility changes in the `app/` files. Sorry about that.

Handling a fairly large conflict like this will be a little tricky. Off the top of my head what I would probably do is: 

1. Create a 2nd copy of the branch your working in to test on (I do this for all complicated git tasks so that I can get it wrong and easily go back and start again). Work in this copy.
2. Reset all of the files in the `app/` directory to their unmodified state when you started this PR (just to save some conflict resolution since they're getting deleted anyway). You can `checkout` individuals files at specific points in this history. See https://stackoverflow.com/questions/7147270/hard-reset-of-a-single-file.
3. Squash all of your commits into a single commit (which you'll want to do eventually anyway)
4. Rebase your branch against `master`
5. Fix any remaining conflicts

I'm happy to provide more specific walk throughs on the git process if that would be helpful, and I can also play around with the simplest way to get this work if it turns out to be difficult, so feel free to ask for help.

QuLogic:
I'd say it's better to rebase to `master` before squashing (and maybe not to a single commit either), because then every commit is smaller and easier to manage. Of course, that's assuming you didn't make a lot of back-and-forth commits over the same lines, but git can sometimes figure it out anyway.

app/app.py
QuLogic: You could also replace the whole loop with a dictionary comprehension if you don't need to support Python 2.6:
```python
engine.opts = {key: value.GetValue() for key, value in options.items()}
```

engines/jsonengine.py
```diff
@@ -108,7 +109,7 @@ def insert_statement(self, values):
                 offset += 1
 
         keys = [columnname[0] for columnname in self.table.columns]
-        tuples = (zip(keys, values))
+        tuples = (list(zip(keys, values)))
```
QuLogic: `OrderdedDict` accepts a generator, so `list` not required.

engines/postgres.py
```diff
@@ -1,3 +1,4 @@
+from past.builtins import basestring
```
QuLogic: The usage of `basestring` is not right in this file. It attempts to `decode()` the object, but `basestring` covers both `str` and `bytes`. `str` does not have a `decode` method.

lib/engine.py
```diff
@@ -552,16 +555,16 @@ def get_input(self):
...
-                    print opt[1]
+                    print(opt[1])
                     self.opts[opt[0]] = getpass.getpass(" ")
                 else:
                     prompt = opt[1]
                     if opt[2]:
                         prompt += " or press Enter for the default, %s" % opt[2]
                     prompt += ': '
-                    self.opts[opt[0]] = raw_input(prompt)
+                    self.opts[opt[0]] = input(prompt)
```
QuLogic: You haven't `from builtins import input`, so this is "unsafe" on Python 2.

lib/repository.py
```diff
@@ -219,9 +224,9 @@ def run(self):
                         download_from_repository("scripts/" + script_name,
                                                  os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
                     except Exception as e:
-                        print e
+                        print(e)
                         pass
-                update_progressbar(float(index + 1)/float(total_script_count))
+                update_progressbar(float(index + 1),float(total_script_count))
```
QuLogic: This shouldn't be changed (and is messing up Travis testing.)

lib/tools.py
```diff
@@ -146,10 +147,10 @@ def reset_retriever(scope):
     }
 
     warn_msg = warning_messages[scope].format(HOME_DIR)
-    confirm = raw_input(warn_msg)
+    confirm = input(warn_msg)
```
QuLogic: Same issue with `input` here.

lscolumns.py
```diff
@@ -1,10 +1,15 @@
...
+from past.utils import old_div
+from builtins import range
 import sys
 from term_size import get_terminal_size
 
 
 def get_columns(values, cols):
     columns = []
-    col_size = len(values) / cols
+    col_size = old_div(len(values), cols)
```
QuLogic: No need for `old_div`; use `len(values) // cols`.

scripts/prism_4km_monthly_modern.py
```diff
@@ -34,14 +37,14 @@ def download(self, engine=None, debug=False):
         Script.download(self, engine, debug)
 
         clim_vars = ['ppt', 'tmax', 'tmean', 'tmin']
-        years = range(1981, 2015)
+        years = list(range(1981, 2015))
         months = ['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11', '12']
```
QuLogic: I don't see why these need to be here and not just on the `for` loop. For `months`, use `range(1, 13)` and format using `02d` to get the zero-padding.

app/connect_wizard.py
QuLogic: This applies to many other cases of `for something in list(dict.keys()):` and `if something in list(dict.keys()):`, so I won't comment on them all.

lib/engine.py
goelakash: fixed

lib/tools.py
goelakash: fixed this too.

goelakash:
@ethanwhite @QuLogic I have fixed all the current issues.

henrykironde:
@QuLogic, thanks for the contribution. @goelakash, this is ready. I just can't merge it in because postgres is failing on my server. Totally something not related to this. I guess @ethanwhite, could help us. Otherswise when I fix my station, I will get it in. :+1:

goelakash:
@henrykironde Please review.

henrykironde:
I am actually looking at the changes now.

lib/templates.py
```diff
@@ -51,7 +51,9 @@ def reference_url(self):
                 return None
 
     def checkengine(self, engine=None):
-        if not engine:
+        if isinstance(engine,type(engine)):
```
goelakash: Was getting `__nonzero__` related errors from builtins.object type.

QuLogic: If it's an instance of itself? How could that ever be `False`? I'm pretty sure what this test is really asking is `if engine is None:`.

henrykironde: yes that is basically checking that

henrykironde:
Overall this looks nice to me. If you are to revisit `def checkengine(self, engine=None)` that is fine too both ways we obtain the result. 
@QuLogic, thank you for the suggestions.
@ethanwhite, if you have any suggestions before we merge.

goelakash:
@henrykironde I changed the check condition to `engine == None`. I think thats better now. :+1:

.gitignore
```diff
@@ -8,3 +8,4 @@ retriever.egg-info
 apidocs
 .DS_Store
 .cache/*
+*.bak
```
goelakash: This should stay until we are sure that the build is compatible with python 3.

QuLogic:
> engine == None

The idiomatic way is `engine is None`, which is why I wrote that.

ethanwhite:
LGTM. Thanks for rerolling this @goelakash.

@henrykironde and @QuLogic are right that `engine is None` is the idiomatic what to write this conditional, so @goelakash if you'll go ahead and make that change then we'll get this in.

henrykironde:
@goelakash are changes final on this?
If yes i would merge it after we clear the building status

goelakash:
@henrykironde Yes the changes are final.
The branch builds without errors on my repository (https://travis-ci.org/goelakash/retriever/builds/135682707)

henrykironde:
Perfect, Lets first deal with these error and we could go ahead and merge it or after the revert

goelakash:
@henrykironde Since the size of ct_column has been changed, so is the database dump from sqlite.
So, before @ethanwhite 's commit, the 3rd line in the sqlite dump read:
```
CREATE TABLE DelMoral2010_species_plot_year (plot_id_year TEXT(20), plot_name TEXT(4), plot_number INTEGER, year INTEGER, species TEXT(20), count REAL );
```
But the species type is `TEXT(50)`. So I think the `known_md5` value must be changed.

goelakash:
@henrykironde @ethanwhite All the tests pass now :)

PR #498, opened 2016-06-01
improve test platform

henrykironde:
1. add windows platform tests Appveyor using appveyor.yml and remove WX requirement in setup.py
2. rename engines/csv.py → engines/csvengine.py

3. improve regression

- remove server dependency by dumping csv files using the excel dialect(function  to_csv)
- sort files when getting md5sums for files in directory
- add xml and json tests


henrykironde:
Note:  We still have to add `md5sums` for `Delmoral`. I suggest we add them later when we have a new copy of `Delmoral` merged. 

we could discus the refactoring of the  `def to_csv(self)` method of all engines and `format_insert_value` in the `[csv|xml|json]engines` based on the level of refinement that we need (DRY).

The tests on windows currently do not test PostgreSQL. I have not figured out how to set their image   so that it reads the passwords from the `pgpass.conf` file

engines/csvengine.py
```diff
@@ -97,6 +98,12 @@ def table_exists(self, dbname, tablename):
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
 
+    def to_csv(self):
+        """Export table from json engine to CSV file"""
```
ethanwhite: Should this be something like `"""Export sorted version of CSV file"""`

engines/jsonengine.py
```diff
@@ -117,6 +116,14 @@ def table_exists(self, dbname, tablename):
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
 
+    def to_csv(self):
+        """Export table from json engine to CSV file"""
+        keys = [columnname[0] for columnname in self.table.columns]
+        filename =self.table_name()
```
ethanwhite: missing space after `=`

engines/jsonengine.py
```diff
@@ -117,6 +116,14 @@ def table_exists(self, dbname, tablename):
...
+        csv_outfile =json2csv(str(filename),header_values=keys)
```
ethanwhite: `filename` isn't already a string? Just wondering why we need the `str(filename)` call.

Missing spaces after `=` and `,`

engines/mysql.py
```diff
@@ -72,7 +74,7 @@ def insert_data_from_file(self, filename):
                 self.cursor.execute(statement)
 
                 self.cursor.execute(mysql_set_autocommit_on)
-            except Exception as e:
+            except Exception :
```
ethanwhite: Extra space before `:`

engines/mysql.py
```diff
@@ -91,6 +93,25 @@ def table_exists(self, dbname, tablename):
...
+        sql_query = ("SELECT * FROM " + self.table_name() + ";")
+        self.cursor.execute(sql_query)
+        row = self.cursor.fetchone()
+        colnames = [tuple_i[0] for tuple_i in self.cursor.description]
+        csv_out = open(csvfile_output, "wb")
+        csv_writer = csv.writer(csv_out, dialect='excel')
+        csv_writer.writerow(colnames)
+
+        while row is not None:
+            csv_writer.writerow([values for values in row])
```
ethanwhite: What is `row`'s type? I'm guessing it's a tuple in which case `[values for values in row]` is just converting it to a list? It seems like `.writerow(row)` ought to work, or if not `.writerow(list(row))`, but I could definitely be missing something.

engines/postgres.py
```diff
@@ -122,6 +124,28 @@ def format_insert_value(self, value, datatype):
                 pass
         return Engine.format_insert_value(self, value, datatype)
 
+    def to_csv(self):
+        tablename = str(Engine.table_name(self))
+        tablefile = str(self.table.name)
+        csvfile_output = os.path.normpath(str(tablefile + '.csv'))
```
ethanwhite: Are the `str` calls necessary on these three lines? The last line seems to be a string + a string, which should already be a string. It seems likely the values inside the `str`'s on 128 and 129 are strings as well.

engines/postgres.py
```diff
@@ -122,6 +124,28 @@ def format_insert_value(self, value, datatype):
...
+        self.cursor.execute(sql_query)
+
+        row = self.cursor.fetchone()
+        colnames = [tuple_i[0] for tuple_i in self.cursor.description]
+        csv_out = open(csvfile_output, "wb")
+        csv_writer = csv.writer(csv_out, dialect='excel')
+        csv_writer.writerow(colnames)
+
+        while row is not None:
+            csv_writer.writerow([values for values in row])
```
ethanwhite: Same question about need for list comprehension as in MySQL.

ethanwhite:
Looking really good @henrykironde! I'm making a bunch of in-line suggestions for cleaning things up. One bigger picture thing I noticed is that all of the SQL `to_csv` methods are nearly identical. From what I can see there are only two differences:

1. How `tablename` and `csvfile_output` are determined
2. The `self.connection.text_factory = str` line in sqlite

I'd suggest moving most of the logic in these 3 functions to a single `sql_to_csv` function, probably in `lib.tools`. That function can take three arguments: `tablename`, `csvfile_output` and `sqlite_engine=False`. If `sqlite_engine` is set to `True`, then it triggers a conditional in `sql_to_csv` that runs the `self.connection.text_factory = str` line. The `to_csv` functions can then determine `tablename` and `csvfile_output` and pass those as arguments. The `to_csv` for SQLite can also pass `sqlite_engine=True`.

Does that make sense as a way to reduce the code duplication?

ethanwhite:
Also, I've noted a few of these in-line, but I noticed quite a few `str()` calls wrapping things that I'm pretty sure are already strings. Is there something happening with these calls that I'm not understanding?

ethanwhite:
I don't fully understand the addition of the `populate` argument yet, but my impression is that the default behavior (i.e., the one that happens most of the time) is to populate the table. Most uses keep the current default, which is `False` and then do an `if not` check, which returns `True`. Given this I think it would be more readable to have the default set to `populate=True` and the checks be of the form of:

```
if populate:
    self.create_table()
```

This also aligns with how standard reading of `populate` as indicating adding something.

lib/templates.py
```diff
@@ -33,11 +34,13 @@ def __str__(self):
             desc += "\n" + self.reference_url()
         return desc
 
-    def download(self, engine=None, debug=False):
+    def download(self, engine=None, debug=False, populate=False):
+        """when initializing the selected engine create_db=false will not overwrite the existing db"""
```
ethanwhite: Looks like the name of the argument and that described in the docstring don't match.

lib/templates.py
```diff
@@ -83,18 +86,24 @@ class BasicTextTemplate(Script):
...
         for key, value in self.urls.items():
-            self.engine.auto_create_table(self.tables[key], url=value)
-            self.engine.insert_data_from_url(value)
-            self.tables[key].record_id = 0
+            if not populate:
+                self.engine.auto_create_table(self.tables[key], url=value)
+                self.engine.insert_data_from_url(value)
+                self.tables[key].record_id = 0
+            else:
+                self.engine.auto_create_table(self.tables[key], url=value, populate=True)
```
ethanwhite: Since `populate` is passed in to this function, this line and 99 can be run as a single line before the conditional as:

```
self.engine.auto_create_table(self.tables[key], url=value, populate=populate)
```

lib/tools.py
```diff
@@ -111,13 +103,13 @@ def choose_engine(opts, choice=True):
     else:
         if not choice:
             return None
-        print "Choose a database engine:"
+        print ("Choose a database engine:")
```
ethanwhite: Extra space after `print`

lib/tools.py
```diff
@@ -111,13 +103,13 @@ def choose_engine(opts, choice=True):
...
         for engine in engine_list:
             if engine.abbreviation:
                 abbreviation = "(" + engine.abbreviation + ") "
             else:
                 abbreviation = ""
-            print "    " + abbreviation + engine.name
+            print( "    " + abbreviation + engine.name)
```
ethanwhite: Extra space between `(` and `"`

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+    Alex,US,25
+    Alex,PT,25
+    """
+    try:
+        fp = open(input_file, 'r')
+    except IOError as e:
+        print "I/O error({0}): {1}".format(e.errno, e.strerror)
+
+    if not is_valid_json(input_file):
+        print "invalid json file"
```
ethanwhite: Let's use Python 3 print functions for all new print lines we're adding at this point.

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+                else:
+                    for p_strings in previous_list:
+                        new_list.append("".join(str(p_strings) + str(item[column_name]) + ","))
+                    previous_list = new_list
+
+        for lines in previous_list:
+            outfile.write("\n" + str(lines[0:-1]))
+    outfile.close()
+    fp.close()
+    os.system(" rm -r {}".format(input_file))
```
ethanwhite: Extra space after first `"`

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+    <row>
+       <User>Ben</User>
+       <Country>US</Country>S
+       <Age>24</Age>
+    </row>
+    </root>
+
+    User,Country,Age
+    Alex,US,25
+    Alex,Uk,25
```
ethanwhite: The `Uk` here doesn't match the `PT` above.

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+    Ben,US,24
+    """
+    fp = open(input_file, 'r')
+    tree = ET.parse(fp)
```
ethanwhite: `fp` isn't very descriptive both here and in the json function. `xml_file`, `xml_input` or anything more descriptive would be better.

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+                        new_list.append("".join(str(p_strings) + value_x + ","))
+                        previous_lists = new_list
+        else:
+            print ("no header provided")
+            exit()
+        for lines in previous_lists:
+            outfile.write("\n" + str(lines[0:-1]))
+    outfile.close()
+    fp.close()
+    os.system(" rm -r {}".format(input_file))
```
ethanwhite: Extra space after first `"`

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+    return outputfile
+
+
+def getmd5(path=None, mode='rb', lines=None):
+    """Get MD5 of a file, files in a directory or MD5 value of a set of lines
+
+    default mode 'rb'
```
ethanwhite: This is already clear from the function definition so I don't think including it in the docstring adds anything. Listing the possible modes would though. e.g. "Possible values for mode include the default, 'rb', and 'rU', which ..."

ethanwhite:
For the new `get_md5`:

1. `sum` is a Python built-in so let's use `checksum`
2. `sum` is set to `md5()` three times and never set to anything else. We should set it once at the beginning of the function.
3. Lines 332-335 can be reduced to: `lines = open(file_path, mode)`
4. In terms of general design it seems to me like we have two main things: the location of some data and what type of data it is, so I think of this function definition as
  
  `def getmd5(data, data_type='lines', mode='rb')`

  with possible values for `data_type` being `lines`, `file`, `dir`. The processing then happens as:

  ```
 if data_type == 'lines':
      ...
      return sum.hexdigest()
  elif data_type == 'file':
      files = [os.path.normpath(data)]
  elif data_type == 'dir':
      for root, directories...
  ...
  for file_path in files:
      lines = open(file_path, mode)
      ...
  ```

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+
+    # write sorted row content to csv filename with header "infields"
+    tmp = open(sorted_txt, "rU")
+    in_txt = csv.reader(tmp, delimiter=',')
+    out_csv = csv.writer(open(filename, 'wb'))
+    out_csv.writerow(infields)
+    out_csv.writerows(in_txt)
+    tmp.close()
+    os.remove(os.path.normpath("tempfile"))
+    return (filename)
```
ethanwhite: No parentheses need here.

ethanwhite:
We shouldn't touch any of the `wx` related stuff in `setup.py` here since it is coming out in #499 where it makes sense to remove it. I know you took this out due to AppVeyor issues, so let's do all of the other cleanup, make sure everything passes, and then get rid of these changes before merging. We'll get #499 in at about the same time and then everything should be good to go. Will that work or am I missing something important?

test/test_retriever.py
```diff
@@ -153,4 +155,4 @@ def test_format_insert_value_string_complex():
 def test_getmd5():
     """Test md5 sum calculation"""
     lines = ['a,b,c\n', '1,2,3\n', '4,5,6\n']
-    assert getmd5(lines) == '0bec5bf6f93c547bc9c6774acaf85e1a'
+    assert getmd5(lines=lines) == '0bec5bf6f93c547bc9c6774acaf85e1a'
```
ethanwhite: If you end up making my suggested changes to `getmd5` this call will need to be modified to:

`getmd5(data=lines, data_type='lines')

test/test_regression_cli.py
```diff
@@ -1,153 +1,136 @@
...
+    os.chdir(dir_path)
+    # os.chdir("./test/")
     os.system("retriever update")
 
 
 def teardown_module():
-    """Cleanup temporary output files after testing and return to root directory"""
-    os.system("rm output_*")
+    """ teardown any state that was previously setup with a setup_method
+    call.
```
ethanwhite: The files being removed aren't being created `setup_module`, so I think the original docstring is probably clearer.

test/test_regression_cli.py
```diff
@@ -1,153 +1,136 @@
...
+filedb_md5 = [
+    # ('DelMoral2010', '0'),
+    ('AvianBodySize', '47bdcf49c28fad5a444709f33cbc5c20'),
+    ('MoM2003', 'a605380ab4610d54f431e5413ef35c5e')
+]
+
+
+def setup_module(module):
+    """ setup up any state specific to the execution
+        of the given module.
```
ethanwhite: Let's be specific about what the function is doing here. e.g. """Set working directory for running tests and update the retriever scripts before starting".

test/test_regression_cli.py
```diff
@@ -1,153 +1,136 @@
...
+    # at this point the current directory is ./test
+    if os.path.exists("output_dumps"):
+        os.system("rm -r output_dumps")
+    os.makedirs("output_dumps")
+    os.chdir("output_dumps")
+    if arg:
+        engine = choose_engine(arg.__dict__)
+        script_list = SCRIPT_LIST()
+        scripts = name_matches(script_list, arg.dataset)
+        for dataset_i in scripts:
```
ethanwhite: Any reason to not just use `dataset` for the variable here?

test/test_regression_cli.py
```diff
@@ -1,153 +1,136 @@
...
+    os.chdir("..")
+    current_md5 = getmd5("output_dumps", mode)
+    return current_md5
+
+
+@pytest.mark.parametrize("dataset,expected", download_md5)
+def test_download_regression(dataset, expected):
+    """Check for regression for a particular dataset downloaded only"""
+    os.system("retriever download {0} -p raw_data/{0}".format(dataset))
+    current_md5 = getmd5("raw_data/%s" % (dataset), mode="rU")
```
ethanwhite: Use `{}` and `.format`.

test/test_regression_cli.py
```diff
@@ -1,153 +1,136 @@
...
+            if arg.engine in ["csv", 'json', 'xml']:
+                # the engines in the list need to download the data
+                dataset_i.download(engine, arg.debug, False)
+                dataset_i.engine.final_cleanup()
+
+            dataset_i.download(engine, arg.debug, True)
+            dataset_i.engine.final_cleanup()
+    os.chdir("..")
+    current_md5 = getmd5("output_dumps", mode)
+    return current_md5
```
ethanwhite: I'm confused as to why we're handling csv, json, and xml differently than the other engines. I would have installed them using the CLI in the tests below just like you did for the database engines. Can you explain why we're installing them from here instead?

ethanwhite:
Overall this looks great @henrykironde! There's a ton of great work in hear and I really like the use of the pytest decorators, which I'd never seen before.

Once you've explained the reasoning for `populate` and `tocsv` to me a little more and changed or futher discussed the other suggestions, then I want to go back and:

- [ ] review the `populate` related changes
- [ ] review the `tocsv` function in `test_regression_cli.py`
- [ ] test all of this locally

engines/postgres.py
henrykironde: very good question :+1: 
So basically most of these `str() ` problems in conjunction with list compression where initially introduced as I was debugging. At some point I had to pre-process each value.

lib/tools.py
```diff
@@ -160,3 +152,235 @@ def reset_retriever(scope):
...
+    return True
+
+
+def json2csv(input_file, output_file=None, header_values=None):
+    """Convert Json file to CSV
+
+    [
+    {"User": "Alex", "Country": "US", "Age": "25"}
+    ]
+    Alex,US,25
```
henrykironde: Thanks, that would look more clear
User,Country,Age
Alex,US,25

lib/tools.py
henrykironde: I was actually going to explain why we need different modes after running through several pitfalls, but we can drop that.

henrykironde:
I got some errors as I was working on this, but my quick guess would be the line in lines = open(file_path, mode) am working on it

ethanwhite:
Now that #499 is in we've got some merge conflicts here do the the overlapping changes to the `wx` dependency. Since we're reworking this anyway we can probably rebase this off of master and clean up the associated conflicts. Sorry about the extra work.

henrykironde:
closed because moved to #511

Issue #441, opened 2016-03-08
Removing the GUI


ethanwhite:
As we start preparing to move to 2.0 it's time to consider removing the GUI.

In 2010 when we started this @bendmorris and I both felt it was pretty important, because many scientists were still doing a lot of data related work by hand in graphical environments. In 2016, thanks to cultural changes driven by increased amounts of data and interest in reproducibility, there has been a major move towards most data analysis being done in scripts in R, Python, Julia, etc.

Maintaining the GUI, and in particular our ability to distribute it across operating systems, has taken up a lot of time, certainly well into the 100s of development hours. Even after solving all the major challenges, issues still pop up with OS upgrades. At this point in time my opinion is that most people who will benefit from the kinds of data package management that we do will be doing scripted data analysis and will therefore benefit more from having this time spent on improving core functionality and maintaining wrappers for other languages.

I'm opening this issue for discussion prior to @henrykironde and I making any final decisions.

goelakash:
I think the only thing that the GUI does good is categorizing data, which is a quick and easy way to _find_ datasets.

ethanwhite:
That's a good point @goelakash. It would be good for us to add that functionality to the `retriever ls` command. On it's own `retriever ls` already shows the names of the groups, but it would be good to be able to use it on just a particular group or combination of groups, e.g.,

```
retriever ls animals
retriever ls animals and timeseries
```

davharris:
It should also be reasonably straightforward to make a web page for viewing/filtering the data sets, I think.

Deborah-Digges:
+1 to removing the GUI and adding a web interface.

goelakash:
What's the advantage of a web page over a desktop client?

davharris:
If all the web site does is categorize the data, it wouldn't need a python interpreter, etc. It would also be much more portable than the current GUI.

ethanwhite:
Thanks for the feedback everyone.

This is now officially planned. I'm not sure if @henrykironde is working on this already, but it should be a fairly straightforward, if rather extensive, change. I test ran this a while ago and it basically came down to removing the `app` directory and cleaning up any references to it. This will include cleaning up the docs as well as the code.

#497 should go in before this, but after that I don't see any need to wait further. If @henrykironde isn't working on it and someone else wants to feel free to let us know and we can assign it. @henrykironde if you are working on this already go ahead and assign yourself to the issue.

Deborah-Digges:
[...]

PR #499, opened 2016-06-01
Issue 441: Remove the GUI and all references

Deborah-Digges:
1. Removed the _appdoc folder
2. Removed the graphical parameter to the function `check_for_updates`
3. Removed references of the GUI app from the docs
4. Removed the GUI option from the CLI options
5. Used help as the default option instead of launching the GUI
6. Removed the app folder
7. Made changes to setup.py to remove the check for wx
8. Made changes in setup.py to remove retriever.app as a module
9. Made sure all tests are running
10. Checked installation 
11. Checked a few CLI options(though not exhaustively)

Fixes: https://github.com/weecology/retriever/issues/441


Deborah-Digges:
Open items:

1. `isgui` flag in `__init__.py`. It is used for windows installations. Can I remove it?
2. In `lib/repository.py`, there is a section of code that shows a pop up to a user if the version of retriever that is running is outdated; this popup uses `wx`. For now, I have commented this section. We will have to replace this with a CLI confirmation message

Deborah-Digges:
The AppVeyor build job is failing with this message:

```
Build started
git clone -q https://github.com/weecology/retriever.git C:\projects\retriever
git fetch -q origin +refs/pull/499/merge:
git checkout -qf FETCH_HEAD
Specify a project or solution file. The directory does not contain a project or solution file.
```

Is there some configuration change needed for this?

ethanwhite:
> The AppVeyor build job is failing...

We just turned on AppVeyor yesterday as part of testing #498, so none of the configuration is in yet and you can safely ignore that check.

Deborah-Digges:
oh! thanks @ethanwhite.
@ethanwhite @henrykironde  Do let me know what you think about these changes.

ethanwhite:
This is really excellent @Deborah-Digges! Thanks!!

The PR looks perfect to me, which just leave the two open items you noted:

1. The code on lines 57-64 can be replaced with `DATA_DIR = '.'`. That code handled the fact that installer based GUI on Windows would by default store data in the system path where the Windows installer put the retriever.
2. I actually think the entire `InitThread` class in `lib/repository.py` can be removed. As you can see from the comment on lines 116/117 or 87/88 (depending on new or original version of the file) this functionality as been disabled for a long time. If we're going to have the code start phoning home again to check for updates we'll need to implement it from scratch since the way we're handling releases has changed. So as long as removing `InitThread` doesn't break anything let's just go ahead and remove it.

henrykironde:
@Deborah-Digges, I was going through the code and I think also lib/download.py needs to be revised.
over all we good to merge. We are waiting for some changes from the improve test PR.

Deborah-Digges:
@ethanwhite @henrykironde thank you for reviewing these changes.

I have added the following changes according to @ethanwhite's suggestions:

1.  Removed the module lib/repository.py as the only function being used is check_for_updates, which internally calls InitThread to check for updates in the old way. I have also removed all references to this function
2. I have removed the `isgui` flag in `__init__.py` and also lines 57-64.

ethanwhite:
Looks like that last set of changes broke something according to Travis. It might be passing locally due to cached versions of the dataset scripts. Try running

`retriever reset all`

And then rerunning the tests locally to see whether this still passes for you or not.

Deborah-Digges:
Oh. It looks like the `InitThread` class can't be removed entirely, but only the auto update part. I've made the appropriate changes for this and followed the testing strategy you've mentioned. All tests are passing now 👍

ethanwhite:
Ah, got it. Sorry about pointing you in the wrong direction on that.

This looks ready to go to me. Thanks so much for your help on this! In it goes.

Deborah-Digges:
Thank you for your prompt reviews, Ethan. 👍

PR #501, opened 2016-06-04
Remove string join operation from loop to improve performance

Deborah-Digges:
Moved the string join operation out of the loop


Deborah-Digges:
@henrykironde does this change look OK?

ethanwhite:
Looks good to me.

ethanwhite:
Thanks!

henrykironde:
:+1:

PR #502, opened 2016-06-04
Issue 290: Out of memory error when downloading large dataset

Deborah-Digges:
I've removed the part where the entire file is being read into memory to clean line endings and have opened the file in 'rU' mode instead.

However, the tests for md5 comparisons for csv, mysql and postgres are all failing with this change.

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -154,7 +154,7 @@ class Engine(object):
 
         source = (skip_rows,
                   (self.table.column_names_row - 1,
-                   (open, (file_path, "rb"))))
+                   (open, (file_path, "rU"))))
         lines = gen_from_source(source)
 
         header = next(lines)
@@ -162,7 +162,7 @@ class Engine(object):
 
         source = (skip_rows,
                   (self.table.header_rows,
-                   (open, (file_path, "rb"))))
+                   (open, (file_path, "rU"))))
 
         if not self.table.delimiter:
             self.auto_get_delimiter(header)
@@ -383,20 +383,14 @@ class Engine(object):
 
         return db_name
 
-    def download_file(self, url, filename, clean_line_endings=True):
+    def download_file(self, url, filename):
         """Downloads a file to the raw data directory."""
         if not self.find_file(filename):
             path = self.format_filename(filename)
             self.create_raw_data_dir()
             print("Downloading " + filename + "...")
-            file = urllib.request.urlopen(url)
-            local_file = open(path, 'wb')
-            if clean_line_endings and (filename.split('.')[-1].lower() not in ["exe", "zip", "xls"]):
-                local_file.write(file.read().replace("\r\n", "\n").replace("\r", "\n"))
-            else:
-                local_file.write(file.read())
-            local_file.close()
-            file.close()
+            response = urllib.urlretrieve(url, path)
+            
 
     def download_files_from_archive(self, url, filenames, filetype="zip",
                                     keep_in_dir=False, archivename=None):
@@ -425,21 +419,22 @@ class Engine(object):
             else:
                 self.create_raw_data_dir()
                 if not downloaded:
-                    self.download_file(url, archivename, clean_line_endings=False)
+                    self.download_file(url, archivename)
                     downloaded = True
 
                 if filetype == 'zip':
                     archive = zipfile.ZipFile(archivename)
-                    open_archive_file = archive.open(filename)
+                    open_archive_file = archive.open(filename, 'r')
                 elif filetype == 'gz':
                     # gzip archives can only contain a single file
-                    open_archive_file = gzip.open(archivename)
+                    open_archive_file = gzip.open(archivename, 'r')
                 elif filetype == 'tar':
-                    archive = tarfile.open(filename)
+                    archive = tarfile.open(filename, 'r')
                     open_archive_file = archive.extractfile(filename)
 
                 fileloc = self.format_filename(os.path.join(archivebase,
                                                             os.path.basename(filename)))
+
                 unzipped_file = open(fileloc, 'wb')
                 for line in open_archive_file:
                     unzipped_file.write(line)
@@ -605,7 +600,7 @@ class Engine(object):
         for inserting bulk data from files can override this function."""
         data_source = (skip_rows,
                        (self.table.header_rows,
-                        (open, (filename, 'r'))))
+                        (open, (filename, 'rU'))))
         self.add_to_table(data_source)
 
     def insert_data_from_url(self, url):
@@ -684,7 +679,7 @@ def filename_from_url(url):
 def gen_from_source(source):
     """Returns a generator from a source tuple.
     Source tuples are of the form (callable, args) where callable(*args)
-    returns either a generator or another source tuple.
+    returns either a generator or another source tuple. 
     This allows indefinite regeneration of data sources.
     """
     while isinstance(source, tuple):
--- a/lib/templates.py
+++ b/lib/templates.py
@@ -117,9 +117,10 @@ class DownloadOnlyTemplate(Script):
     def download(self, engine=None, debug=False):
         if engine.name != "Download Only":
             raise Exception("This dataset contains only non-tabular data files, and can only be used with the 'download only' engine.\nTry 'retriever download datasetname instead.")
-        Script.download(self, engine, debug)
-        for filename, url in list(self.urls.items()):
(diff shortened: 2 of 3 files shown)


lib/engine.py
```diff
@@ -426,12 +419,12 @@ def download_files_from_archive(self, url, filenames, filetype="zip",
 
                 if filetype == 'zip':
                     archive = zipfile.ZipFile(archivename)
-                    open_archive_file = archive.open(filename)
+                    open_archive_file = archive.open(filename, 'rU')
```
henrykironde: This looks fine to me. but before changing this. i would recommend dealing with `def download_file(self, url, filename, clean_line_endings=True):`

lib/engine.py
```diff
@@ -379,20 +379,13 @@ def database_name(self, name=None):
...
+    def download_file(self, url, filename):
         """Downloads a file to the raw data directory."""
         if not self.find_file(filename):
             path = self.format_filename(filename)
             self.create_raw_data_dir()
             print "Downloading " + filename + "..."
-            file = urllib.urlopen(url)
-            local_file = open(path, 'wb')
-            if clean_line_endings and (filename.split('.')[-1].lower() not in ["exe", "zip", "xls"]):
-                local_file.write(file.read().replace("\r\n", "\n").replace("\r", "\n"))
```
henrykironde: Thanks @Deborah-Digges, so here we are reading the entire file which is bad, to have a step wise change I suggest first a few ways to go about this,
1) use the old method to download the file and use the new method to download the same file. try to analyze both files and see what is happening.
2) you could also change the file.read() to reading line by line or a chunk read(size) and reprocessing. Travis is a linux based platform so we should not be having much of the problems comparing to windows platforms.

Deborah-Digges:
Hi @henrykironde @ethanwhite 
I was trying to figure out why the tests were failing; so I executed the following commands on my branch to get `output_file_after` and on the master branch to get `output_file_before`.
```
mysql -u travis -Bse "DROP DATABASE IF EXISTS testdb"
retriever install mysql DelMoral2010 -u travis -d testdb
mysqldump testdb -u travis --compact --compatible=no_table_options --no-create-db --no-create-info --result-file=output_file
```

I'm diff-ing the two files, but it seems like there is no difference between the two:
```
diff output_file_before output_file_after 
```

Despite this, the mysql regression tests fail on my new branch, while the tests on the master branch pass. Any idea on what might be causing them to fail?

henrykironde:
I am going to check it in 30 minutes.

henrykironde:
@Deborah-Digges, could you upload the two dumps you have

Deborah-Digges:
- [output_file_after](https://drive.google.com/open?id=0B44W6q42oyuDWG5XWDFZYUl2UG8)
- [output_file_before](https://drive.google.com/open?id=0B44W6q42oyuDMXluWGpnTlkwaWs)

henrykironde:
@Deborah-Digges, the PR #511 may resolve this issue. Please try again later if it is merged.

Deborah-Digges:
Sure 👍 thanks @henrykironde !

Deborah-Digges:
@henrykironde I've pulled in the latest changes and still don't have any luck with these tests. Any idea on what may be the problem ?

henrykironde:
@Deborah-Digges taking a look right now

lib/engine.py
```diff
@@ -383,20 +383,13 @@ def database_name(self, name=None):
...
             print("Downloading " + filename + "...")
-            file = urllib.request.urlopen(url)
-            local_file = open(path, 'wb')
-            if clean_line_endings and (filename.split('.')[-1].lower() not in ["exe", "zip", "xls"]):
-                local_file.write(file.read().replace("\r\n", "\n").replace("\r", "\n"))
-            else:
-                local_file.write(file.read())
-            local_file.close()
-            file.close()
+            urllib.urlretrieve(url, path)
```
henrykironde: This is fine because we are downloading the file in chunks as per the documentation.

lib/engine.py
```diff
@@ -425,17 +418,17 @@ def download_files_from_archive(self, url, filenames, filetype="zip",
...
                     downloaded = True
 
                 if filetype == 'zip':
                     archive = zipfile.ZipFile(archivename)
-                    open_archive_file = archive.open(filename)
+                    open_archive_file = archive.open(filename, 'rU')
                 elif filetype == 'gz':
                     # gzip archives can only contain a single file
-                    open_archive_file = gzip.open(archivename)
+                    open_archive_file = gzip.open(archivename, 'rU')
```
henrykironde: so basically when we read a file with universal newlines ''rU', we are reading it in as text. Python maps all of the characters \r, \n and \r\n to \n. However the retriever also handles binary data and this may be a bug.

henrykironde:
Overall the and from what I envision, reading all files with 'uR' may not be good and i think that is why they introduced that concept of `clean_line_endings`.  So instead, you could probably leave that there and only clean file using read(chunk) and open as  ''rU', and write using newline="\n")

ethanwhite:
@Deborah-Digges - are you only seeing the same failures as the CI is (it looks like just MoM2003) or are you seeing other tests failing as well?

henrykironde:
@Ethan, I think we can change the hashes, I am getting the same results on my local machine, what  i am worried about are the lines that are reading every thing using uR.

henrykironde:
@Deborah-Digges, After trying to replicate what @Ethan was getting on his system, I have finally got an error. 

@Deborah-Digges, To get the same 
uninstall retriever `pip uninstall retriever`
(repeat until you are sure that there is no other version install)

do a reset all
`retriever reset all` 
(removes all the data and scripts)

Install the new branch 
 Either:
       `python setup.py install`
 Or
       `pip install . --upgrade`
 
and try to get csv for mom

`retriever install csv mom2003`

I guess we are soon solving this:
Also take a few minutes and look at these https://pythonconquerstheuniverse.wordpress.com/2011/05/08/newline-conversion-in-python-3/
http://www.shutupandship.com/2012/01/file-modes-in-python-text-binary-and.html

I am not sure if we are doing the downloads with `rU` right

Deborah-Digges:
@henrykironde @ethanwhite Thanks for looking into this! 
It was only the Mom tests that were failing. It seems like something is going wrong with these downloads. Let me look into it.

Thanks @henrykironde for investigating!

Deborah-Digges:
@henrykironde @ethanwhite I've added a few changes and the tests are now passing.
I had to open the file in `rU` mode in a few other places.

ethanwhite:
Nice sleuthing @Deborah-Digges. I did some additional manual testing and unfortunately it looks like this is causing some issues in some other places. I see errors when working with Gentry:

```
ethan@stice:~/RetrieverCode ((248f961...))$ retriever install sqlite gentry -f gentry.sqlite
=> Installing Alwyn H. Gentry Forest Transect Dataset
Downloading gentry_sites_data.txt...
Creating table Gentry_sites...
Downloading all_Excel.zip...es: 226 / 226
Downloading CURUYUQU.xls...
Extracting data from NDAKANNI.xls...
_locate_stream(Workbook): seen
    0  7 7 7 7 7 7 7 7 7 7 7 7 7 0 0 0 0 0 0 0 
   20  0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 
   60= 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 
   80  0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2 3 
Workbook corruption: seen[10] == 7
Done!
```

And also some corruption of data from Bioclim. Those are the only two I've checked, but I expected based on the kinds of data that are (i.e., not simple text data) they could have issues.

Based on a quick experiment it looks like removing the `rU` arguments from the function calls in `download_files_from_archive` may solve these problems. All this function is doing is extracting files from the archives. They are read later by other parts of the code. So, it should be OK to remove these and retest to see if everything is still working.

Deborah-Digges:
@ethanwhite I've made the changes and tested it for gentry and it seems to be working now. 
Does this look ok?

henrykironde:
@Deborah-Digges, This looks fine, I will make one last test in the morning and I will merge if no problems, The main problem was coming from using 'rU' for the data that is not in text format like Bioclim.

henrykironde:
@Deborah-Digges, Thanks for the great efforts on this issue

Deborah-Digges:
Thanks @ethanwhite and @henrykironde for help on this issue!

PR #508, opened 2016-06-06
Extend length of ct_column to 50

### LINKED ISSUE #505: Can't install AmnioteLH dataset to MySQL
Here's the output:

```
retriever install mysql Amniote -u travis -d testdb
=> Installing Amniote life History database
Creating database testdb...
Creating table testdb.main...
INSERT INTO testdb.main (class, ordered, family, genus, species, subspecies, common_name, trait, trait_value) VALUES ('Aves', 'Accipitriformes', 'Accipitridae', 'Accipiter', 'albogularis', null, 'Pied Goshawk', 'litter_or_clutch_size_n', null);
(1406, u"Data too long for column 'trait' at row 1")
Done!
```
The table `main` for this script is also empty in MySQL:
```
mysql> use testdb
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> show tables
    -> ;
+------------------+
| Tables_in_testdb |
+------------------+
| main             |
+------------------+
1 row in set (0.00 sec)

mysql> select * from main;
Empty set (0.00 sec)
```

ethanwhite:
The ct_column column was being set to be a character column of max length 20.
This turned out to be too short in some cases so this extends it to 50.

Fixes #505.

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -177,7 +177,7 @@ class Engine():
             self.auto_get_datatypes(pk, lines, columns, column_values)
 
         if self.table.columns[-1][1][0][:3] == "ct-" and hasattr(self.table, "ct_names") and not self.table.ct_column in [c[0] for c in self.table.columns]:
-            self.table.columns = self.table.columns[:-1] + [(self.table.ct_column, ("char", 20))] + [self.table.columns[-1]]
+            self.table.columns = self.table.columns[:-1] + [(self.table.ct_column, ("char", 50))] + [self.table.columns[-1]]
 
         self.create_table()
 


ethanwhite:
@goelakash can you give this and #507 a review and a test to see if they fix #505 for you as well?

goelakash:
This doesn't, but changing the field to 100 (both for species in `AmnioteLH.script` and for `ct_names` in `engine.py`) allows the MySQL installation.
Tested on 16.04, mysql 5.7.12

goelakash:
The tests are failing though.

henrykironde:
I dont think tests can fail because of these
 changes

henrykironde:
wow Del moral uses Ct_

goelakash:
@henrykironde I always thought that's one of the reasons it was included in the tests, wasn't it?

henrykironde:
Tests should cover most of the parts of the tool,

ethanwhite:
> The tests are failing though.

Yes they were. Even with simple PRs like this where it's tempting to believe they can't influence the tests we should let them run through before merging. It's always worth the extra couple of minutes.

In this case I suspect that the test is probably wrong in the sense that we were probably truncating something in the `ct_column` before, but we should go ahead and try to figure that out and update the hash if needed.

> @henrykironde I always thought that's one of the reasons it was included in the tests, wasn't it?

Yes, Del Moral is there to be a regression test fro cross-tab data.

@goelakash - would you mind running this version on Del Moral, running the version before this on Del Moral, and figuring out what the differences in the resulting data are? If you find that the differences are only something in the `ct_column` being truncated then we can fix the hash in the test. If it's something else then we'll have a better idea of what's going on and can proceed from there. Thanks.

goelakash:
Yeah, it's just the description of the fields that's causing the issue. The old file had species type = TEXT(20), but the new one has TEXT(50). I updated the hash for that.

henrykironde:
So we changed `TEXT(20), to TEXT(50)` I expected all engines to fail on Delmoral, I am wondering how other engines managed to have Delmoral pass.

ethanwhite:
>  I expected all engines to fail on Delmoral, I am wondering how other engines managed to have Delmoral pass.

Presumably they aren't reporting that level of data in the dump. In general I had tried to strip out as much of that as possible so I'm not too surprised that the other engines weren't failing.

henrykironde:
Thanks,the hash tests are passing for me on I will go ahead and merge the current PR and close the revert.

PR #509, opened 2016-06-06
Revert "Extend length of ct_column to 50"

henrykironde:
Reverts weecology/retriever#508
Tests failing


ethanwhite:
Seems unlikely that #508 actually broke anything. More likely that something was getting truncated in the Del Moral data that was incorporated into the original hash. I'd rather check and fix the hash than revert a change that is presumably an improvement.

PR #511, opened 2016-06-07
improve tests platform

henrykironde:
Add appveyor
improve regression tests
remove server dependency by dumping csv files using the excel dialect(function to_csv)
sort files when getting md5sums for files in directory
add xml and json tests
add unit tests for the functions used in testing

### DIFF
--- a/engines/csvengine.py
+++ b/engines/csvengine.py
@@ -1,10 +1,12 @@
 from builtins import str
 from builtins import object
 from builtins import range
+
 import os
-import platform
+
 from retriever.lib.models import Engine, no_cleanup
 from retriever import DATA_DIR
+from retriever.lib.tools import sort_csv
 
 
 class DummyConnection(object):
@@ -99,6 +101,10 @@ class engine(Engine):
         """Check to see if the data file currently exists"""
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
+
+    def to_csv(self):
+        """Export sorted version of CSV file"""
+        return sort_csv(self.table_name())
 
     def get_connection(self):
         """Gets the db connection."""
--- a/engines/jsonengine.py
+++ b/engines/jsonengine.py
@@ -9,6 +9,8 @@ import json
 from retriever.lib.models import Engine
 from retriever import DATA_DIR
 from collections import OrderedDict
+from retriever.lib.tools import json2csv, sort_csv
+
 
 class DummyConnection(object):
     def cursor(self):
@@ -119,6 +121,12 @@ class engine(Engine):
         """Check to see if the data file currently exists"""
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
+
+    def to_csv(self):
+        """Export table from json engine to CSV file"""
+        keys = [columnname[0] for columnname in self.table.columns]
+        csv_outfile = json2csv(self.table_name(), header_values=keys)
+        return sort_csv(csv_outfile)
 
     def get_connection(self):
         """Gets the db connection."""
--- a/engines/sqlite.py
+++ b/engines/sqlite.py
@@ -99,6 +99,10 @@ class engine(Engine):
                 self.existing_table_names.add(line[0].lower())
         return self.table_name(name=tablename, dbname=dbname).lower() in self.existing_table_names
 
+    def to_csv(self):
+        self.connection.text_factory = str
+        Engine.to_csv(self)
+
     def get_connection(self):
         """Gets the db connection."""
         import sqlite3 as dbapi
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -12,8 +12,7 @@ import gzip
 import tarfile
 import urllib
 import csv
-import itertools
-from decimal import Decimal
+
 from retriever import DATA_SEARCH_PATHS, DATA_WRITE_PATH
 from retriever.lib.cleanup import no_cleanup
 from retriever.lib.warning import Warning
@@ -475,6 +474,24 @@ class Engine(object):
             key
             )
             for key in list(script.urls.keys()) if key])
+
+    def to_csv(self):
+        # due to Cyclic imports we can not move this import to the top
+        from retriever.lib.tools import sort_csv
+        csvfile_output = (self.table_name() + '.csv')
+        csv_out = open(csvfile_output, "wb")
+        csv_writer = csv.writer(csv_out, dialect='excel')
+        self.get_cursor()
+        self.cursor.execute("SELECT * FROM " + self.table_name() + ";")
+        row = self.cursor.fetchone()
+        colnames = [tuple_i[0] for tuple_i in self.cursor.description]
+        csv_writer.writerow(colnames)
+        while row is not None:
+            csv_writer.writerow(row)
+            row = self.cursor.fetchone()
+        csv_out.close()
+        self.disconnect()
+        return sort_csv(csvfile_output)
 
     def final_cleanup(self):
         """Close the database connection."""
--- a/test/test_regression_cli.py
+++ b/test/test_regression_cli.py
@@ -1,34 +1,54 @@
 from __future__ import print_function
+import imp
 import os
-from unittest import TestCase
-from hashlib import md5
 
-import nose
+import pytest
+from retriever.lib.tools import getmd5
+from retriever import HOME_DIR, ENGINE_LIST
+
+mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, csv_engine, download_engine, json_engine, xml_engine = ENGINE_LIST()
+dir_path = os.path.dirname(os.path.abspath(__file__))
+
+download_md5 = [
+    # ('DelMoral2010', '0'),
+    ('avianbodysize', 'dce81ee0f040295cd14c857c18cc3f7e'),
+    ('mom2003', 'b54b80d0d1959bdea0bb8a59b70fa871')
+]
+
+csv_md5 = [
+    # ('DelMoral2010', '0'),
+    ('EA_avianbodysize2007', '873597e93bb6163aedd88e285b246c34'),
+    ('EA_mom2003', '5fe8b4f182ba2b8be3bb0ddecdb4673c')
+]
+
+db_md5 = [
+    # ('DelMoral2010', '0'),
+    ('EA_avianbodysize2007', '79680888f7768474479e70c87cd36c9d'),
+    ('EA_mom2003', '520b29f78dc96e8f2d48b5ef9264ff86')
+]
+
+filedb_md5 = [
+    # ('DelMoral2010', '0'),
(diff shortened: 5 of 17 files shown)


henrykironde:
@Deborah-Digges,  @goelakash, @QuLogic! please feel free to review and comment on this PR.

test/test_retriever.py
```diff
@@ -19,6 +20,29 @@
...
+def create_file(data, output='output_file'):
+    """Takes a string and creates a files for tests"""
+    output_file = os.path.normpath(os.path.join(test_dir, output))
+    with open(output_file, 'w') as testfile:
+        testfile.write(data)
+        testfile.close()
+    return output_file
+
+
+# def setup_module():
```
henrykironde: I will edit this

engines/csvengine.py
```diff
@@ -100,6 +97,11 @@ def table_exists(self, dbname, tablename):
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
 
+    def to_csv(self):
+        """Export sorted version of CSV file"""
+        from retriever.lib.tools import sort_csv
```
ethanwhite: Is there a reason for this import to be inside the function instead of at the top of the module? If not let's move it up.

engines/jsonengine.py
```diff
@@ -77,9 +78,6 @@ def disconnect(self):
                 current_output_file.write('\n]')
                 current_output_file.close()
             except:
```
ethanwhite: Given the comment here we can probably just remove the try/except piece of this and just keep the code in the try. I'm OK with not doing that now, but if that's what you want to do then let's not remove the comment since that will make figuring out why the error catching code is there.

engines/jsonengine.py
```diff
@@ -120,6 +118,13 @@ def table_exists(self, dbname, tablename):
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
 
+    def to_csv(self):
+        """Export table from json engine to CSV file"""
+        from retriever.lib.tools import json2csv, sort_csv
```
ethanwhite: Same as above. Imports should be at the top of the module unless there's a good reason not to have them there.

engines/sqlite.py
```diff
@@ -99,6 +99,9 @@ def table_exists(self, dbname, tablename):
                 self.existing_table_names.add(line[0].lower())
         return self.table_name(name=tablename, dbname=dbname).lower() in self.existing_table_names
 
+    def to_csv(self):
+        Engine.to_csv(self)
```
ethanwhite: Seems like this should happen automatically via inheritance. Am I missing something?

henrykironde: yes I  forgot to remove that part.

lib/engine.py
```diff
@@ -476,6 +475,25 @@ def exists(self, script):
             )
             for key in list(script.urls.keys()) if key])
 
+    def to_csv(self):
+        from retriever.lib.tools import sort_csv
```
ethanwhite: Move import to top of module?

lib/engine.py
```diff
@@ -476,6 +475,25 @@ def exists(self, script):
...
+        if self.name is "sqlite":
+            self.connection.text_factory = str
```
ethanwhite: Sorry I didn't think about this before, but is it possible to set this in the `get_connection()` method for `sqlite` instead of using the conditional and setting it here?

henrykironde: very good suggestion.

lib/engine.py
```diff
@@ -476,6 +475,25 @@ def exists(self, script):
...
+        csvfile_output = (self.table_name() + '.csv')
+        self.get_cursor()
+        self.cursor.execute("SELECT * FROM " + self.table_name() + ";")
+        row = self.cursor.fetchone()
```
ethanwhite: Seems like it would be more readable to move this down to be the line right before the `while` loop starts.

test/test_regression_cli.py
```diff
@@ -1,34 +1,54 @@
...
-                 'MoM2003': 'a55c8308722c8e20950e0d1e6d9639e6'}
-}
+
+def get_script_module(script_name):
+    """Load a script module"""
+    file, pathname, desc = imp.find_module(script_name, [os.path.join(HOME_DIR, "scripts")])
+    return imp.load_module(script_name, file, pathname, desc)
+
+
+def dump_csv(dataset, engines):
```
ethanwhite: Since this function finishes by calculating the md5 we should name it something reflects that like `get_csv_md5`. Alternatively we could move the last two lines into the `test_` functions and leave this named as it is.

test/test_regression_cli.py
```diff
@@ -39,116 +59,69 @@ def setup_module():
...
-for dbms in known_md5s:
-    for dataset in known_md5s[dbms]:
-        stub_test = _test_factory('check_%s_regression' % dbms, 'test_%s' % dataset,
-                                  dataset, known_md5s[dbms][dataset])
-        setattr(dbms_test_classes[dbms], stub_test.__name__, stub_test)
-        del(stub_test)
-
-if __name__ == '__main__':
-    nose.runmodule()
+def dump_dir():
```
ethanwhite: Can this code be moved to pytest's `setup_function()` and then the explicit execution of `dump_dir` removed from line 44?

test/test_retriever.py
```diff
@@ -19,6 +20,29 @@
                                        shortname='test')
 test_engine.opts = {'database_name': '{db}_abc'}
 HOMEDIR = os.path.expanduser('~')
+test_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+def create_file(data, output='output_file'):
+    """Takes a string and creates a files for tests"""
```
ethanwhite: "Takes a string and creates a files for tests" -> "Writes a string to a file for use by tests"

test/test_retriever.py
```diff
@@ -19,6 +20,29 @@
...
+def setup_module():
+    """"change directory to test directory"""
+    dir_path = os.path.dirname(os.path.abspath(__file__))
+    os.chdir(dir_path)
+
+
+def teardown_method():
+    """Cleanup temporary output files after testing"""
+    os.system("rm -r output_*")
+    os.system("rm -r raw_data/MoM2003")
```
ethanwhite: The `raw_data/MoM2003` directory shouldn't be getting created in this test module should it? If not then we can remove this line.

test/test_retriever.py
```diff
@@ -19,6 +20,29 @@
...
+
+def setup_module():
+    """"change directory to test directory"""
+    dir_path = os.path.dirname(os.path.abspath(__file__))
+    os.chdir(dir_path)
+
+
+def teardown_method():
+    """Cleanup temporary output files after testing"""
+    os.system("rm -r output_*")
```
ethanwhite: It looks like some of these tests are also creating files of the form `output.*`, so should this line be changed to `os.system("rm -r output*")`?

test/test_retriever.py
```diff
@@ -113,8 +141,11 @@ def test_find_file_present():
 
     """
     test_engine.script.shortname = 'AvianBodySize'
-    os.chdir('./test/')
-    assert test_engine.find_file('avian_ssd_jan07.txt') == os.path.normpath('raw_data/AvianBodySize/avian_ssd_jan07.txt')
+    # os.chdir('test')
+    dir_path = os.path.dirname(os.path.abspath(__file__))
+    os.chdir(dir_path)
```
ethanwhite: Do these two lines need to be called again since they are called in `setup_module()`?

test/test_retriever.py
```diff
@@ -149,10 +180,55 @@ def test_format_insert_value_string_simple():
 
 def test_format_insert_value_string_complex():
     """Test formatting of values for insert statements"""
-    assert test_engine.format_insert_value('my notes: "have extra, stuff"', 'char') == '\'my notes: \\"have extra, stuff\\"\''
+    assert test_engine.format_insert_value('my notes: "have extra, stuff"',
+                                           'char') == '\'my notes: \\"have extra, stuff\\"\''
 
 
 def test_getmd5():
```
ethanwhite: Since we have two of these functions let's specify the differences in the names: `test_getmd5_lines` and `test_getmd5_path`.

test/test_retriever.py
```diff
@@ -149,10 +180,55 @@ def test_format_insert_value_string_simple():
...
-    """Test md5 sum calculation"""
+    """Test md5 sum calculation given a line"""
+    lines = ['a,b,c\n', '1,2,3\n', '4,5,6\n']
+    assert getmd5(data=lines, data_type='lines') == '0bec5bf6f93c547bc9c6774acaf85e1a'
+
+
+def test_getmd5():
+    """Test md5 sum calculation given a path to data source"""
```
ethanwhite: The docstring doesn't match the code here and the code is the same as in the function above.

ethanwhite:
This looks really great @henrykironde! I've made some minor cleanup suggestions in-line.

The one more major suggestion relates tests you've added to the end of `test_retriever.py`. In general I don't think we want to use hashes for unit tests like these, because we can't easily see if the hash represents the answer we want or not. Instead we want to compare the resulting file directly to what it should look like as a string. There are some examples of this in `test_integration.py`.

Thanks again for all your hard work on this. It's getting super close.

ethanwhite:
Regarding the one major suggestion, I'd also be happy if you wanted to pull those last few tests from `test_retriever` to get this merged since they aren't crucial to the main goal of the PR. In that case just let me know when the minor cleanup is done and we'll get this in.

test/test_regression_cli.py
```diff
@@ -1,34 +1,54 @@
-from __future__ import print_function
```
goelakash: This is needed for py3.

test/test_retriever.py
```diff
@@ -1,16 +1,19 @@
 """Tests for the EcoData Retriever"""
-from future import standard_library
-standard_library.install_aliases()
 
 import os
-from io import StringIO
```
goelakash: Needed for py3.

test/test_retriever.py
```diff
@@ -1,16 +1,19 @@
 """Tests for the EcoData Retriever"""
-from future import standard_library
```
goelakash: Again, py3.

test/test_regression_cli.py
henrykironde: @goelakash, yes I have seen that even in json some imports are missing, I do have them on my system, I have no idea how they have been removed. when I committed. Lately my python environment had been a stranger to me :)

lib/engine.py
henrykironde: taking this up is erroring.
May be i am missing something.

henrykironde:
Thanks @goelakash and @ethanwhite . Let me know if i missed something. I am going to redo a cross check after lunch

ethanwhite:
Awesome work on this @henrykironde! In it goes. :tada:

PR #514, opened 2016-06-09
Cleanup of temp test files

goelakash:


test/test_regression_cli.py
```diff
@@ -59,9 +59,12 @@ def setup_module():
 
 def teardown_module():
     """Cleanup temporary output files after testing and return to root directory"""
+    os.system("rm -r output*")
     os.system("rm output*")
```
henrykironde: remove the old ones

henrykironde:
:+1:

test/test_regression_cli.py
goelakash: I couldn't remove `output_json.csv` and `output_xml.csv` from these commands. Any idea how to remove those as well?

test/test_retriever.py
```diff
@@ -36,6 +36,7 @@ def setup_module():
 def teardown_method():
     """Cleanup temporary output files after testing"""
     os.system("rm -r output*")
+    os.system("rm output*")
```
henrykironde: remove this too

henrykironde:
try using os.remove()

goelakash:
Didn't work.

henrykironde:
okey there may be an error in the code in test/test_retriever.py   `def test_xml2csv()` i wiill check it out and clarify

henrykironde:
is it currently working ?
if not : 

In test/test_retriever.py we trying to set up the testing by os/chir(dir_path)
but it looks like it does not change to that directory 

so what I think is `def test_xml2csv()` and `test_json2csv()` are run from the main retriever directory a go head and  try to use os.system("rm -r test/output*") in the Tear down 

to test the directory you can use this , change assert to` assert os.getcwd() ==""`

henrykironde:
on more solution is to try deleting those files right before the assert statements.

henrykironde:
@goelakash is this working as required?

goelakash:
@henrykironde Let me get back to you in a few hours.

goelakash:
@henrykironde Yes, it works now.

PR #516, opened 2016-06-10
Python3

goelakash:
Added absolute imports (which are default on python 3).
Needed to add symlinks to cloned repos from `lib\site-packages\` to for the package `retriever` to be found.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -5,6 +5,7 @@ download published ecological data, and store the data in a database.
 
 """
 from __future__ import print_function
+from __future__ import absolute_import
 from builtins import str
 
 import os
@@ -13,8 +14,8 @@ from os.path import join, isfile, getmtime, exists
 import imp
 import platform
 
-from lib.compile import compile_script
-from _version import __version__
+from retriever.lib.compile import compile_script
+from retriever._version import __version__
 
 current_platform = platform.system().lower()
 if current_platform != 'windows':
--- a/__main__.py
+++ b/__main__.py
@@ -7,8 +7,9 @@ The main() function can be used for bootstrapping.
 
 """
 from __future__ import print_function
+from __future__ import absolute_import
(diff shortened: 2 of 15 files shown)


goelakash:
@henrykironde Figured that only VERSION was needed from `retriever.__init__.py` before installation of `retriever` package, so moved that to `setup.py`. Otherwise the build fails on travis and appveyor.

henrykironde:
@goelakash, Please, could you clean up the commits for this PR and edit the description to be more clear.

goelakash:
@henrykironde Yes, but I wanted to have your opinion on whether this is the right way to store the `VERSION` variable. Keeping two copies and updating them both seems not to be a good idea to me.

I am also considering moving the entire retriever package code to a sub-dir called `retriever` and test those changes. This is because the `setuptools.setup` function for python3 seems to have a problem in correctly detecting the package in the same folder as `setup.py`.

henrykironde:
@goelakash, I guess you mean something like this `http://www.diveintopython3.net/packaging.html#structure`
Please, go ahead if this structure works as default for the python3 shift and correctly detecting the package locations.
About the version, having one copy is definitely a good idea as long as we can access effectively from there, I think for now we can go with what you have suggested.

goelakash:
@henrykironde #525 is the solution suggested by @ethanwhite . Guess I should rebase on that once that is merged?

henrykironde:
@goelakash, I think we can close the other one.
We then include the suggested changes in a new branch or even here. 
then squash the changes to represent something like changing setup module to create version module or anything that is appropriate.

ethanwhite:
It doesn't matter to me which PR it goes through, but I'd like to get the change in #525 in quickly since we basically have a broken build at the moment (i.e., #522).

goelakash:
@henrykironde Please review this.

PR #524, opened 2016-06-14
remove space from comment

henrykironde:
for some reason the  space in the comment causes error 
`#retriever`
`# retriever`


ethanwhite:
Why have the tests continued to pass every time if this script was repeatedly broken?

ethanwhite:
Ah right, we turned off the DelMoral tests in the testing revision. We shouldn't definitely get those turned back on when we get the chance.

henrykironde:
yah delmoral is very intersting a data set,  but I am still rectifying some more issues to get it back in.

Issue #522, opened 2016-06-13
Install error due to missing `builtins` module


ethanwhite:
From https://github.com/ropensci/ecoretriever/issues/66#issuecomment-225685553

```
$ pip install git+https://github.com/weecology/retriever --upgrade
Collecting git+https://github.com/weecology/retriever
  Cloning https://github.com/weecology/retriever to c:\users\hurlbert\appdata\local\temp\pip-xsdwwn-build
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "c:\users\hurlbert\appdata\local\temp\pip-xsdwwn-build\setup.py", line 25, in <module>
        from __init__ import VERSION
      File "__init__.py", line 8, in <module>
        from builtins import str
    ImportError: No module named builtins
```

I'm guessing this means we missed adding `builtins` as a requirement somewhere in the Python 2/3 compatibility work. @goelakash can you take a look at this?

goelakash:
It requires the `future` package to be already installed.
We could add these lines before the import from `__init__`:
```
import pip
pip.main(['install', 'future'])
```

ethanwhite:
I think it probably just needs to get added to `install_requires` in `setup.py` so that `pip` installs it automatically. We should probably also add it to the `includes` list to make sure that it gets added to the installers.

goelakash:
Because I import `VERSION` before the setup function is invoked, that won't work, because `__init__` gets imported before anything is installed.
If we move `VERSION` to a separate file, then it will be ok to add those packages to the required lists.

ethanwhite:
Ah, yes, you're totally right. I've done some looking around trying to find the standard place to store version info and apparently there really isn't one.

I kind of like what @jiffyclub has done in `snakeviz`:
https://github.com/jiffyclub/snakeviz/search?utf8=%E2%9C%93&q=version&type=Code

But since we already have a `version.py` that is doing something else we'd need to go with `_version.py` (which is standard name for a file storing this information (see http://stackoverflow.com/a/7071358/133513).

So, this would move `VERSION` into `setup.py` so we wouldn't need to import `__init__` there. `setup.py` would generate `_version.py` and `__init__` would import `_version.py` to get the version number. If we go this route then we should honor the standard and use `__version__` for the import into `version.py`.

What do you think?

goelakash:
Yes, generating a `__version__.py` module via `setup.py` seems to be a good idea. I will send a PR shortly.

[...]

PR #525, opened 2016-06-14
Moved VERSION to setup.py

goelakash:
and write its value to a separate __version__.py file.
#522 
@ethanwhite @henrykironde

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -14,12 +14,13 @@ import imp
 import platform
 
 from lib.compile import compile_script
+from __version__ import VERSION
 
 current_platform = platform.system().lower()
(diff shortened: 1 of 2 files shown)


ethanwhite:
This is the right basic idea, but the convention is the have the member variable be `__version__`, so we should set this up so that the import in `__init__` is `from _version import __version__`.

@henrykironde - does this look like a reasonable general solution to you?

henrykironde:
yes this is a good solution to keep the setup module and project source

henrykironde:
This looks fine to me .

ethanwhite:
We should still update this to match the standard.

PR #527, opened 2016-06-15
revert delmoral to simple script, and add ignore Ascii to the failing…

henrykironde:
after testing on delmoral I realized that it was failing on Json because by default json dump() tries to alter to ascii.


henrykironde:
With the change the files are downloading as expected.We shall go ahead and confirm the hashes  after converting to_csv.

ethanwhite:
:+1:

PR #533, opened 2016-06-16
Unicode changes for python3 compatibility

goelakash:
1. open all files in non-binary mode by default.
2. Use io.BytesIO() for py2 and io.StringIO() for py3 for csv.writer
3. Update getmd5 to handle 'bytes' type class in py3.

All the tests pass for python 2 and for python 3 locally on Ubuntu 16.04 (64 bit).
@henrykironde @ethanwhite

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -16,7 +16,7 @@ import sys
 # sys removes the setdefaultencoding method at startup; reload to get it back
 reload(sys)
 if hasattr(sys, 'setdefaultencoding'):
-    # set default encoding to latin-1 to avoid ascii encoding issues
+    # set default encoding to utf-8 to avoid ascii encoding issues
     sys.setdefaultencoding('utf-8')
 from retriever import VERSION, MASTER, SCRIPT_LIST, sample_script, current_platform
 from retriever.engines import engine_list
--- a/__main__.py
+++ b/__main__.py
@@ -17,7 +17,7 @@ import sys
 reload(sys)
 if hasattr(sys, 'setdefaultencoding'):
     # set default encoding to latin-1 to avoid ascii encoding issues
-    sys.setdefaultencoding('latin-1')
+    sys.setdefaultencoding('utf-8')
 from retriever import VERSION, MASTER, SCRIPT_LIST, sample_script, current_platform
 from retriever.engines import engine_list
 from retriever.lib.repository import check_for_updates
--- a/engines/postgres.py
+++ b/engines/postgres.py
@@ -1,5 +1,5 @@
 import os
-
+from builtins import str
 from retriever.lib.models import Engine, no_cleanup
 
 
@@ -96,8 +96,6 @@ CSV HEADER"""
     def insert_statement(self, values):
         """Returns a SQL statement to insert a set of values"""
         statement = Engine.insert_statement(self, values)
-        if isinstance(statement, basestring):
-            statement = statement.decode("utf-8", "ignore")
         return statement
 
     def table_exists(self, dbname, tablename):
--- a/engines/postgres.py
+++ b/engines/postgres.py
@@ -96,6 +96,8 @@ CSV HEADER"""
     def insert_statement(self, values):
         """Returns a SQL statement to insert a set of values"""
         statement = Engine.insert_statement(self, values)
+        if isinstance(statement, bytes):
+            statement = statement.decode("utf-8", "ignore")
         return statement
 
     def table_exists(self, dbname, tablename):
--- a/lib/compile.py
+++ b/lib/compile.py
@@ -25,7 +25,7 @@ def compile_script(script_file):
     keys_to_ignore = ["template"]
 
     for line in [line.strip() for line in definition]:
-        if line != "" and ':' in line and not line[0] == '#':
+        if line and ':' in line and not line[0] == '#':
             split_line = [a.strip() for a in line.split(":")]
             key = split_line[0].lower()
             value = ':'.join(split_line[1:])
--- a/lib/compile.py
+++ b/lib/compile.py
@@ -15,7 +15,7 @@ SCRIPT = HtmlTableTemplate(%s)""",
 
 
 def compile_script(script_file):
-    definition = open(script_file + ".script", 'rb')
+    definition = open(script_file + ".script", 'r')
 
     values = {}
     urls = {}
@@ -25,7 +25,7 @@ def compile_script(script_file):
     keys_to_ignore = ["template"]
 
     for line in [line.strip() for line in definition]:
-        if line and ':' in line and not line[0] == '#':
+        if line != "" and ':' in line and not line[0] == '#':
             split_line = [a.strip() for a in line.split(":")]
             key = split_line[0].lower()
             value = ':'.join(split_line[1:])
@@ -134,7 +134,7 @@ def compile_script(script_file):
         template = "default"
     script_contents = (script_templates[template] % script_desc)
 
-    new_script = open(script_file + '.py', 'wb')
+    new_script = open(script_file + '.py', 'w')
     new_script.write(script_contents)
     new_script.close()
 
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -277,12 +277,12 @@ class Engine(object):
             thistype = self.datatypes[thistype]
 
             if isinstance(thistype, tuple):
-                if len(datatype) > 1 and datatype[1] > 0:
+                if len(datatype) > 1 and int(datatype[1]) > 0:
                     thistype = thistype[1] + "(" + str(datatype[1]) + ")"
                 else:
                     thistype = thistype[0]
             else:
-                if len(datatype) > 1 and datatype[1] > 0:
+                if len(datatype) > 1 and int(datatype[1]) > 0:
                     thistype += "(" + str(datatype[1]) + ")"
         else:
             thistype = ""
@@ -393,7 +393,7 @@ class Engine(object):
             self.create_raw_data_dir()
             print("Downloading " + filename + "...")
             response = urllib.request.urlretrieve(url, path)
-            
+
 
     def download_files_from_archive(self, url, filenames, filetype="zip",
                                     keep_in_dir=False, archivename=None):
@@ -438,7 +438,7 @@ class Engine(object):
                 fileloc = self.format_filename(os.path.join(archivebase,
                                                             os.path.basename(filename)))
 
-                unzipped_file = open(fileloc, 'wb')
+                unzipped_file = open(fileloc, 'w')
                 for line in open_archive_file:
                     unzipped_file.write(line)
                 open_archive_file.close()
@@ -477,8 +477,8 @@ class Engine(object):
         # due to Cyclic imports we can not move this import to the top
         from retriever.lib.tools import sort_csv
         csvfile_output = (self.table_name() + '.csv')
-        csv_out = open(csvfile_output, "wb")
-        csv_writer = csv.writer(csv_out, dialect='excel')
+        csv_out = open(csvfile_output, "w")
+        csv_writer = csv.writer(csv_out, dialect='excel', lineterminator='\n')
         self.get_cursor()
         self.cursor.execute("SELECT * FROM " + self.table_name() + ";")
         row = self.cursor.fetchone()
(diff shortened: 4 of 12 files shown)


ethanwhite:
Congrats on being almost there! Did you want us to take a look now or should we wait until you have some time to run down the failing Windows tests?

goelakash:
@ethanwhite Let me locally test the windows build.

henrykironde:
@goelakash, as we figure out what is happening to windows, let's add python3 to the matrix on languages in Travis https://docs.travis-ci.com/user/languages/python

henrykironde:
@goelakash ,perfect work...adding python3.5 on Travis.

henrykironde:
If you have a windows system, test only one engine and one data set. you will just comment out the other datasets. For example if you test EA_mom2003
```
 db_md5 = [
    # ('DelMoral2010', '0'),
    # ('EA_avianbodysize2007', '79680888f7768474479e70c87cd36c9d'),
    ('EA_mom2003', '520b29f78dc96e8f2d48b5ef9264ff86')
]
```
Also comment out the lines that delete the data after every test.
Then to test the engines you could either comment others out of use 

`py.test -k "test_postgres_regression" -v `

This will test only the postgres engine.

Once you have the data let me know we can examine the files dumped on windows and linux

henrykironde:
also you are using ` UTF-8`
in main we set encoding to Latin

```
import sys
# sys removes the setdefaultencoding method at startup; reload to get it back
reload(sys)
if hasattr(sys, 'setdefaultencoding'):
    # set default encoding to latin-1 to avoid ascii encoding issues
    sys.setdefaultencoding('latin-1')
```
I don't know if that is required to change encoding.


Having said that, recently I found out that the test module have no encoding set. If you add encoding to the tests modules like in main above, they may pass because you are reading in mode 'rb' by default.

goelakash:
@henrykironde I have solved the conflicts and squashed the commits. Needed to add os checks for `csv.writer` as it adds extra `\r` character when `csv.writer.writerows()` is used.
[StackOverflow](http://stackoverflow.com/a/17725590/3285138)

lib/repository.py
```diff
@@ -101,7 +101,7 @@ def run(self):
             scripts = []
             print("Downloading scripts...")
             for line in version_file:
-                scripts.append(line.strip('\n').split(','))
+                scripts.append(line.decode("utf-8").strip('\n').split(','))
```
henrykironde: Any reason why we are using utf-8 here?

lib/tools.py
```diff
@@ -295,7 +299,10 @@ def getmd5(data, data_type='lines', mode='rb'):
     for file_path in files:
         lines = open(file_path, mode)
         for line in lines:
-            checksum.update(line)
+            if type(line) == bytes:
+                checksum.update(line)
+            else:
+                checksum.update(str(line).encode("utf-8"))
```
henrykironde: and here 'utf-8'

lib/repository.py
goelakash: The string needs to be unicode, but the urlopen function gives bytes.
We could also decode it using `latin-1` if that is the default behaviour needed.

lib/tools.py
goelakash: same reason, need unicode instead of bytes.

lib/repository.py
henrykironde: I think `line.decode().strip('\n').split(',')) ` would work. Since Latin-1 is the default.

goelakash:
#487
@henrykironde Why are we using `latin-1` instead of `utf-8`?

lib/repository.py
goelakash: We can't keep `latin-1` as default as Python 3 keeps `utf-8` as default encoding for source and raises an error if that is changed.

henrykironde:
"Why are we using latin-1 instead of `utf-8`? `utf-8` can also represent charsets that latin-1 can't handle.",
I for one wanted to change to `utf-8` but since `Latin-1` was chosen, I was hesitant for a change.
However we should stick to one. So u either change to `utf-8` in the main or keep `Latin-1`.  which ever does the job.

lib/repository.py
henrykironde: what is the error message?

henrykironde:
you could change the whole project to use `utf-8 `.

@goelakash Let me know once you are done with changes to the PR. great work

lib/repository.py
goelakash: There is no error, just that `sys.setdefaultencoding` function is unavailable in python3.
https://docs.python.org/3/library/sys.html

goelakash:
@henrykironde The changes are in.

henrykironde:
@goelakash, do we still need to explicitly give decode("utf-8") and encode("utf-8") with arguments?
you can call it with no argument since you set the encoding to `utf` or get the encoding with `sys.getdefaultencoding()`

goelakash:
@henrykironde You're right. I changed it accordingly.

lib/engine.py
```diff
@@ -477,8 +477,11 @@ def to_csv(self):
         # due to Cyclic imports we can not move this import to the top
         from retriever.lib.tools import sort_csv
         csvfile_output = (self.table_name() + '.csv')
-        csv_out = open(csvfile_output, "wb")
-        csv_writer = csv.writer(csv_out, dialect='excel')
+        csv_out = open(csvfile_output, "w")
+        if os.name == 'nt':
+            csv_writer = csv.writer(csv_out, dialect='excel', lineterminator='\n')
+        else:
```
henrykironde: I think the `if else` part can go away and we just use c`sv_writer = csv.writer(csv_out, dialect='excel', lineterminator='\n')`

lib/tools.py
```diff
@@ -328,9 +335,12 @@ def sort_csv(filename):
     infields = next(csv_reader_infile)
 
     #  write the data to a temporary file and sort it
-    file_temp = open(os.path.normpath("tempfile"), 'wb')
+    file_temp = open(os.path.normpath("tempfile"), 'w')
 
-    csv_writer = csv.writer(file_temp, dialect='excel', escapechar='\\')
+    if os.name == 'nt':
```
henrykironde: @goelakash,  I think these two checks should also go if we do not need them, and we can merge. 👍

goelakash: I don't think we can remove those before correcting the hashes. The `db_md5

henrykironde: yes go ahead and correct the hashes

henrykironde: @goelakash, I think if we do this if may work.
```
    file_temp = open(os.path.normpath("tempfile"), 'w')
    csv_writer = csv.writer(file_temp, dialect='excel', escapechar='\\', lineterminator='\n')
    for row in csv_reader_infile:
        csv_writer.writerow(row)
    file_temp.close()
    input_file.close()

    # sort the temp file
    sorted_txt = sort_file(os.path.normpath("tempfile"))

    # write sorted row content to csv filename with header "infields"
    tmp = open(sorted_txt, "rU")
    in_txt = csv.reader(tmp, delimiter=',')
    out_csv = csv.writer(open(filename, 'w'), lineterminator='\n')

    out_csv.writerow(infields)
    out_csv.writerows(in_txt)
    tmp.close()
    os.remove(os.path.normpath("tempfile"))
    return filename
```

henrykironde: Test that and let me know

goelakash: Doesn't work. At the moment I think this issue is pertinent to python3, so I think a separate issue (if needed) can be created. Although I'm totally ok with having Carriage returns in the file.
@ethanwhite

__main__.py
```diff
@@ -17,7 +17,7 @@
 reload(sys)
 if hasattr(sys, 'setdefaultencoding'):
     # set default encoding to latin-1 to avoid ascii encoding issues
```
ethanwhite: We should update this comment as well.

engines/postgres.py
```diff
@@ -96,8 +96,6 @@ def insert_data_from_file(self, filename):
     def insert_statement(self, values):
         """Returns a SQL statement to insert a set of values"""
         statement = Engine.insert_statement(self, values)
-        if isinstance(statement, basestring):
-            statement = statement.decode("utf-8", "ignore")
```
ethanwhite: Is the need for these lines removed by the new default encoding? (to be honest I'm not sure why they are there, but just wanted to make sure we weren't breaking an edgecase somewhere)

lib/compile.py
```diff
@@ -25,7 +25,7 @@ def compile_script(script_file):
     keys_to_ignore = ["template"]
 
     for line in [line.strip() for line in definition]:
-        if line and ':' in line and not line[0] == '#':
+        if line != "" and ':' in line and not line[0] == '#':
```
ethanwhite: Is there a reason for making this change? I thought `if line` was generally preferred for this sort of thing.

lib/table.py
```diff
@@ -110,7 +112,11 @@ def combine_on_delimiter(self, line_as_list):
         """Combine a list of values into a line of csv data"""
         dialect = csv.excel
         dialect.escapechar = "\\"
-        writer_file =  io.BytesIO()
+        if sys.version_info >= (3, 0):
+            writer_file =  io.StringIO()
+        else:
+            writer_file =  io.BytesIO()
```
ethanwhite: @goelakash - can you walk me through the difference in `io.StringIO()` and `io.BytesIO()` between Python 2 & 3?

lib/tools.py
```diff
@@ -281,9 +281,13 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
 def getmd5(data, data_type='lines', mode='rb'):
     """Get MD5 of a data source"""
     checksum = md5()
+    # mode = 'rb'
```
ethanwhite: Looks like an old debugging line that should be deleted.

ethanwhite:
Sorry to be late to the party. I've gone through the PR, though I think maybe it changed a little as I was working.

I have two major questions:

1. I can't quite follow the current state of the line endings issue/discussion. Can someone give me a short summary of where we are at and what we've tried. It did seem from a naive look at the code that we could use `out_csv = csv.writer(open(filename, 'w'), lineterminator='\n')` for all operating systems, but I suspect that both @henrykironde and I are missing something here.

2. I also don't understand yet the work being done with `io.StringIO()` vs `io.BytesIO()`. In `combine_on_delimiter()` this happens differently for Python 2 vs Python 3, and (if I'm understanding thigns correctly) then this cascades to create the new logic in `getmd5()` about checking if `type(line) == bytes`. All of this logic doesn't read very naturally. I'm assuming this is something we need for 2/3 compatibility that I don't understand, and given that I'm happy to include this. I'd just like to understand `io.StringIO()` vs `io.BytesIO()` a little more.

Thanks for all your hard work on this @goelakash! It's looking great and is really close to going in.

engines/postgres.py
goelakash: `basestring` is a parent of `str` and `unicode` in python 2. There is no `basestring` in Python 3, and the strings are `unicode` by default. 
Since the `decode()` function converts `bytes` to `str`, and the `str` are already unicode as I import `builtins.str` at the beginning, that means that I should've replaced `basestring` with `bytes` here.

lib/table.py
goelakash: @ethanwhite This code gives an error in Python3:
```
line_as_list = [line.encode() for line in line_as_list]
writer_file =  io.BytesIO()
writer = csv.writer(writer_file, dialect=dialect, delimiter=self.delimiter)
for line in line_as_list:
    assert isinstance(line,bytes)
    writer.writerow(line)```
Error: 
```
>           writer.writerow(line)
E           TypeError: a bytes-like object is required, not 'str'
```
So I had to use `io.StringIO` for Python3. I have no idea why csv can't understand `bytes` type when it sees it (as the assert statement passes without error).

henrykironde:
@goelakash, @ethanwhite I am okey with merging this PR if we all think the changes are final
@goelakash, thanks for porting p3. 

just a something we could think about later, if need this line 
https://github.com/goelakash/retriever/blob/3e925abf2084c703a39840b8d95488644a464904/lib/engine.py#L90

lib/table.py
ethanwhite: @goelakash you're totally right. it seems strange that it works this way, but clearly it does. Thanks for the explanation.

ethanwhite:
Awesome work on this @goelakash and thanks for taking the time to explain the complexities of the 2/3 compatibility to both of us. Based on reading it after the fact I can only imagine that it must have been quite challenging to get everything working in the first place. :+1: :tada: :smile:

goelakash:
@ethanwhite @henrykironde Thanks, it was well worth the effort.

PR #540, opened 2016-06-27
utf-8 encode and decode 

goelakash:
Encoding the text as unicode, ignoring the characters which are non-utf8, and then decoding back seems to alleviate the problem. Downside is that now we lose certain characters like Ö and É.
Status: in progress

@ethanwhite @henrykironde

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -15,7 +15,7 @@ import gzip
 import tarfile
 import urllib.request, urllib.parse, urllib.error
 import csv
-
+import io
 from retriever import DATA_SEARCH_PATHS, DATA_WRITE_PATH
 from retriever.lib.cleanup import no_cleanup
 from retriever.lib.warning import Warning
@@ -157,7 +157,7 @@ class Engine(object):
 
         source = (skip_rows,
                   (self.table.column_names_row - 1,
-                   (open, (file_path, "rU"))))
+                   (io.open, (file_path, "rU", -1, 'utf-8', 'ignore'))))
         lines = gen_from_source(source)
 
         header = next(lines)
@@ -165,7 +165,7 @@ class Engine(object):
 
         source = (skip_rows,
                   (self.table.header_rows,
-                   (open, (file_path, "rU"))))
+                   (io.open, (file_path, "rU", -1, 'utf-8', 'ignore'))))
 
         if not self.table.delimiter:
             self.auto_get_delimiter(header)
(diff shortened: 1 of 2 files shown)


ethanwhite:
I like the idea of adding new tests here, but I think a better way to handle it would be to:

1. Add an integration test in `test_integration.py` that has some of the relevant unicode characters
2. Potentially add a unit test that sends those characters directly to the function that was throwing the error

In general we don't want tests that will fail inconsistently and for unclear reasons, which I think is what randomly selected datasets and seeing if they execute will do. The approach in this PR would also make for a pretty lengthy test run, which can be problematic as devs tend to run tests less frequently when they are slow.

I do think that, separately from this PR, it would be great to have a system that can be used to run all of the datasets and check for errors (and ultimately changes in the output). This is something that @henrykironde and I have talked about but haven't had time to fully implement. We do have `try_install_all.py` which can be used (if it's still operational) to do these this kind of testing locally after a major change. In fact, it would be good to run this now under both Python 2 & 3 and open any issues for datasets that are erroring. In general we may want `try_install_all.py` to skip the really big datasets (e.g., FIA and vertnet) since they can take hours to run.

goelakash:
@ethanwhite The tests that I added were wrong anyways :joy: 
I found the problem to be mainly with reading of the source, which we were doing with the low-level `open()` function. I replaced that that with `io.open()` which has a consistent API across py2/3.

Also, I checked DelMoral tests with these changes. Once the CI tests pass, I think this can be merged, pending further review.

lib/engine.py
```diff
@@ -157,15 +157,15 @@ def auto_create_table(self, table, url=None, filename=None, pk=None):
 
         source = (skip_rows,
                   (self.table.column_names_row - 1,
-                   (open, (file_path, "rU"))))
+                   (io.open, (file_path, "rU", -1, 'utf-8', 'ignore'))))
```
goelakash: -1 -> default buffer argument value.
https://docs.python.org/2/library/io.html#io.open

ethanwhite:
Looks good. Thanks! I'm going to go ahead and merge this.

I would love to see an integration test involving unicode if you get the chance since we run into unicode issues frequently and it would be good to avoid regressions with it in the future.

goelakash:
@ethanwhite Okay, I suppose I need to experiment with `subprocess` to come up with something.

ethanwhite:
@henrykironde ah, I hadn't understood that this was what `ignore` was doing. I completely agree that we shouldn't be dropping characters. Sorry for not understanding this when I merged.

PR #543, opened 2016-06-30
Change names of JSON files to shortname.json instead of their .script…

goelakash:
… names

This is useful as it helps to
1. quickly search for the required datapackage.json
file using the `retriever ls` from the terminal.
2. find the datapackage.json file for editing by the CLI tool.


goelakash:
@henrykironde

PR #544, opened 2016-06-30
Indexing

henrykironde:
add indexing to the engines and resolve the issues of the current indexing for file database systems

There was no indexing for sqlite, postgres and mysql. Current indexing for csv json and xml, was not resetting the index for the tables.

 set default encoding for test modules to avoid ASCII related issues.

 change the test dump directory to pytest's temporary directory


henrykironde:
@ethanwhite, could the conflicts be the reason for the failures?

engines/sqlite.py
```diff
@@ -10,13 +10,13 @@ class engine(Engine):
...
     datatypes = {
-        "auto": "INTEGER",
+        "auto": ("INTEGER", "AUTOINCREMENT"),
         "int": "INTEGER",
         "bigint": "INTEGER",
         "double": "REAL",
         "decimal": "REAL",
         "char": "TEXT",
-        "bool": "INTEGER",
+        "bool": "REAL",
```
ethanwhite: Seems like boolean should stays as "INTEGER" in SQLite:
https://www.sqlite.org/datatype3.html#section_2_1

henrykironde: I will check that

lib/engine.py
```diff
@@ -365,7 +361,7 @@ def create_table_statement(self):
 
         create_stmt += ', '.join(column_strings)
         create_stmt += " );"
-
+        print (create_stmt)
```
henrykironde: I this will go

ethanwhite:
Replaced by #546.

Issue #246, opened 2014-11-09
Auto-numbered primary keys should start from 1 for each table


ethanwhite:
In some cases when generating auto-numbered primary keys for more than one table the auto-numbers continue to advanced from one table to the next. e.g., the first tables primary key will end at 35455 and the next tables primary key will start at 35456. Every tables auto-numbered primary key should start at 1.

henrykironde:
@ethanwhite any dataset to replicate this case

ethanwhite:
The Portal dataset is where I'd seen it before, but as you can see in #528 the whole dataset is actually erroring at the moment.

henrykironde:
validation and closing needed for this issue

henrykironde:
closing fixed in #546 8766cd9

Issue #331, opened 2015-07-27
Potential issues with skipping (primary key) fields


ethanwhite:
It is possible to skip fields when importing. In the text scripts this is done using `skip` as the type. In at least one case (PortalMammals; csv) this seems to have caused an issue where the field itself was not created, but the heading was (see #327, #330). This may also be related to #246, where auto-numbered primary key fields were not starting at the right number.

henrykironde:
I can't get the exact problem set scenario. 
Since Auto-numbered is a field we create, a case in which we skip it is a little abstract to me.

ethanwhite:
OK, let's go ahead and close this for now. There are also no current uses of `skip` in any of the scripts.

Issue #241, opened 2014-11-09
Cross-tab restructuring only works when adding primary key


ethanwhite:
Scripts for handling cross-tab data are only currently working if an auto-number primary key is added by the Retriever. To replicate:

1. Create a new script for handling cross-tab data
2. Don't including a `pk-auto` column

This will result in an index out of range error.

The source of the issue is probably related to Line 62 of `lib/engine.py`.

henrykironde:
validation and closing needed for this issue

henrykironde:
closing fixed in 	27f02f2

Issue #188, opened 2014-06-18
Allow indication of existing primary key in .json files w/o full column list


ethanwhite:
Many existing datasets have a primary key and it would be nice to be able to indicate this in a .json file without having to list out all of the columns and their types. The frictionless data schema calls for

```
"primaryKey": "a"
```

or 

```
"primaryKey": ["a", "c"]
```

So we should use that and process it properly.

ethanwhite:
I've updated this issue to reflect the new JSON script approach.

henrykironde:
I think we need to update this issue or close it.
Scripts can have more information added, but we are currently not processing this information. We shall need specific reasons or goals to process such information.

ethanwhite:
I'm fine with closing this.

PR #546, opened 2016-06-30
adding indexing to all engines

### LINKED ISSUE #528: Portal installation is erroring
```
ethan@stice:~$ retriever install csv Portal
=> Installing Portal Project Mammals (Ecological Archives 2009)
Downloading portal_plots.txt...
Downloading Portal_rodents_19772002.csv...
Downloading portal_species.txt...
WARNING (PortalMammals:plots): Exception in line 1: invalid literal for int() with base 10: 'd=document'
WARNING (PortalMammals:main): Exception in line 1: list index out of range
WARNING (PortalMammals:main): Exception in line 2: list index out of range
WARNING (PortalMammals:main): Exception in line 3: list index out of range
...
WARNING (PortalMammals:main): Exception in line 35547: list index out of range
WARNING (PortalMammals:main): Exception in line 35548: list index out of range
WARNING (PortalMammals:main): Exception in line 35549: list index out of range
Done!
```

Looks like maybe a recent change broken something?

henrykironde:
add indexing to the engines and resolve the issues of the current indexing for file database systems

There was no indexing for sqlite, postgres and mysql. Current indexing for csv json and xml, was not resetting the index for the tables. For each table created we reset `self.auto_column_number = 1`.
When a table needs to add an auto record_id column, use `create = true ` in `self.table.get_insert_columns(join=False, create=True)` when getting the insert columns for the `CREATE TABLE` statement. Use False when creating the insert statement since the auto record_id column is not needed. 

 set default encoding for test modules to avoid ASCII related issues
 change the test dump directory to pytest's temporary directory
fixes #528,#549,#246


henrykironde:
@ethanwhite, could you consider this PR since it is ready for merge instead of the indexing PR

engines/csvengine.py
```diff
@@ -39,7 +39,7 @@ class engine(Engine):
         "double": "REAL",
         "decimal": "REAL",
         "char": "TEXT",
-        "bool": "INTEGER",
+        "bool": "REAL",
```
ethanwhite: Seems like this should be `INTEGER` for the same reason as SQLite since in concept it should be 0 or 1. Is there a reason for making it `REAL`?

lib/engine.py
```diff
@@ -272,12 +268,12 @@ def convert_data_type(self, datatype):
...
         elif thistype[0:3] == "ct-":
             thistype = thistype[3:]
-
         if thistype in list(self.datatypes.keys()):
             thistype = self.datatypes[thistype]
-
             if isinstance(thistype, tuple):
-                if len(datatype) > 1:
+                if datatype[0] == 'pk-auto':
+                    thistype =thistype
```
ethanwhite: Add space after `=`

ethanwhite: What is the relationship between this conditional and the following conditional a few lines up:

```
         if thistype[0:3] == "pk-":
             thistype = thistype.lstrip("pk-")
             thispk = True
```

From a first read it seems like they're doing similar things maybe?

lib/engine.py
```diff
@@ -288,8 +284,10 @@ def convert_data_type(self, datatype):
             thistype = ""
 
         if thispk:
-            thistype = self.pkformat % thistype
-
+            if isinstance(thistype, tuple):
+                thistype = self.pkformat % thistype
+            else:
+                thistype = self.pkformat % (thistype, "")
```
ethanwhite: This if/else is basically a check on the length of `thistype` right? The first line works if it's 2 items, the second if it's only 1. We have a similar check above.

I guess in general I'm wondering if we really need three separate sets of if/else clauses that all set `thistype`. It seems like it would be clearer to have one big if/else clause, potentially with some compound conditions (i.e., conditionals that use `and` or `or`). That way it's clear that that chunk of code is being used to set `thistype` based on the conditions.

lib/engine.py
```diff
@@ -350,15 +348,13 @@ def create_table(self):
     def create_table_statement(self):
         """Returns a SQL statement to create a table."""
         create_stmt = "CREATE TABLE " + self.table_name() + " ("
-
-        columns = self.table.get_insert_columns(join=False)
-
+        # get insert coloumns including Auto field for creating table
```
ethanwhite: "coloumns" -> "columns"

henrykironde:
@ethanwhite, after reviewing the `if/else` statement and the data they are processing, I think it is better to leave them in that state. However, we could hold on merging and discus it later. 

we are trying to get the values of the data types first. Then if  we need to specify a primary key, there are two ways but they give us three versions

Sqlite auto increment with key in the middle : `INTEGER PRIMARY KEY AUTOINCREMENT `
Other auto increments with key at the end : `INT(5) NOT NULL AUTO_INCREMENT PRIMARY KEY `,
`Serial PRIMARY KEY`, 

And primary keys for non auto increment columns like `(pk-char, 20)`, `pk-int`

ethanwhite:
> we are trying to get the values of the data types first. Then if we need to specify a primary key, there are two ways but they give us three versions

Ah, OK, thanks. I think the refactoring you did makes this clearer. I got interrupted mid-review, but I'll try to finish it now.

henrykironde:
what I think, if we try to `set the primary key` as we` format the dataset type to match engine specific type`, we get more lines of code and  more complex when reading

lib/engine.py
```diff
@@ -264,32 +260,38 @@ def auto_get_delimiter(self, header):
...
+        if key[0:3] == "pk-":
+            key = key.lstrip("pk-")
             thispk = True
-        elif thistype[0:3] == "ct-":
-            thistype = thistype[3:]
-
-        if thistype in list(self.datatypes.keys()):
-            thistype = self.datatypes[thistype]
+        elif key[0:3] == "ct-":
+            key = key[3:]
```
ethanwhite: Looks like these 4 lines could be more safely and concisely written as

```
if key[0:3] == "pk-" or key[0:3] == "ct-":
    key = key[3:]
```

The original use of `lstrip` is actually dangerous because it's not stripping `pk-` it's stripping `p` and `k` and `-` so if any of this shows up on the left, not as part of `pk-` it will still get removed. So this change both makes the original `if` safer and the statement shorter and more clear.

lib/engine.py
```diff
@@ -264,32 +260,38 @@ def auto_get_delimiter(self, header):
...
 
+        # format the dataset type to match engine specific type
+        thistype = ""
```
ethanwhite: It seems like this would be better off in an `else` at the end for clarity. i.e., this is what ends up being the value if it isn't set in the `if.

lib/engine.py
ethanwhite: Oops, never mind. I was reading past `thispk = True`. Still probably best to change the `lstrip` to `key[3:]`.

lib/engine.py
ethanwhite: Nevermind, I see why you did it this way. This function is confusing to read (hence my reviewing mistakes), but I agree we should just leave it like it is at the moment since I can't figure out a better way to write it.

ethanwhite:
@henrykironde - yep, you're totally right. Just took me a while to understand that. Sorry for the noise.

henrykironde:
@ethanwhite, its totally fine, because it gives me more time to critically get things working well. hopefully we can resolve all the issues related to primary key and have things working well. 

this is helping me clarify #241,#331 and #188

#188 is more complicated than we initially thought, but working on it

lib/engine.py
```diff
@@ -348,17 +350,18 @@ def create_table(self):
...
 
     def create_table_statement(self):
-        """Returns a SQL statement to create a table."""
-        create_stmt = "CREATE TABLE " + self.table_name() + " ("
-
-        columns = self.table.get_insert_columns(join=False)
+        """Returns a SQL statement to create a table
 
+        use create=True to get insert columns including Auto field for creating table
+        default is false and used when returning only the colums for inserts
```
ethanwhite: It looks like these new lines in the docstring should be in `get_insert_columns` instead of `create_table_statement`.

henrykironde: we could add some description in the `get_insert_columns` too. but i think they would help someone debugging to clearly understand the end result of  using create =True in this part and why

lib/engine.py
```diff
@@ -541,22 +538,19 @@ def format_insert_value(self, value, datatype):
                 return "null"
         elif datatype in ("double", "decimal"):
             if strvalue:
-                return strvalue
+                decimals = float(strvalue)/1.0
```
ethanwhite: Are you sure we need the `/1.0`. I understand that we need `float` to convert `12` into `12.0`, but I don't see any difference in behavior between `float(strvalue)` and `float(strvalue)/1.0`.

lib/table.py
```diff
@@ -117,19 +116,17 @@ def combine_on_delimiter(self, line_as_list):
...
     def values_from_line(self, line):
         linevalues = []
-        if (self.pk and self.contains_pk is False):
-            column = 0
-        else:
-            column = -1
-
+        column = 0
+        if self.columns[column][1][0] == 'pk-auto':
+            column = 1
```
ethanwhite: It seems like this is clearer unless I'm missing something.

```python
if self.columns[column][1][0] == 'pk-auto':
    column = 1
else:
    column = 0
```

lib/table.py
```diff
@@ -171,20 +168,22 @@ def get_insert_columns(self, join=True):
             self.cleaned_columns = True
         for item in self.columns:
             thistype = item[1][0]
-            if ((thistype != "skip") and (thistype != "combine") and
-                    (self.contains_pk is True or thistype[0:3] != "pk-")):
+            if (thistype != "skip") and (thistype != "combine"):
                 columns += item[0] + ", "
         columns = columns.rstrip(', ')
+        if not create and self.columns[0][0] == 'record_id':
+            columns = columns.lstrip("record_id,").strip()
```
ethanwhite: This will remove any of the characters in `record_id,` not just `record_id,`.

Would `columns.replace("record_id,", "", 1)` be better? The `1` limits the number of replacements to the first value in case there happened to be another column named `record_id`.

lib/table.py
```diff
@@ -171,20 +168,22 @@ def get_insert_columns(self, join=True):
...
         if join:
             return columns
         else:
             return columns.lstrip("(").rstrip(")").split(", ")
 
     def get_column_datatypes(self):
         """Gets a set of column names for insert statements."""
-        columns = []
+        columns = []
```
ethanwhite: Remove extra whitespace.

test/test_integration.py
```diff
@@ -1,5 +1,7 @@
 """Integrations tests for EcoData Retriever"""
 from __future__ import print_function
+from imp import reload
```
ethanwhite: Unused import?

ethanwhite:
Looking great overall. I've made some minor suggestions in-line.

The one thing I don't understand at this point is why we're changing to using `latin-1` in the tests since we're currently using `utf8`. I understand we have some work to do there, but that seems like a separate PR. Is there a reason to make these changes to the tests here?

lib/engine.py
```diff
@@ -348,17 +350,18 @@ def create_table(self):
...
-        create_stmt = "CREATE TABLE " + self.table_name() + " ("
-
-        columns = self.table.get_insert_columns(join=False)
+        """Returns a SQL statement to create a table
 
+        use create=True to get insert columns including Auto field for creating table
+        default is false and used when returning only the colums for inserts 
+        """
+        create_stmt = "CREATE TABLE " + self.table_name() + " ("
+        columns = self.table.get_insert_columns(join=False, create=True)
```
ethanwhite: Is this this only use of `create=True`? If so, instead of putting the new code in `get_insert_columns`, which then requires explaining it here, why not just put the code here? E.g., something like

```
columns = self.table.get_insert_columns(join=False)
if self.columns[0][0] == 'record_id' and columns[0] == 'record_id':
    del columns[0]
```

I'm not sure if we need both checks, but this should, I think, replicate what the code is currently doing.

lib/table.py
henrykironde: very true

lib/engine.py
henrykironde: its not only used here. I will go ahead and change the doc string.

ethanwhite:
This looks good to go to me now.

@henrykironde - is this ready to merge from your perspective?

henrykironde:
this is also ready

PR #548, opened 2016-07-01
check if file is already downloaded

henrykironde:
When using the download engine and installing the data somewhere other than the working directory,
the retriever didn't check to see if the file was already there. This change checks the destination and
if the file is already there it does not download/move the file and reports that it already exists.

This also required changing the download engines regression tests so that they were still testing
the main behavior of the code.

References commit  #0912e5f7


henrykironde:
@ethanwhite, I dont think we need new test cases for these changes.

ethanwhite:
Totally agree. It's weird that codecov fails with such a tiny change in coverage.

henrykironde:
And about the regression tests, I think we have this behavior tested. Since we ship with Delmoral and Avianbodysize and only download Mom2003.  So I guess we clear the previous PR and get this one in after

`os.system("retriever download {0} -p raw_data/{0}".format(dataset)) getmd5(data="raw_data/{0}".format(dataset), data_type='dir', mode="rU")`

ethanwhite:
This looks ready to go from a quick read. I didn't quite understand your last comment. Are you saying we should get #546 merged first and then merge this one?

henrykironde:
we can merge this

PR #550, opened 2016-07-04
Get try_install_all running and exclude more datasets

ethanwhite:
This updates try_install_all to get it running again. There was a very
strange error trigger by the `if value` check for the download only engine.
For some reason the value for the download only engine doesn't evaluate to
True or False when using `if value`.

Also excludes some of the new large datasets.

### DIFF
--- a/try_install_all.py
+++ b/try_install_all.py
@@ -24,7 +24,7 @@ if len(sys.argv) > 1:
     ]
 SCRIPT_LIST = SCRIPT_LIST()
 TEST_ENGINES = {}
-IGNORE = ["AvianBodyMass", "FIA"]
+IGNORE = ["AvianBodyMass", "FIA", "Bioclim", "PRISM", "vertnet"]
 
 for engine in ENGINE_LIST:
     opts = {}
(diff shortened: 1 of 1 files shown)


henrykironde:
@ethanwhite still looking at this, Its not your password issue. give me an hour will get back to you on this

henrykironde:
there  is an invisible kind of bug when seting `opts`, So mysql calls `get_input` later https://github.com/weecology/retriever/blob/master/engines/mysql.py#L105
and postgres on the other hand calls `get_input ()`  first https://github.com/weecology/retriever/blob/master/engines/postgres.py#L128.  On command line, this works because we are setting the opt before all the get connection or get_cursor functions are called. with the try_install_all.py  we can go ahead and change the `get_connection` function in MySql or add a line    
``` 
try:
        TEST_ENGINES[engine.abbreviation] = choose_engine(opts)
        TEST_ENGINES[engine.abbreviation].get_input()
        TEST_ENGINES[engine.abbreviation].get_cursor()
    except:
```

ethanwhite:
Thanks for the PR for the fix @henrykironde. I've merged it and it fixed the MySQL issue locally. Is there anything else for me to do for this PR?

henrykironde:
@ethanwhite, nothing needed we can merge

ethanwhite:
Great. Go ahead and push the big green button.

PR #552, opened 2016-07-05
Repair encoding

henrykironde:
I am using this as a test for latin encoding to see how we can change


henrykironde:
@goelakash  and @ethanwhite as a reference to 2dbc9ef
Let is discus the the encoding issues because this is giving me the right results.

I am wondering why we changed back to the UTF8, I may be missing something

ethanwhite:
> I am wondering why we changed back to the UTF8 encoding, I may be missing something

Here's some of the earlier discussion:
https://github.com/weecology/retriever/pull/533#issuecomment-226974683

henrykironde:
I remember now, @akshayah3, had opted for utf8 but I guess I didn't get to test the PR. Earlier I was still concerned with the some of those characters. If we want to switch to utf8 I have no problem, we just have to get the encoding and decoding right .

an example of integration  test that could ensure this would be 

```
non_ascii_csv = {'name': 'non_ascii_csv',
              'raw_data': "a,b,c\nBetulaceae,sinuata (Regel) A,Löve & D. Löve\néléphant,fièvre,öve"
              'script': "shortname: non_ascii_csv\ntable: non_ascii_csv, http://example.com/non_ascii_csv.txt\n*column: record_id, pk-auto\n*column: a, char\n*column: b, char, 20"
              'expect_out': '"record_id","a","b","c"\n1,Betulaceae,sinuata (Regel) A,Löve & D. Löve\n2,éléphant,fièvre,öve'}
```

goelakash:
@ethanwhite @henrykironde The main issue is that many files are encoded in `latin-1`, but Python 3 and other files in general are encoded in `utf-8` by default. We could follow any standard, but since `latin-1` avoids data loss, it should be preferred.

ethanwhite:
@goelakash I had been under the (apparently mistaken) impression that `utf-8` was basically a superset of things like `latin-1`. Is that not the case?

goelakash:
@ethanwhite [This](http://stackoverflow.com/questions/7163485/why-doesnt-this-conversion-to-utf8-work) explains it.

goelakash:
I quote one answer that sums it up nicely with an example
```
UTF-8 is not a subset of Latin-1. UTF-8 encodes ASCII with the same single bytes. For all other code points, it's all multiple bytes.

Put simply, \xf1 is not valid UTF-8, as Python tells you. "Unexpected end of input" indicates that this byte marks the beginning of a multi-byte sequence which is not provided.
```

ethanwhite:
@goelakash: OK, thanks. I was making that exact mistake of confusing unicode and `utf-8`. Since this PR has a bunch of other stuff mixed in with it, would you mind opening a new one that just makes the `utf-8` to `latin-1` changes in all the appropriate locations? Thanks.

goelakash:
@ethanwhite No problem.

henrykironde:
@ethanwhite, @goelakash if code fails on appveyor python3 environment, do not be frustrated. For some reason appveyor fails to install psycopg2.
I have created an issue #https://github.com/appveyor/ci/issues/909 on their page, hopefully they get it resolved soon.

goelakash:
@henrykironde psycopg2 will fail on Python 3.5, but not older versions.

henrykironde:
@goelakash, they updated something somewhere. http://www.appveyor.com/updates/2016/06/28
They suggest using older images `os: Previous Visual Studio 2015` but I have not read through yet. 
We may also think of other ways to get psycopg2 installed.

May be change the version we are using in the requirements part
http://www.stickpeople.com/projects/python/win-psycopg/
or install the right version for appveyor directly/manually using the appveyor.yaml file
`pip install -e path_to_source`
and specify the one we want.

Some known sources for binaries
http://www.lfd.uci.edu/~gohlke/pythonlibs/

or install from source.

Last, since we are not testing Postgres on windows yet, we can find a way to ignore that specific error

henrykironde:
This was a test PR

PR #553, opened 2016-07-08
latin-1 encoding changes for python 2/3 compatibility

goelakash:
As latin-1 is the preferred encoding scheme for the datasets, this
commit changes the default encoding used by Engine base class to
open files in text-mode using latin-1 encoding.

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -17,7 +17,7 @@ import sys
 reload(sys)
 if hasattr(sys, 'setdefaultencoding'):
     # set default encoding to utf-8 to decode source text
-    sys.setdefaultencoding('utf-8')
+    sys.setdefaultencoding('latin-1')
 from retriever import VERSION, MASTER, SCRIPT_LIST, sample_script, current_platform
 from retriever.engines import engine_list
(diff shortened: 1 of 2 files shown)


henrykironde:
Thanks @goelakash could you also include and integration test. 


```
non_ascii_csv = {'name': 'non_ascii_csv',
              'raw_data': "a,b,c\nBetulaceae,sinuata (Regel) A,Löve & D. Löve\néléphant,fièvre,hfièvre"
              'script': "shortname: non_ascii_csv\ntable: non_ascii_csv, http://example.com/non_ascii_csv.txt\n*column: record_id, pk-auto\n*column: a, char\n*column: b, char, 20"
              'expect_out': '"record_id","a","b","c"\n1,Betulaceae,sinuata (Regel) A,Löve & D. Löve\n2,éléphant,fièvre,hfièvre'}
```
I tried but couldn't pass on python3

ethanwhite:
This is now also set in a couple of the `test` files, so we should go ahead and update those at the same time.

PR #554, opened 2016-07-08
Use older worker image (VS 2015) for appveyor

goelakash:
Due to updation of appveyor worker environment, the psycopg2 build fails.
Thus, the appveyor.yml has been updated to use a previous worker image
from 2015.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,4 @@
+os: Previous Visual Studio 2015
 environment:
 
   matrix:


goelakash:
@henrykironde @ethanwhite If this passes, then it should be merged before my other PR for `latin-1` encoding.

goelakash:
@ethanwhite I suppose this should be closed as the other #553 passed without errors?

henrykironde:
since appveyor has started working, we do not need to use an older image

PR #559, opened 2016-07-11
remove the default mandatory auto key field.

henrykironde:
currently all the file based engines create an auto increment field
if the table columns are not defined.

When fields are not defined, the auto pk is None and the previous
version was by default setting auto field as the first colum value.

This check as been removed and put in auto_get_datatypes and will be part of
adding the capability of only defining the primary key in the scripts
ref:#188

We also remove the escaping of characters in the file systems.

### DIFF
--- a/engines/csvengine.py
+++ b/engines/csvengine.py
@@ -56,7 +56,7 @@ class engine(Engine):
         self.auto_column_number = 1
         self.output_file = open(self.table_name(), "w")
         self.output_file.write(
-            ','.join(['"%s"' % c[0] for c in self.table.columns]))
+            ','.join(['"%s"' % c for c in self.table.get_insert_columns(join=False,create=True)]))
 
     def disconnect(self):
         """Close the last file in the dataset"""
@@ -74,7 +74,7 @@ class engine(Engine):
 
     def format_insert_value(self, value, datatype):
         """Formats a value for an insert statement"""
-        v = Engine.format_insert_value(self, value, datatype)
+        v = Engine.format_insert_value(self, value, datatype, escape=False)
         if v == 'null':
             return ""
         try:
@@ -86,17 +86,14 @@ class engine(Engine):
 
     def insert_statement(self, values):
         """Returns a comma delimited row of values"""
+
         if not hasattr(self, 'auto_column_number'):
             self.auto_column_number = 1
-        offset = 0
-        for i in range(len(self.table.columns)):
-            column = self.table.columns[i]
-            if 'auto' in column[1][0]:
-                values = values[:i + offset] + \
-                    [self.auto_column_number] + values[i + offset:]
-                self.auto_column_number += 1
-                offset += 1
-        return ','.join([str(value) for value in values])
+        insert_stmt = ','.join([str(value) for value in values])
+        if self.table.columns[0][1][0][3:] == 'auto':
+            insert_stmt = str(self.auto_column_number) + "," + insert_stmt
+            self.auto_column_number += 1
+        return insert_stmt
 
     def table_exists(self, dbname, tablename):
         """Check to see if the data file currently exists"""
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -171,12 +171,6 @@ class Engine(object):
         if not self.table.columns:
             lines = gen_from_source(source)
 
-            if pk is None:
-                self.table.columns = [("record_id", ("pk-auto",))]
-            else:
-                self.table.columns = []
-                self.table.contains_pk = True
-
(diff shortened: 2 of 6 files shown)


engines/csvengine.py
```diff
@@ -86,17 +86,16 @@ def format_insert_value(self, value, datatype):
...
-                values = values[:i + offset] + \
-                    [self.auto_column_number] + values[i + offset:]
-                self.auto_column_number += 1
-                offset += 1
-        return ','.join([str(value) for value in values])
+
+        insert_stmt = ','.join([str(value) for value in values])
+
+        if self.table.columns[0][1][0][3:] == 'auto':
+            insert_stmt = str(self.auto_column_number)+"," + insert_stmt
```
ethanwhite: add spaces around first `+`

engines/csvengine.py
```diff
@@ -86,17 +86,16 @@ def format_insert_value(self, value, datatype):
...
-            column = self.table.columns[i]
-            if 'auto' in column[1][0]:
-                values = values[:i + offset] + \
-                    [self.auto_column_number] + values[i + offset:]
-                self.auto_column_number += 1
-                offset += 1
-        return ','.join([str(value) for value in values])
+
+        insert_stmt = ','.join([str(value) for value in values])
+
```
ethanwhite: I'd probably remove this blank line to make it clearer that this is directly related to the following `if` statement.

lib/engine.py
```diff
@@ -510,7 +509,7 @@ def format_filename(self, filename):
         """Returns the full path of a file in the archive directory."""
         return os.path.join(self.format_data_dir(), filename)
 
-    def format_insert_value(self, value, datatype):
+    def format_insert_value(self, value, datatype, escape=True):
         """Formats a value for an insert statement, for example by surrounding
         it in single quotes."""
```
ethanwhite: We should document what `escape` is doing in the docstring.

lib/engine.py
```diff
@@ -541,11 +540,12 @@ def format_insert_value(self, value, datatype):
...
-            if hasattr(self.table, "escape_double_quotes") and self.table.escape_double_quotes:
-                strvalue = self.escape_double_quotes(strvalue)
-            if hasattr(self.table, "escape_single_quotes") and self.table.escape_single_quotes:
-                strvalue = self.escape_single_quotes(strvalue)
+            if escape:
+                # automatically escape quotes in string fields
+                if hasattr(self.table, "escape_double_quotes") and self.table.escape_double_quotes:
+                    strvalue = self.escape_double_quotes(strvalue)
+                if hasattr(self.table, "escape_single_quotes") and self.table.escape_single_quotes:
+                    strvalue = self.escape_single_quotes(strvalue)
```
ethanwhite: It's not clear to me from the code why we would want to escape sometimes and not others. Can you add some description of this to the docstring or in the inline comment on 544.

ethanwhite:
Looks great. I

ethanwhite:
Looks great. I suggested one very minor cleanup and suggested expanding the documentation a little and then this looks ready to go.

lib/engine.py
henrykironde: @ethanwhite, after we removed the issue of creating auto fields for all tables when using file database engines, the file based engines where escaping some strings. This made the md5sums different between the file based engines (xml, json, csv) and the non file based engines (postgres,mysql, sqlite).

So the file based engines inherit the Engine's `format_insert_value()` function but set the flag to false so that no escaping is done for them.

henrykironde:
Thanks for the reviews, let me know if the docstring is clear

lib/engine.py
ethanwhite: Ah, very interesting. Really nice solution.

After reading through the docstring and the function I think a broader rewording is really needed to capture what's going on with this function anyway. I gave it a go and came up with:

```
        """Format a value for an insert statement based on data type

        Different data types need to be formated differently to be properly
        stored in database management systems. The correct formats are
        obtained by:

        1. Removing extra enclosing quotes
        2. Harmonizing null indicators
        3. Cleaning up badly formatted integers
        4. Obtaining consistent float representations of decimals

        The optional `escape` argument controls whether additional quotes in
        strings are escaped, as needed for SQL database management systems
        (escape=True), or not escaped, as needed for flat file based engines
        (escape=False). """
```

ethanwhite:
See what you think about the substantially expanded version of the docstring of suggested in-line.h

lib/engine.py
henrykironde: I think this is the best docstring we have :)

PR #561, opened 2016-07-13
correct auto key for cross-tab

henrykironde:
Cross-tab restructuring only worked when adding primary key.
This PR fixes#241 and includes integration test for cross tab
with and without auto primary keys.

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 import imp
 import os
 import shutil
+import pytest
 
 from retriever.lib.compile import compile_script
 from retriever import HOME_DIR, ENGINE_LIST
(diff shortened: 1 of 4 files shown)


ethanwhite:
@goelakash would you be interested in reviewing this PR to get some experience on the other side of the process?

goelakash:
Sure, let me see.
On Jul 13, 2016 9:01 PM, "Ethan White" <notifications@github.com> wrote:
>
> @goelakash would you be interested in reviewing this PR to get some
experience on the other side of the process?

test/test_regression_cli.py
```diff
@@ -46,7 +46,6 @@ def get_csv_md5(dataset, engines, tmpdir):
 
 def setup_module():
     """Update retriever scripts and cd to test directory to find data"""
-    os.chdir("./test/")
```
goelakash: @henrykironde Why is this removed?

henrykironde: for some reason, it was causing some errors, I will go ahead and test it once again and let you know.

ethanwhite: The original reason for this line was to allow us to include the raw data for two of the regression test cases instead of downloading it. The retriever's default search looks in a couple of places including `./raw_data` to see if the data exists before downloading it, so this is moving the working directory into `test` so that it sees `./test/raw_data`. I'm not sure this is the best design, but that's why it was there.

henrykironde: Yes that is true , but for some reason, it seemed like the tests where in that directory`./test/` before. regression tests are run.
When we try to get in to that directory, it will error. I will go through the test and confirm.

test/test_integration.py
```diff
@@ -39,7 +54,18 @@ def teardown_module():
     for test in tests:
         shutil.rmtree(os.path.join(HOME_DIR, "raw_data", test['name']))
         os.remove(os.path.join(HOME_DIR, "scripts", test['name'] + '.script'))
-        os.remove(test['name']+'_'+test['name']+'.txt')
+        os.system("rm -r *{}".format(test['name']))
```
henrykironde: remove the file using *

test/test_regression_cli.py
```diff
@@ -46,12 +46,13 @@ def get_csv_md5(dataset, engines, tmpdir):
...
+    os.chdir(os.path.dirname(os.path.abspath(__file__)))
```
henrykironde: added `os.chdir("./test/")`
but using a full path got from the file location
Note: we should avoid reference paths when ever possible. Sometimes the program just gets losts if one test errors

test/test_regression_cli.py
```diff
@@ -46,12 +46,13 @@ def get_csv_md5(dataset, engines, tmpdir):
...
     os.system("retriever update")
 
 
 def teardown_module():
     """Cleanup temporary output files after testing and return to root directory"""
+    os.chdir(os.path.dirname(os.path.abspath(__file__)))
```
henrykironde: make sure you are in the test directory and then remove all the files not needed.
Note: according to the test design `raw_data/AvianBodySize`  and `test/raw_data/DelMoral2010` should not be deleted

test/test_retriever.py
```diff
@@ -196,7 +192,7 @@ def test_json2csv():
     """Test json2csv function
     creates a json file and tests the md5 sum calculation"""
     json_file = create_file("""[ {"User": "Alex", "Country": "US", "Age": "25"} ]""", 'output.json')
-    output_json = json2csv(json_file, "test/output_json.csv", header_values=["User", "Country", "Age"])
```
henrykironde: we do not need the `test/`  for all modules we setup to the test directory

henrykironde:
I am going to over write this PR. 
Though I am not sure why we did not merge it.

If there are some more changes needed, let me know or we could have the same comments in the next PR.  
:)

test/test_integration.py
```diff
@@ -51,17 +77,7 @@ def get_script_module(script_name):
 csv_engine.opts = {'engine': 'csv', 'table_name': './{db}_{table}.txt'}
```
henrykironde: ` csv_engine.opts = {'engine': 'csv', 'table_name': './{db}_{table}.txt'}`

line will be removed

ethanwhite:
So is this PR superseded by #569 now? If so, go ahead and close. If not, let me know and I'll finish the review.

henrykironde:
closing, changes in  #569

Issue #562, opened 2016-07-14
Add "dataset out of service" functionality


ethanwhite:
## Current situation

Currently if a dataset that we have a script for stops working because of some change on the data providers end (e.g., #513) we have two choices. Keep the script, in which case the user tries to install the dataset and gets an error, or remove the script, in which case code that used to run the retriever for that dataset starts failing in a confusing way by saying "that dataset doesn't exist". Neither of these communicates to the user what has actually happened. The later also makes it harder to go back and fix the broken script later.
## Proposed solution

Add "dataset out of service" functionality. We would add an indicator of some kind to the script that triggers it to not try to install but instead report back to the user something like "This dataset is currently not available. Either the data is no longer openly available or the way in which it is provided has changed so that the Retriever doesn't know how to work with it".

We may also want to have the retriever return an error code (1 instead of 0) if this happens, but I'm not sure about the general thought on that for this sort of situation.

ethanwhite:
As part of adding this we should apply it to NPN and eBird, both of which are broken at the moment due to these kinds of changes.

henrykironde:
"This dataset is currently not available. Either the data is no longer openly available or the way in which it is provided has changed so that the Retriever doesn't know how to work with it". is perfect.

ethanwhite:
Glad you liked it. In additional to a generic message, it might be useful to allow a custom message for a dataset. Then we could do things like.

"The eBird dataset is now longer available because it has been removed from DataOne"

or

"The AvianBodyMass dataset is only available via a CD that comes with the CRC Handbook of Avian Body Mass. If you wish to use this data you need to: 1. Obtain the CD; 2. Put the datafiles in `.retriever/raw_data/AvianBodyMass/`; and 3. Change line XX of the avianbodymass.py script to `unavailable = False`.

ethanwhite:
As part of this change we should also remove the listing of these datasets from `retriever ls`.

jainamritanshu:
I would like to work on this if its still open :smile:

henrykironde:
Go for it

ethanwhite:
I think the way we're failing when a dataset doesn't exist is sufficient for the time being.

Issue #551, opened 2016-07-05
Current dataset errors


ethanwhite:
Running `try_install_all.py` on my home machine currently lists these failures (MySQL didn't run for some reason):

```
Engine, Dataset, Error
('xml', 'EA_Rodriguez-Buritica2013', IndexError('list index out of range',))
('sqlite', 'EA_Rodriguez-Buritica2013', IndexError('list index out of range',))
('postgres', 'EA_Rodriguez-Buritica2013', IndexError('list index out of range',))
('json', 'EA_Rodriguez-Buritica2013', IndexError('list index out of range',))
('csv', 'EA_Rodriguez-Buritica2013', IndexError('list index out of range',))
('xml', 'npn', HTTPError())
('sqlite', 'npn', HTTPError())
('postgres', 'npn', HTTPError())
('json', 'npn', HTTPError())
('download', 'npn', HTTPError())
('csv', 'npn', HTTPError())
('xml', 'EA_Colin2014', IndexError('list index out of range',))
('xml', 'EA_Fray_Jorge2013', IndexError('list index out of range',))
('sqlite', 'EA_Fray_Jorge2013', OperationalError('no such column: ND',))
('postgres', 'EA_Fray_Jorge2013', ProgrammingError('column "nd" does not exist\nLINE 1: ...pling_date, cover) VALUES (\'ADBE\', \'+D+P\', 3, \'Jun-97\', ND);\n                                                                   ^\n',))
('postgres', 'bbs50stop', ProgrammingError('relation "bbs50.species" does not exist\nLINE 1: INSERT INTO BBS50.species (species_id, aou, english_common_n...\n                    ^\n',))
('xml', 'Dryad_tamburello_etal_2013', IndexError('list index out of range',))
('postgres', 'Dryad_tamburello_etal_2013', ProgrammingError('relation "homeranges.ranges" does not exist\nLINE 1: INSERT INTO HomeRanges.ranges (taxon, common_name, class, or...\n                    ^\n',))
('xml', 'MammalSuperTree', Exception("This dataset contains only non-tabular data files, and can only be used with the 'download only' engine.\nTry 'retriever download datasetname instead.",))
('sqlite', 'MammalSuperTree', Exception("This dataset contains only non-tabular data files, and can only be used with the 'download only' engine.\nTry 'retriever download datasetname instead.",))
[...]

PR #563, opened 2016-07-14
add downloadonly datasets to IGNORE

henrykironde:
change the IF statmenets to make sure that only non IGNORE
dataset are logged

### DIFF
--- a/try_install_all.py
+++ b/try_install_all.py
@@ -24,7 +24,8 @@ if len(sys.argv) > 1:
     ]
 SCRIPT_LIST = SCRIPT_LIST()
 TEST_ENGINES = {}
-IGNORE = ["AvianBodyMass", "FIA", "Bioclim", "PRISM", "vertnet"]
+IGNORE = ["AvianBodyMass", "FIA", "Bioclim", "PRISM", "vertnet","NPN", "mammsupertree", "eBirdOD"]
+IGNORE = [dataset.lower() for dataset in IGNORE]
 
 for engine in ENGINE_LIST:
     opts = {}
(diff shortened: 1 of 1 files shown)


henrykironde:
@ethanwhite  and @goelakash  do not review

henrykironde:
Ready for review.

I changed the IGNORE to lowercase and the dataset names from the script to lower

```
IGNORE = ["AvianBodyMass", "FIA", "Bioclim", "PRISM", "vertnet","NPN", "mammsupertree", "eBirdOD"]
IGNORE = [dataset.lower() for dataset in IGNORE]
```
We were not matching some of the dataset that are in the ignore because of difference in Caps and lowers.

`if module.SCRIPT.shortname.lower() not in IGNORE:`

this resolves just a small part of the errors from #551

goelakash:
@henrykironde Why are we ignoring `AvianBodyMass` ?

henrykironde:
let me find out.

henrykironde:
This is the data base that need a CD
```
"""Retriever script for CRC Handbook of Avian Body Masses companion CD

NOTE: This data is not publicly available. To download the data, you'll need
the CRC Avian Body Masses CD. Create a new directory at
raw_data/AvianBodyMass and copy the contents of the CD there before running
this script.

"""
```

goelakash:
CD, as in Compact Disc? :stuck_out_tongue:

henrykironde:
:) yah you load it and download the data directory to the raw_data/AvianBodyMass
hehe

goelakash:
Wait, are you serious?

henrykironde:
hahaha Yes the [avianbodymass scripts ](https://github.com/weecology/retriever/blob/master/scripts/CRC_avianbodymass.py) gets the data from the raw_data folder. And I don't have a copy here but  they sell the data with the book
https://www.crcpress.com/CRC-Handbook-of-Avian-Body-Masses-Second-Edition/Dunning-Jr/p/book/9781420064445

goelakash:
So that means that this dataset is unavailable now?

henrykironde:
Yes, unavailable for testing. Users who want to use it can buy the CD and call the retriever to install it into which ever engine they want.

ethanwhite:
I think we should go ahead flag AvianBodyMass as "out of service" as part of #562.

try_install_all.py
```diff
@@ -24,7 +24,8 @@
     ]
 SCRIPT_LIST = SCRIPT_LIST()
 TEST_ENGINES = {}
-IGNORE = ["AvianBodyMass", "FIA", "Bioclim", "PRISM", "vertnet"]
+IGNORE = ["AvianBodyMass", "FIA", "Bioclim", "PRISM", "vertnet","NPN", "mammsupertree", "eBirdOD"]
```
ethanwhite: I don't think we should ignore NPN or eBirdOD since these are legitimately erroring at the moment. We should fix those errors via #562 instead.

henrykironde: The eBirdOD folks decided to tighten up on how people get their data and they moved it to another site. But we have vertnet which has ebird data.(for now)

In the case of NPN, we shall have to go through their sites, I guess a lot has changed on how to get data. They must have improved their service. I took a quick look and we need to discus what exactly we want to fetch from their repo. 


For this PR, I think ebird and NPN will take some time to get back to working state.

As we try to finish up the work on improving tests, It would be good if we can ignore those for now and solve the rest. The main goal is to have a working test on all scripts ASAP. 
And we shall set travis to run the test once major changes happen and periodically.

ethanwhite: I agree that we shouldn't try to get eBird and and NPN working immediately. What I'm suggesting is preventing them from erroring and providing a useful message instead. My concern with ignoring them here is that they actually are failing and we want the testing to fail as well until they aren't throwing errors. Once tests get turned off they tend to get forgotten and not turned back on because they aren't throwing errors. Happy to chat more about this in person tomorrow so that we can figure out what's best.

ethanwhite: Here's a quick fix for this for now: #565.

PR #564, opened 2016-07-15
Add latin-1 encoding to try_install_all

ethanwhite:
Since try_install_all.py doesn't directly call the CLI the default
encoding isn't set.

This is actually the code I used when updating #551.


henrykironde:
This is ready to be merged

PR #565, opened 2016-07-15
Provide informative failures for eBird and NPN

### LINKED ISSUE #513: Obsolete links for some tables
While testing [this issue](https://github.com/weecology/retriever/issues/290), I used the following command:
```
retriever install postgres eBirdOD
```

However, it seems like the url for this table is no longer valid:
https://dataone.ornith.cornell.edu/metacat/d1/mn/v1/object/CLOEODDATA.05192014.1

and gives me the following  message:
```
Downloading /Users/ddigges/.retriever/raw_data/eBirdOD/CLOEODDATA.05192014.1...from: https://dataone.ornith.cornell.edu/metacat/d1/mn/v1/object/CLOEODDATA.05192014.1
[Errno socket error] [Errno 8] nodename nor servname provided, or not known
```

We should look into any tables that have obsolete links and remove them.

ethanwhite:
eBird has removed it's data from DataOne and so no longer works.

NPN updated to a new API which requires an extensive reworking of the script.

These changes print useful information about why the datasets won't install,
instead of erroring without useful information.

This will also prevent them from triggering error reporting in try_install_all.py

Fixes #513.

### DIFF
--- a/scripts/eBird_observation.py
+++ b/scripts/eBird_observation.py
@@ -1,6 +1,8 @@
 #retriever
 """EcoData Retriever script for the eBird Observation Dataset"""
 
+import sys
+
 from retriever.lib.templates import Script
 from retriever.lib.models import Table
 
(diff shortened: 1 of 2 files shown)


ethanwhite:
Actually, this may not do what we want in `try_install_all.py`. I tested it in isolation, but I think in a full run of `try_install_all.py` it will cause the run to stop once it hits one of these datasets. Closing.

henrykironde:
@ethanwhite, how about we create an issue that points to these datasets and also to the `try_install_all.py`, 
My idea is to set a flat in the script and since the addendum carries that type of info
when we are running the scripts, we check if that flat is set

and we buy pass the download

```
             try:
                if not flag:
                    module.SCRIPT.download(value)
            except KeyboardInterrupt:
                pass
            except Exception as e:
```
when we logging the last error we check if that flag is set and log the message from the addendum

ethanwhite:
Sounds good. Open the issue and then ping me and I'll go ahead and get #563 merged. Then you can get this all cleaned up properly when you have time.

PR #568, opened 2016-07-16
Add row value length checks before adding to table

### LINKED ISSUE #567: Remove extra newline characters from dataset before writing to tables
A few datasets, like `SonoranPerennials` have extra `\n` in one of their raw files (Line 67 in `Species.csv` for this one as an example), which results in the xml errrors from #551 .  Need to process these before adding to tables in `engine.py`.

goelakash:
Some datasets have an extra newline character in their rows, which causes
retriever to read them as separate instances on data, This causes mismatch
int the numbers if columns and data values, producing IndexError exceptions
in a few cases.

This commit adds a loop to keep adding lines to a row until the number of values
equals the number of columns, before adding the row to the table.

Closes #567 and reduces the severity of #551

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -95,11 +95,26 @@ class Engine(object):
             real_line_length = sum(1 for _ in len_source)
 
         total = self.table.record_id + real_line_length
+        pos = 0
         for line in real_lines:
             if not self.table.fixed_width:
                 line = line.strip()
             if line:
                 self.table.record_id += 1
+
+                # Check for single row distributed over multiple lines
+                val_list = self.table.split_on_delimiter(line)
+                while len(val_list) < len(self.table.get_column_datatypes()):
+                    line = line.rstrip('\n')
+                    if type(real_lines) != 'list':
+                        line += next(real_lines)
+                    else:
+                        line += real_lines[pos+1]
+                        real_lines.pop(pos+1)
+                    val_list = (self.table.split_on_delimiter(line))
+                pos += 1
+                assert(len(val_list) == len(self.table.get_column_datatypes()))
+
                 linevalues = self.table.values_from_line(line)
 
                 types = self.table.get_column_datatypes()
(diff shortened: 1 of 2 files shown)


henrykironde:
could you also come up with a sample test scenario for this.  So that we can see clearly what we are dealing with. You could create a simple sample data file and share it here.

henrykironde:
something you may consider when solving this is that when we have a primary key that is auto or a column with skip, we may be getting varying lengths of the insert statements. Just cross check.
This may not be affecting anything at that point in the code though.

goelakash:
@henrykironde Do you mean an integration test? How can I add a test for a class method, `Engine` in this case?
This code will basically read this:
```
col1, col2, col3
1, 2
, 3
```
as 
```
col1, col2, col3
1, 2, 3
```

henrykironde:
@goelakash, This is really a good sample. But the case is weird. We have good integration test ready for merge soon. We can add the sample later after these two merges.

henrykironde:
give me 40 minutes, will show you

henrykironde:
@goelakash, this is kind of tricky.
So if we are getting this in one line , which seems to be the case, we do not need an integration test.

However for the data you have provided which is line by line,
Given that, our data set with errors looks like this 

```
col1, col2, col3
1, 2
, 3
```

we could have something like 

```
col1, col2, col3
1, 2,
, 3,
```

To add integration sample, we can add the dataset in the [integration test](https://github.com/weecology/retriever/blob/master/test/test_integration.py) .

For example:
```
given_name = {'name': 'given_name',
              'raw_data': "col1,col2,col3\n1,2\n\n,3",
              'script': "shortname: given_name\ntable: simple_csv, http://example.com/given_name.txt",
              'expect_out': 'col1,col2,col3\n1,2,\n,3,\n'}
```

To add test for class methods of Engine, use [test_retriever.py] (https://github.com/weecology/retriever/blob/master/test/test_retriever.py)

henrykironde:
do not forget to squash the commits since they are doing the same thing 👍

goelakash:
@henrykironde Added tests and squashed commits.

henrykironde:
Thanks @goelakash,this is fine, the case is just rare :)

PR #576, opened 2016-07-18
add skip test

### LINKED ISSUE #574: skip column in dataset fails
when a dataset has the `skip` datatype, There are some failures in the xml and json engines due to the removed column.

MySql also fails to skip the data correctly 
**example dataset:** 

```
skip_csv = {'name': 'skip_csv',
              'raw_data': "a,b,c\n1,2,3\n4,5,6\n",
              'script': "shortname: skip_csv\ntable: skip_csv, http://example.com/skip_csv.txt\n*column: a, skip\n*column: b, int\n*column: c, int",
```
mysql result:

```
assert 'b,c\n1,2\n4,5\n' == 'b,c\n2,3\n5,6\n'
```

henrykironde:
when skip is set in the dataset, the column should not be included
In the engine assert check was previously checking raw lines.
This PR does not make the an assertion check on values.
fixes #574

### DIFF
--- a/engines/xmlengine.py
+++ b/engines/xmlengine.py
@@ -124,7 +124,7 @@ class engine(Engine):
 
     def to_csv(self):
         """Export table from xml engine to CSV file"""
-        keys = [columnname[0] for columnname in self.table.columns]
+        keys = self.table.get_insert_columns(join=False, create=True)
         csv_outfile = xml2csv(self.table_name(), header_values=keys)
         return sort_csv(csv_outfile)
 
--- a/engines/jsonengine.py
+++ b/engines/jsonengine.py
@@ -121,7 +121,7 @@ class engine(Engine):
 
     def to_csv(self):
         """Export table from json engine to CSV file"""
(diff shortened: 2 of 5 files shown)


engines/xmlengine.py
```diff
@@ -124,7 +124,7 @@ def table_exists(self, dbname, tablename):
 
     def to_csv(self):
         """Export table from xml engine to CSV file"""
-        keys = [columnname[0] for columnname in self.table.columns]
+        keys = [columnname for columnname in self.table.get_insert_columns(join=False, create=True)]
```
ethanwhite: Is the list comprehension doing anything now? It seems like `get_insert_columns(join=False)` will return a list and the list comprehension will just take the items from that list and use them to make a new list.

ethanwhite:
Looks good to me. One possible opportunity for simplification mentioned in-line.

engines/xmlengine.py
henrykironde: Very true

PR #579, opened 2016-07-19
Fix multi-line row check

ethanwhite:
The type check was against `'list'` instead of `list`.

Also adds back the test for this functionality since it appears to work fine after some minor cleanup.


henrykironde:
Coding is fun, I had just started to wonder how new errors come up. 
But thats the fun part of it all. We missed that.

 @ethanwhite, Thanks

PR #581, opened 2016-07-19
Use `extract_values` when checking for mid-row line endings

### LINKED ISSUE #580: Checking for row over multiple lines breaks fixed-width tables
The `val_list = self.table.split_on_delimiter(line)` line added in 9f05ea314d36cce2 runs `split_on_delimiter` without checking if the table has a delimiter. When the table is fixed width this results in an error:

```
ethan@stice:~$ retriever install sqlite bbs
=> Installing USGS North American Breeding Bird Survey
Creating table BBS_species...
"delimiter" must be an 1-character string
Done!
```

ethanwhite:
When handling extra mid-row line endings we split the line to check if there
are enough values. This was added in 9f05ea3, but it ran `split_on_delimiter`
without checking if the table has a delimiter and so errored on fixed-width
tables. This replaces `split_on_delimiter` with `extract_values` to properly
split the line in both delimited and fixed-width cases.

Fixes #580.

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -109,7 +109,7 @@ class Engine(object):
                 self.table.record_id += 1
 
                 # Check for single row distributed over multiple lines
-                val_list = self.table.split_on_delimiter(line)
+                val_list = self.table.extract_values(line)
                 while len(val_list) < len(self.table.get_column_datatypes()):
                     line = line.rstrip('\n')
                     if type(real_lines) != 'list':
@@ -117,7 +117,7 @@ class Engine(object):
                     else:
                         line += real_lines[pos+1]
                         real_lines.pop(pos+1)
-                    val_list = (self.table.split_on_delimiter(line))
+                    val_list = (self.table.extract_values(line))
                 pos += 1
 
                 linevalues = self.table.values_from_line(line)


lib/engine.py
```diff
@@ -109,15 +109,15 @@ def source_gen():
                 self.table.record_id += 1
 
                 # Check for single row distributed over multiple lines
-                val_list = self.table.split_on_delimiter(line)
+                val_list = self.table.extract_values(line)
                 while len(val_list) < len(self.table.get_column_datatypes()):
                     line = line.rstrip('\n')
                     if type(real_lines) != 'list':
```
henrykironde: are we considering this `if type(real_lines) != list:`  or `if type(real_lines) != 'list':`

ethanwhite: If I understand correctly, you're pointing out that the `'list'` vs `list` issue from #579 isn't fixed here. That's just because I started working on this before #579 was merged and this is forked off of master. The merge should leave the changes from #579 in tact.

henrykironde:
@ethanwhite and @goelakash, could we actually have an example of the database and may be lines where this happens.

I know something similar was happening in the integrations. when we create the integration  datasets from the raw string and the engine bulk inserts it.
 But I guess in the case of integrations, we can clean that up as we create the file and replace all "\n".

ethanwhite:
I agree that we should look at a case where 9f05ea3 is needed, to make sure this is an issue with the raw data vs something we're introducing elsewhere and could fix. That said, I think we should go ahead and get this fix in so that things are working. We can always remove all of the code later if we find a better solution to the extra line endings.

goelakash:
@ethanwhite @henrykironde I found it in Line 68 in `Species.csv` from `SonoranPerennials` dataset. There may be more, but I thought it was generic enough that it could be solved easily.

ethanwhite:
@goelakash - ah yep, there it is. The things people do to data :smile:.

This is the sort of thing should be (and is) handled by modules like `csv`. At some point we may want to think about swapping in the use of `csv` to handle some of the complexity we're managing manually at the moment.

goelakash:
The cleaning should preferably be done in `Engine`, and since this isn't exclusive to csv theoretically, maybe we could refactor the code there a bit as well.

ethanwhite:
Agreed. I opened #584 as a starting point for discussion on this.

PR #583, opened 2016-07-19
refine conversion of strings to floats in the engine

henrykironde:
We did not handle the conversion of strings to floats appropriately.
In some dataset, where null values are not specified,
installing will give some error.
for example, in FrayJorge 'ND' looks like the null value and
since we can not change to float, the system will error,

In the current PR, if we can't change a value to float, we
take that as a null

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -565,10 +565,11 @@ class Engine(object):
                 return "null"
         elif datatype in ("double", "decimal"):
             if strvalue:
-                decimals = float(strvalue)
-                return str(decimals)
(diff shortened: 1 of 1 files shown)


henrykironde:
@ethanwhite can we have this in, I want to test the current errors we have

ethanwhite:
Yeah, sorry. I saw the "failing" check and thought you were still working on it, but it was just `codecov`. Might be worth checking if we can tweak it's yaml file to adjust what % change triggers a failed check.

PR #589, opened 2016-07-20
Changing default script type to JSON

goelakash:
@henrykironde @ethanwhite 
Needs some work. At the moment, the download tests are failing, with the engine saying that that `avianbodysize` and `mom2003` are not available, though they can be downloaded manually by `retriever download avianbodysize`/`mom2003`.


goelakash:
Will re-open after updating the branch.

PR #594, opened 2016-07-24
Cleanups and safety checks for JSON branch

goelakash:


goelakash:
Same `pyyaml` install failures as before on Python 3.4 on Appveyor.

Issue #595, opened 2016-07-29
Remove `pyyaml` dependency.


goelakash:
Currently affects `scripts-to-json` branch.

henrykironde:
Could you give a few lines of code that I could run to see that problem we are having clearly. Pyyaml doesn't seem to be maintained and it is hard to install on Mac. We should indeed remove it. I think there is a way to tell json to return either Unicode, which looks like the default or to use the defined encoding to get the string equivalent.

goelakash:
@henrykironde Currently there seems to be no way to return non-unicode strings using `json`. I am running into issues with Python 3 while implementing this solution - stackoverflow.com/questions/956867/how-to-get-string-objects-instead-of-aunicode-ones-from-json-in-python

henrykironde:
Looking at [the solution](http://stackoverflow.com/questions/956867/how-to-get-string-objects-instead-of-unicode-ones-from-json-in-python)

could this be one solution to removing the dependency

```
import json
data = """{"name": "Masses of Mammals", "shortname": "MoM2003", "description": "A data set of compiled"}"""
script_contents = json.dumps(json.loads(data))

new_script = open('json_file.py', 'w')
new_script.write("import os\n")
new_script.write("os.getcwd()\n")
new_script.write(script_contents)
new_script.close()
```

henrykironde:
@goelakash  another thing I tried

import json
original_list = ['a', 'b']
json_list = json.dumps(original_list)
json_string = json.dumps(json_list, ensure_ascii=False)
script_contents=json.loads(json_string)
print  json_list
new_script = open('json_file.py', 'w')
new_script.write("import os\n")
new_script.write("os.getcwd()\n")
new_script.write(script_contents)
new_script.close()

henrykironde:
@goelakash in edit_json we still have yaml https://github.com/weecology/retriever/blob/data-package/lib/datapackage.py#L376

when we are trying to limit editing of scripts to only json files  in the try block, it looks like you are trying to open a file but passing the script name 
`retriever edit_json MammalLH` 

```
    try:
        contents = yaml.load(
            open(os.path.join(HOME_DIR, 'scripts', json_file), 'r'))
    except FileNotFoundError:
        print("Script not found.")
        return
```

I am assuming this should have been something like what you have in line 384-385 where you add a postfix `.json`

goelakash:
@henrykironde The `yaml` dependency is removed, but now all the strings in python scripts have `u'` prepended to them.

[...]

PR #599, opened 2016-08-07
Add docs for new script creation and CLI bug-fixes

goelakash:
1. Docs now have information regarding new script creation with explanation of the JSON keys.
2. Fixed input and parsing issues with CLI
3. Removed `mediatype` attribute from JSON
4. Add compiled py files


goelakash:
@henrykironde

goelakash:
@henrykironde Why is the codecov/patch set at ~63% target?

henrykironde:
@goelakash I didn't set up that, but I will check.

henrykironde:
So we do not need to add the .py compiled files. 
we only add .py files that are manually tailored like BBS, gentry.

goelakash:
@henrykironde Removed those.

docs/scripts.rst
```diff
@@ -1,74 +1,128 @@
-===================================================== 
-Writing and loading Scripts for the EcoData Retriever
-===================================================== 
- 
+=====================================================
+Creating scripts for the EcoData Retriever
+=====================================================
```
henrykironde: this will error due to extrapolated "==="

lib/datapackage.py
```diff
@@ -205,51 +221,51 @@ def edit_dict(obj, tabwidth=0):
...
+                    elif selection == '5' or selection == "":
                         pass
                     else:
                         raise RuntimeError("Invalid input!")
 
                 elif isinstance(val, list):
 
                     for i in range(len(val)):
-                        print(str(val[i]))
+                        print(i+1, '. ', str(val[i]))
```
henrykironde: space

.gitignore
```diff
@@ -1,12 +1,10 @@
...
 build/
 connections.config
 sqlite.db*
-scripts/
 retriever.egg-info
 apidocs
 .DS_Store
 .cache/*
-*.bak
+scripts/*.py
```
henrykironde: I would think that it is good to block the entire scripts folder, unless otherwise

lib/datapackage.py
```diff
@@ -27,30 +27,30 @@ def create_json():
         if script_exists:
             print("Dataset already available. Check the list or try a different shortname")
 
-    contents['title'] = input("Title/Name: ")
-    contents['description'] = input("Description: ")
-    contents['citation'] = input("Citation: ")
-    contents['homepage'] = input("Site/Homepage of dataset: ")
```
henrykironde: The doc string in line 13, edit it to

```
def create_json():
    '''
    Creates datapackage.JSON script.
    
    http://specs.frictionlessdata.io/data-packages/#descriptor-datapackagejson
    Takes input from user via command line.
    Usage: retriever create_json
    '''
```

lib/datapackage.py
```diff
@@ -27,30 +27,30 @@ def create_json():
...
+    contents['title'] = input("Title/Name: ").strip()
+    contents['description'] = input("Description: ").strip()
+    contents['citation'] = input("Citation: ").strip()
+    contents['homepage'] = input("Site/Homepage of dataset: ").strip()
     tags = input("Tags (separated by ';'): ").split(';')
     contents['keywords'] = [tag.strip() for tag in tags if tag.strip() != ""]
     contents['resources'] = []
```
henrykironde: this could be refactored 
```
    while script_exists:
        contents['name'] = input("Shortname (Give a unique identifier for script): ")
        script_exists = contents['name'].lower() in short_names
        if script_exists:
            print("Dataset already available. Check the list or try a different shortname")

    contents['title'] = input("Title/Name: ").strip()
    contents['description'] = input("Description: ").strip()
    contents['citation'] = input("Citation: ").strip()
    contents['homepage'] = input("Site/Homepage of dataset: ").strip()
    tags = input("Tags (separated by ';'): ").split(';')
    contents['keywords'] = [tag.strip() for tag in tags if tag.strip() != ""]
    contents['resources'] = []
```

lib/datapackage.py
```diff
@@ -27,30 +27,30 @@ def create_json():
...
             table = {}
-            table['name'] = input("Table name: ")
-            table['url'] = input("Table URL: ")
+            table['name'] = input("Table name: ").strip()
+            table['url'] = input("Table URL: ").strip()
             table['dialect'] = {}
 
-            d_opts = ['nulls (separated by \';\')',  # list of str
-                      'replace_columns (separated by \';\')',  # list of tuples
+            d_opts = ['replace_columns (separated by \';\', with comma-separated values)',  # list of tuples
```
henrykironde: I think these are defaults for all scripts/tables and we do not need  read/initialize them  them all the time, So we can take them out of the loop.

lib/datapackage.py
```diff
@@ -60,13 +60,27 @@ def create_json():
                       'header_rows (int)']  # int
 
             for i in range(0, len(d_opts)):
```
henrykironde: Two ideas that may help to refactor:

 

1. 

I think we can use a dictionary in this case to avoid looping several times,
The dictionary could have this data structure  
`d_opts = { 'key' : ("prompt mesage","default-value"), ....}`

```
d_opts = { 
	'replacecoloms': ("replace_columns (separated by \';\', with comma-separated values)", ), 
	'nulls': ('nulls (separated by \';\')',)
	'delimiter': ('delimiter',)
	'do_not_bulk_insert': ('do_not_bulk_insert (bool = True/False)', 'False')
	'contains_pk': ('contains_pk (bool = True/False)',  'False')
 }

```

so while adding table you read up the d_opts and populate them based on user's input otherwise take the default.

2. 

we can create smaller functions to be specific like `get_replace_colums()`. The main idea is that if we want to debug something or add some new d_opts keyword like file_encoding, it becomes more easy

lib/datapackage.py
```diff
@@ -110,18 +124,20 @@ def create_json():
...
                                 print("\nWrong type. Either leave blank or try again.\n")
                         except:
                             print("Exception occured. Try Again.\n")
                             val = input(d_opts[i] + ": ")
 
             table['schema'] = {}
+
             table['schema']["fields"] = []
             print("Enter columns [format = name, type, (optional) size]:\n")
             col = input()
```
henrykironde: we could also create our own get_input (prompt, strip=True), not really important thou

lib/datapackage.py
```diff
@@ -139,25 +155,24 @@ def create_json():
...
-                while name.strip() != "":
+                print("Enter names of crosstab column values (Press return after each name):\n")
+                name = input().strip()
+                while name != "":
                     ct_names.append(name)
-                    name = input()
+                    name = input().strip()
 
                 table['schema']['ct_column'] = ct_column
                 table['schema']['ct_names'] = ct_names
```
henrykironde: line [182](https://github.com/weecology/retriever/pull/599/files#diff-fe19ce9787a421223628d4355dc2f5e4R182) 

```
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
         json.dump(contents, output_file, sort_keys=True, indent=4,
                   separators=(',', ': '))
         output_file.write('\n')
         print("\nScript written to " + file_name)
         output_file.close()
```

line [352](https://github.com/weecology/retriever/pull/599/files#diff-fe19ce9787a421223628d4355dc2f5e4R352)

```
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
         json.dump(contents, output_file, sort_keys=True, indent=4,
                   separators=(',', ': '))
         output_file.write('\n')
         print("\nScript written to " + os.path.join(HOME_DIR, 'scripts', file_name))
         output_file.close()
```

These can be refactored

lib/engine.py
```diff
@@ -706,7 +706,8 @@ def filename_from_url(url):
 
 
 def gen_from_source(source):
-    """Returns a generator from a source tuple.
+    """
+    Returns a generator from a source tuple.
```
henrykironde: I think both ways are okey. unless I am wrong

henrykironde:
In case some recommendations do not apply, please done hesitate to let me know.

.gitignore
goelakash: We should track changes to json scripts as often as we can, since they are not mature yet. For the python files that are compiled or written manually, we can force-add those files.

lib/datapackage.py
goelakash: I was thinking of writing an input function that gets an argument for the text prompt that is issued, and returns user input only if it is not "" after stripping.

lib/datapackage.py
goelakash: Okay, lets add that.

lib/engine.py
goelakash: This was to prevent a sphinx-build warning that I got at compile time.

ethanwhite: PEP257 says it should be as it was originally: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

goelakash: Okay, will revert.

lib/datapackage.py
```diff
@@ -9,11 +9,19 @@
 short_names = [script.shortname.lower() for script in SCRIPT_LIST()]
 
 
+def clean_input(prompt, strip=True):
+    val = input(prompt)
+    if strip:
+        val = val.strip()
+    if val == "":
```
henrykironde: why are we using a recursive call?

goelakash: I could use a loop, but this looks cleaner I suppose.

henrykironde:
when you are done with all changes let me know.

I have also tried to test this part 

```
def clean_input(prompt, strip=True):
    val = input(prompt)
    if strip:
        val = val.strip()
    if val == "":
        clean_input(prompt, strip)
    return val
```

but with an empty input , it doesn't seem to be giving me what I expected.

lib/datapackage.py
goelakash: I have edited it, but in a different manner.

goelakash:
@henrykironde  I have changed the `clean_input` function to cover a lot of cases now. See if you can find any issues. I will post some tests as well.

goelakash:
Will close #595

PR #601, opened 2016-08-13
Refactor CLI input for testability and add unit tests

goelakash:
Replaced recursion with iteration to preserve function state.
Added unit tests to test simple CLI input scenarios.

### DIFF
--- a/test/test_retriever.py
+++ b/test/test_retriever.py
@@ -2,9 +2,10 @@
 from future import standard_library
 standard_library.install_aliases()
 from imp import reload
+from future.builtins import input
 
 import os
-
+import pytest
 import sys
 reload(sys)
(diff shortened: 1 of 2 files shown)


test/test_retriever.py
```diff
@@ -240,3 +241,102 @@ def test_sort_csv():
...
+    """Test for non-null string"""
+    assert(is_empty("not empty") == False)
+
+
+def test_is_empty_not_empty_list():
+    """Test for not empty list"""
+    assert(is_empty(["not empty"]) == False)
+
+
+def test_clean_input_empty_input_ignore_empty(monkeypatch):
```
henrykironde: lets get another param name instead of` monkeypatch`

henrykironde: never mind with the `monkeypatch`

PR #606, opened 2016-08-18
Update scripts to datapackage.JSON standard

goelakash:
PR to merge the `data-package` branch to master.

PR #607, opened 2016-08-18
cleanup using autopep8

goelakash:
@henrykironde For the cleanup in #606


henrykironde:
looks great, could you also add clean up on imports.

PR #610, opened 2016-08-19
Use master to get latest JSON scripts

goelakash:
@henrykironde As discussed in #606 comments with @ethanwhite

PR #611, opened 2016-08-19
Update version.txt with JSON scripts

goelakash:
@henrykironde Please merge this before #610


ethanwhite:
We should leave the `.script` files in for the moment and remove them after `data-package` is in. If I understand everything right that way `master` will never be broken since it's OK for both the current `master` and `data-package` to have extra files in that directory that aren't being used. Does that make sense?

henrykironde:
@goelakash, let me know if you finish with the changes

goelakash:
@henrykironde Added the `.script` files.

PR #614, opened 2016-08-20
Correct behavior when empty dataset is specified

goelakash:
Add dataset name check before searching in scripts, and throw relevant error when no name is specified.
https://github.com/weecology/retriever/pull/606#discussion_r75546462

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -38,6 +38,7 @@ def main():
         script_list = SCRIPT_LIST()
 
         args = parser.parse_args()
+
         if args.quiet:
             sys.stdout = open(os.devnull, 'w')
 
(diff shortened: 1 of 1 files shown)


__main__.py
```diff
@@ -91,7 +92,7 @@ def main():
 
             for script in script_list:
                 if script.name:
-                    if args.l!=None:
+                    if args.l is not None:
```
henrykironde: I am not sure of args.l
I have seen it coming in from acaa7076241a36cb754021f1d450522b32a66ae3
but the implementation may have something wrong.

goelakash: @henrykironde The value of l is True when retriever ls is called with the -l flag.

Issue #623, opened 2016-08-24
Scripts in `./scripts` not being found?


ethanwhite:
After running `retriever reset scripts` when I try to do an installation from inside the repository I get a dataset unavailable error. This shouldn't happen since the data is in the local `scripts` directory.

```
ethan@stice:~/RetrieverCode (master)$ retriever install csv MoM2003
The dataset MoM2003 isn't currently available in the Retriever
Run 'retriever ls to see a list of currently available datasets
```

There was a recent discussion about not making a change that would have removed this functionality. Maybe it slipped through in the refactoring around the datapackage changes?

henrykironde:
I think that is because, we didn't include the search paths.  
I remember mention this in one of the PRs 
`ref: current master https://github.com/weecology/retriever/blob/master/__init__.py`
```
]
SCRIPT_WRITE_PATH = SCRIPT_SEARCH_PATHS[-1]
DATA_SEARCH_PATHS = [
    "./",
    "{dataset}",
    "raw_data/{dataset}",
    os.path.join(HOME_DIR, 'raw_data/{dataset}'),
]
```

goelakash:
This should be closed now.

ethanwhite:
Fixed by 88d6fcc5d8422c3b725dbe4b299679c35ceb9e63.

PR #624, opened 2016-08-24
Fixing script install issue in repo

goelakash:
Allow installing scripts even if the retriever HOME_DIR is reset.
#623 
@ethanwhite @henrykironde

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -44,7 +44,8 @@ for dir in (HOME_DIR, os.path.join(HOME_DIR, 'raw_data'), os.path.join(HOME_DIR,
             print("The Retriever lacks permission to access the ~/.retriever/ directory.")
             raise
 SCRIPT_SEARCH_PATHS = [
-    os.path.join(HOME_DIR, 'scripts/'),
+    'scripts',
+    os.path.join(HOME_DIR, 'scripts/')
 ]
 SCRIPT_WRITE_PATH = SCRIPT_SEARCH_PATHS[-1]
 DATA_SEARCH_PATHS = [


__init__.py
```diff
@@ -44,7 +44,8 @@
             print("The Retriever lacks permission to access the ~/.retriever/ directory.")
             raise
 SCRIPT_SEARCH_PATHS = [
-    os.path.join(HOME_DIR, 'scripts/'),
+    'scripts',
+    os.path.join(HOME_DIR, 'scripts/')
 ]
```
henrykironde: and i thing there is one missing that is the ./

goelakash: I think that existed because earlier, users could create a script and keep it in the home directory. With the current setup, there is no need to search the home dir, either the scripts are in `.retriever/scripts`, or they exist inside the `scripts` fodler in the repo.

ethanwhite: The original concept of search `.` was to facilitate users including their own scripts that they hadn''t contributed to the retriever. That said, I'm not aware of folks using this, but that doesn't mean that they aren't. Is there a reason that you see for removing it?

henrykironde:
@goelakash, I will merge this but give some input(Ethan's comment) or feel free to change based on my next PR.

__init__.py
goelakash: @ethanwhite with the json, the only way that a user can create a script is by using the CLI, after which they are stored in .retriever/scripts directory. There does not seem to be any case where a user would want to move a script from this directory to her home directory. Instead, if there are some stray JSON files in the user's home directory, they would get picked up by retriever, and give a compilation error.

henrykironde: Okey, I get what you mean. In this case we assume that only CLI is the only way to create a script.
This sounds good but some geeky individuals want to manually create their own `.json` scripts in the current `"./"` directory and then run them. 

That being said, to account for such rare cases, I have thought of introducing a keyword `"json_type": "retriever_script" `. This can enable us to only compile scripts made for the retriever. 
I have sent you the PR but it is failing on travis and appevyor but passing tests locally.

PR #625, opened 2016-08-25
include the current working directory in SCRIPT_SEARCH_PATHS

henrykironde:


goelakash:
@henrykironde This won't be necessary. Creating JSON scripts by hand is/should be discouraged. If a user wants, they can edit the script by hand, but they don't need to move it to their home directory also.

ethanwhite:
Since we have, thanks to @goelakash's awesome work, moved over to the data package JSON standard I do think there are cases where JSON files could be created separately from our interface. In particular one of my hopes is that we will be able to (probably with a bit of additional work) be able to use data package JSON files generated for data through the broader data package effort being run by OKFN. In this case it seems possible/likely that someone would want to download a datapackage.json file that isn't part of the retriever and use it for a single project. We could force them to put it in a local `scripts` directory, but I think the extra flexibility of being able to put it in the working directory is useful.

As @goelakash and @henrykironde have both pointed out there are issues with non-script JSON files in the working directory currently breaking things. Therefore I propose that we do make this change, but not until we've dealt with the issue of correctly identifying JSON files that are data package files and not processing other JSON files (@henrykironde is actively working on this). How does that sound to everyone?

henrykironde:
changes in this pull request have been added to #626.
I will close this later

henrykironde:
Just to add on like @goelakash stated, this may not necessary. 
I think for end users they may not need it a lot.

 For the case of developers. this helps and saves a lot of time when you debugging data sets.

Issue #318, opened 2015-05-08
Overhaul script updating system


ethanwhite:
The script updating system needs some attention so I'm creating this issue to house high level discussion and planning for this update.
#### 1. Do we want to have separate software releases and script updates?

The current design is intended to let us updating scripts without a new software release.

The nice thing about this is that it's relatively easy to add or update a script, so we might not want to have to roll a new release every time we get a few of these. Perhaps more importantly users may not update the retriever with every release and so if we fix a bug in a script we want them to get the update without any extra work.

The downsides are two fold. First, this makes provenance and reproducibility a little trickier. You need to know both what version of the retriever was used, but also what version of the scripts. Second, it means that the core code is more complicated because it has to handle checking for new scripts and downloading them. We've had a number of issues over the last year related to complexities in this system and as a result it isn't currently accomplishing the goal of actually letting us update scripts.

If we want to keep this separation then there are a number of things below that we need to think about. If we don't then we would just start distributing the scripts with each release.
#### 2. Scripts should not be tied to a release (or be so tied in a more sophisticated way)

Currently the scripts being downloaded are tied to an individual release. When the retriever grabs scripts it looks for them under the tag for that release. This sort of worked a while back when releases were in separate branches, because you could go update the branch for each release with the new script. This wasn't a great solution, but it worked. Now that we are using tags for releases this doesn't work any more because tags are intended to be immutable.

The challenge is doing this in such a way that it allows for changes or additions. In the past we accidentally broke older versions of the retriever by adding scripts for new (just developed) database engines that the older versions of the retriever didn't know about. It tried to load the scripts and errored. We've fixed this in a718c8d798903f1e096c4fb5ca0da9354a1dd536 so that at least things will keep working, and maybe that's enough. We may also just want to view anything that would cause challenges to the script system as a major version upgrade.
#### 3. Improve checking of scripts for updates

The current implementation is intended to check to see if updated scripts exist by comparing hashes of the existing scripts with published hashes of the currently available scripts. If the hashes don't match then the script is supposed to be updated. This has caused issues with development versions resulting in 495d88dda5cdae17e1067759c16080e21f6c3d41, but this means that leads to an error prone flag change with releases to indicate whether the retriever is running as a release or development version.

ethanwhite:
#### Initial thoughts

1. I like @bendmorris's original design of allowing updating scripts without a full release and so I'd like to keep it.
[...]

PR #626, opened 2016-08-26
test retriever using downloaded scripts

henrykironde:
The previous version of tests updated scripts
fetching the master `scripts` and then
used the updated scripts for testing

This commit ensures that the retriever used the local script directory.
we get the `scripts` directory using the path of the test module location.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -44,6 +44,7 @@ for dir in (HOME_DIR, os.path.join(HOME_DIR, 'raw_data'), os.path.join(HOME_DIR,
             print("The Retriever lacks permission to access the ~/.retriever/ directory.")
             raise
 SCRIPT_SEARCH_PATHS = [
+    "./",
     'scripts',
     os.path.join(HOME_DIR, 'scripts/')
 ]
--- a/lib/compile.py
+++ b/lib/compile.py
@@ -141,12 +141,12 @@ def compile_script(script_file):
 
 
 def add_dialect(table_dict, table):
-    '''
+    """
     Reads dialect key of JSON script and extracts key-value pairs to store them
     in python script
 
     Contains properties such 'nulls', delimiter', etc
-    '''
+    """
     for (key, val) in table['dialect'].items():
         # dialect related key-value pairs
         # copied as is
@@ -161,13 +161,13 @@ def add_dialect(table_dict, table):
 
 
 def add_schema(table_dict, table):
-    '''
+    """
     Reads schema key of JSON script and extracts values to store them in
     python script
 
     Contains properties related to table schema, such as 'fields' and cross-tab
     column name ('ct_column').
-    '''
+    """
     for (key, val) in table['schema'].items():
         # schema related key-value pairs
 
@@ -178,14 +178,14 @@ def add_schema(table_dict, table):
             column_list = []
             for obj in val:
                 # fields is a collection of JSON objects
-                #(similar to a list of dicts in python)
+                # (similar to a list of dicts in python)
 
                 if "size" in obj:
                     column_list.append((obj["name"],
                                         (obj["type"], obj["size"])))
                 else:
                     column_list.append((obj["name"],
-                                        (obj["type"], )))
+                                        (obj["type"],)))
 
             table_dict["columns"] = column_list
 
@@ -197,12 +197,16 @@ def add_schema(table_dict, table):
 
 
 def compile_json(json_file):
-    '''
+    """
     Function to compile JSON script files to python scripts
     The scripts are created with `retriever create_json <script_name` using
     command line
-    '''
+    """
     json_object = json.load(open(json_file + ".json", "r"))
+
+    if "retriever" not in json_object.keys():
+        # Compile only files that have retriever key
+        return
 
     values = {}
     values['urls'] = {}
@@ -233,7 +237,7 @@ def compile_json(json_file):
             # Array of table objects
             tables = {}
             for table in value:
-                 # Maintain a dict for table keys and values
+                # Maintain a dict for table keys and values
                 table_dict = {}
 
                 try:
--- a/lib/parse_script_to_json.py
+++ b/lib/parse_script_to_json.py
@@ -15,6 +15,7 @@ def parse_script_to_json(script_file, location=SCRIPT_DIR):
     replace = []
     keys_to_ignore = ["template"]
     urls = {}
+    values["retriever"] = "True"
 
     for line in [str(line).strip() for line in definition]:
         if line and ':' in line and not line[0] == '#':
--- a/test/test_regression_cli.py
+++ b/test/test_regression_cli.py
@@ -9,10 +9,13 @@ if hasattr(sys, 'setdefaultencoding'):
     sys.setdefaultencoding('latin-1')
 import pytest
 from retriever.lib.tools import getmd5
-from retriever import HOME_DIR, ENGINE_LIST
+from retriever import SCRIPT_SEARCH_PATHS, ENGINE_LIST
 
 mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, csv_engine, download_engine, json_engine, xml_engine = ENGINE_LIST()
 dir_path = os.path.dirname(os.path.abspath(__file__))
+site_root = os.path.dirname(os.path.realpath(__file__))
+parent = os.path.abspath(os.path.join(site_root, os.pardir))
+HOMEDIR = os.path.expanduser('~')
 
 download_md5 = [
     # ('DelMoral2010', '0'),
@@ -28,8 +31,13 @@ db_md5 = [
 
 
 def get_script_module(script_name):
-    """Load a script module"""
-    file, pathname, desc = imp.find_module(script_name, [os.path.join(HOME_DIR, "scripts")])
+    """Load a script module from the downloaded scripts directory in the retriever
+
+    This ensures that tests are performed on the retriever source downloaded
+    Since this module is run in the test directory, we get the parent directory
+    using the module's location
+    """
+    file, pathname, desc = imp.find_module(script_name, [os.path.join(parent, "scripts")])
     return imp.load_module(script_name, file, pathname, desc)
 
 
@@ -47,7 +55,6 @@ def get_csv_md5(dataset, engines, tmpdir):
 def setup_module():
     """Update retriever scripts and cd to test directory to find data"""
     os.chdir(os.path.dirname(os.path.abspath(__file__)))
-    os.system("retriever update")
 
 
 def teardown_module():
@@ -109,7 +116,11 @@ def test_csv_regression(dataset, expected, tmpdir):
 
 @pytest.mark.parametrize("dataset, expected", download_md5)
 def test_download_regression(dataset, expected):
-    """Check for regression for a particular dataset downloaded only"""
+    """Check for regression for a particular dataset downloaded only
+
+    "Move the dataset script to the .retriever directory before download."
+    """
+    os.system('cp -r {0} {1}'.format(os.path.join(parent, "scripts"), os.path.join(HOMEDIR, '.retriever')))
     os.system("retriever download {0} -p raw_data/{0}".format(dataset))
     current_md5 = getmd5(data="raw_data/{0}".format(dataset), data_type='dir', mode="rU")
     assert current_md5 == expected
--- a/test/test_regression_cli.py
+++ b/test/test_regression_cli.py
@@ -9,12 +9,12 @@ if hasattr(sys, 'setdefaultencoding'):
     sys.setdefaultencoding('latin-1')
 import pytest
 from retriever.lib.tools import getmd5
-from retriever import SCRIPT_SEARCH_PATHS, ENGINE_LIST
+from retriever import ENGINE_LIST
 
 mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, csv_engine, download_engine, json_engine, xml_engine = ENGINE_LIST()
-dir_path = os.path.dirname(os.path.abspath(__file__))
-site_root = os.path.dirname(os.path.realpath(__file__))
-parent = os.path.abspath(os.path.join(site_root, os.pardir))
+file_location = os.path.dirname(os.path.realpath(__file__))
+retriever_root_dir = os.path.abspath(os.path.join(file_location, os.pardir))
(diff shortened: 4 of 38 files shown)


test/test_regression_cli.py
```diff
@@ -109,7 +116,11 @@ def test_csv_regression(dataset, expected, tmpdir):
 
 @pytest.mark.parametrize("dataset, expected", download_md5)
 def test_download_regression(dataset, expected):
-    """Check for regression for a particular dataset downloaded only"""
+    """Check for regression for a particular dataset downloaded only
+
+    move the dataset script downloaded to the .retriever file before download
```
ethanwhite: "Move the dataset script to the .retriever directory before download."

Missing space after comma.

Why do we need to move the script in this case, but not in the others?

test/test_retriever.py
```diff
@@ -246,107 +244,122 @@ def test_sort_csv():
...
         mock_input.counter += 1
         if mock_input.counter <= 1:
             return ""
         else:
             return "not empty"
+
     mock_input.counter = 0
     monkeypatch.setattr('retriever.lib.datapackage.input', mock_input)
-    assert(clean_input("", ignore_empty=False) == "not empty")
+    assert (clean_input("", ignore_empty=False) == "not empty")
```
ethanwhite: The extra parentheses should be removed here in the same way you've been removing them further down.

ethanwhite:
There's a lot going on in this PR which makes it a little difficult to review. Are most of the changes not related to testing just style cleanup that is diffing badly of should I be trying to meaningfully review the changes in these other areas (e.g., `compile.py`)?

test/test_regression_cli.py
henrykironde: Good question. So the scripts search is current directory, and then scripts directory and .retriever directory. 

In the case of the other tests using `imp.load_module` we give them the path to the script directory. In download only tests, since the scripts cant be  hit 
there are two ways 
Either cd .. to move from current directory` test` to retriever main directory or to copy the scripts in to the .retriever directory.

henrykironde: The only data set that was not getting found was mom2003 since we don't have it in the raw folder.
So I decided to use copy. We can definitely improve it.

henrykironde:
Sorry @ethanwhite, let me explain all the changes that need a review. Others are style changes. I will let you know when i fininsh.

__init__.py
```diff
@@ -44,6 +44,7 @@
             print("The Retriever lacks permission to access the ~/.retriever/ directory.")
             raise
 SCRIPT_SEARCH_PATHS = [
+    "./",
     'scripts',
```
henrykironde: `./` this we talked about in the previous PR so on going discussion.

ethanwhite:
Sounds good. Thanks.

lib/parse_script_to_json.py
```diff
@@ -15,6 +15,7 @@ def parse_script_to_json(script_file, location=SCRIPT_DIR):
     replace = []
     keys_to_ignore = ["template"]
     urls = {}
+    values["retriever"] = "True"
```
henrykironde: This change sets the retriever json files with `"retriever"= "True"` to help identify the right json to compile.

however, I did not see the need to also include it in the written `.py` file since we already have `#retirver ` tag in the `.py` files

lib/compile.py
```diff
@@ -141,12 +141,12 @@ def get_value(key):
```
henrykironde: We do not change anything major in this file. I changed the style in some cases.

test/test_regression_cli.py
```diff
@@ -9,10 +9,13 @@
     sys.setdefaultencoding('latin-1')
 import pytest
 from retriever.lib.tools import getmd5
-from retriever import HOME_DIR, ENGINE_LIST
+from retriever import SCRIPT_SEARCH_PATHS, ENGINE_LIST
 
 mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, csv_engine, download_engine, json_engine, xml_engine = ENGINE_LIST()
 dir_path = os.path.dirname(os.path.abspath(__file__))
+site_root = os.path.dirname(os.path.realpath(__file__))
```
henrykironde: used to get the scripts directory

test/test_retriever.py
```diff
@@ -358,28 +371,26 @@ def test_add_dialect():
...
-    assert(table_dict == result)
+    assert table_dict == result
 
 
 def test_add_schema():
     """Test adding values from schema key to python script"""
-    table = {}
-    table['schema'] = {}
     table_dict = {}
+    table = {'schema': {}}  # dictionary literal for dictionary creation
```
henrykironde: here I used dictionary literal for dictionary creation of the dictionary.

henrykironde:
@ethanwhite, all major changes are commented. The json script changes are from adding the `retriever  True` .

test/test_retriever.py
ethanwhite: Is there a reason not to just write these entirely as a dictionary literal instead of using the appends and assignments. E.g., for the `result` dictionary below I think this is clearer:

```
result = {'columns': [('col1', ('int',)), ('col2', ('char', 20))],
          'ct_column': "'ct_column'",
          'ct_names': ['ct1', 'ct2', 'ct3', 'ct4'],
          'dummy_key': 'dummy_value'
}
```

I know you didn't write these originally, but if we're going to change part of this here we might as well make the full change (or not make any changes here and make the full change elsewhere.

test/test_regression_cli.py
ethanwhite: > Either cd .. to move from current directorytest into the retriever main directory or to copy the scripts in to the .retriever directory.

My initial thought was to not cd into `test` in the first place. We are only doing this because the data for DelMoral and AvianBodySize is stored in `test/raw_data`. If as part of `setup_module` we copy `./test/raw_data` to `./raw_data` the data should be accessible. So we would remove the `os.chdir(os.path.dirname(os.path.abspath(__file__)))` line and replace it with something a system call that does the `raw_data` copying. The `teardown_module` looks like it's probably handling the cleanup of that already.

I'm happy to go with whatever you want, so if you like the current implementation better just let me know.

ethanwhite:
Looks great. See my inline comments for a couple of suggestions. Change whichever of those you find useful and we'll get this in.

lib/compile.py
```diff
@@ -197,92 +197,95 @@ def add_schema(table_dict, table):
...
+    """
     json_object = json.load(open(json_file + ".json", "r"))
 
-    values = {}
-    values['urls'] = {}
-
-    keys_to_ignore = ["template"]
-
-    for (key, value) in json_object.items():
+    if "retriever" in json_object.keys():
```
goelakash: I think that instead of putting the whole code inside this condition, we could instead print an error and return.

test/test_regression_cli.py
henrykironde: Thanks Ethan, I intend to leave them the way they are for this PR/commit. However we are going to change then for #318

henrykironde:
@ethanwhite, @goelakash, did make changes are requested.
@ethanwhite, `Is there a reason not to just write these entirely as a dictionary literal instead of using the appends and assignments. E.g., for the result dictionary below I think this is clearer:`
I have kept the original version.

@goelakash, I have made changed to a negation with a return.

Let me know in case there more changes

ethanwhite:
LGTM. @goelakash let me know if it looks ready to you as well. If it does we'll get this merged.

test/test_regression_cli.py
```diff
@@ -54,17 +48,17 @@ def get_csv_md5(dataset, engines, tmpdir):
...
     """Update retriever scripts and cd to test directory to find data"""
-    os.chdir(os.path.dirname(os.path.abspath(__file__)))
+    os.chdir(rood_dir)
+    os.system('cp -r {0} {1}'.format(os.path.join(rood_dir, "test/raw_data"), rood_dir))
 
 
 def teardown_module():
     """Cleanup temporary output files after testing and return to root directory"""
-    os.chdir(os.path.dirname(os.path.abspath(__file__)))
+    os.chdir(rood_dir)
```
henrykironde: In case tests fail/error in one module, we are sure that these module tests will begin from the right directory. 
This error used to be present where some modules error and the next module run errors if we do not set it to the right directory

test/test_regression_cli.py
```diff
@@ -54,17 +48,17 @@ def get_csv_md5(dataset, engines, tmpdir):
 
 def setup_module():
     """Update retriever scripts and cd to test directory to find data"""
-    os.chdir(os.path.dirname(os.path.abspath(__file__)))
+    os.chdir(rood_dir)
+    os.system('cp -r {0} {1}'.format(os.path.join(rood_dir, "test/raw_data"), rood_dir))
```
henrykironde: we are keeping the raw_data in the test directory and we only copy it the retriever main directory.

henrykironde:
done! :)

test/test_regression_cli.py
ethanwhite: Are we changing the working directory anywhere to anything other than the root directory anywhere in the tests now? If not then we shouldn't need all of this directory changing code. Your totally right that in the old setup we were moving into `test` and then back out of it in some cases, so if we changed in and then failed, we might not have gotten changed out. I'm just wondering if that sort of issue is still possible.

ethanwhite:
Great!

1. `root` not `rood` throughout
2. As I mentioned inline, I'm wondering if we're changing the working directory anywhere to anything other than the root directory anywhere in the tests now? If not then we shouldn't need all of this directory changing code. Your totally right that in the old setup we were moving into test and then back out of it in some cases, so if we changed in and then failed, we might not have gotten changed out. I'm just wondering if that sort of issue is still possible. My guess is that everything should be happening from root at this point in which case we can drop all of the directory changing code now that you're copying the data files into the working directory.

henrykironde:
Thanks Ethan, I think i found the issue.. let me try one thing and will let you know 
```
    workdir = tmpdir.mkdtemp()
    workdir.chdir()
```

This was the cause of the unexpected behavior . Will ping you if i rectify this.

test/test_regression_cli.py
```diff
@@ -46,18 +49,17 @@ def get_csv_md5(dataset, engines, tmpdir):
 
 def setup_module():
     """Update retriever scripts and cd to test directory to find data"""
-    os.chdir(os.path.dirname(os.path.abspath(__file__)))
-    os.system("retriever update")
+    os.chdir(site_dir)
+    os.system('cp -r {0} {1}'.format(os.path.join(site_dir, "test/raw_data"), site_dir))
```
henrykironde: This step is important unless we move the raw_data directory from the test to the  retriever main directory. I prefer leaving it in test and just copy it

test/test_regression_cli.py
```diff
@@ -9,10 +9,13 @@
...
 from retriever.lib.tools import getmd5
-from retriever import HOME_DIR, ENGINE_LIST
+from retriever import ENGINE_LIST
 
 mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, csv_engine, download_engine, json_engine, xml_engine = ENGINE_LIST()
-dir_path = os.path.dirname(os.path.abspath(__file__))
+file_location = os.path.dirname(os.path.realpath(__file__))
+site_dir = os.path.abspath(os.path.join(file_location, os.pardir))
+working_script_dir = os.path.abspath(os.path.join(site_dir, "scripts"))
+HOMEDIR = os.path.expanduser('~')
```
henrykironde: This concept was the main cause of all test failures. 
accessing the

 ```
SCRIPT_WRITE_PATH or DATA_SEARCH_PATHS
```
variables in tests, will not return the desired or intended path values. 
The issue is when we install retriever and run tests some of these paths if not used well will return paths relative to where the retriever was installed. 

Accessing `HOMEDIR` will always return a constant values. 

To set paths in test we should be careful and its better to start from
the location of the file `os.path.dirname(os.path.abspath(__file__))` and we find the retriever main directory.

**Comment by henrykironde** [comment]
@ethanwhite,  Thanks for the previous comments, We did not previously set paths right. currently I think we are on the right track or almost there.

**Comment by ethanwhite** [comment]
Thanks @henrykironde. Should I give this another quick review at this point?

**Comment by henrykironde** [comment]
yes

File: test/test_retriever.py [hunk 13]
```diff
@@ -32,16 +30,19 @@
...
+file_location = os.path.dirname(os.path.realpath(__file__))
+site_dir = os.path.abspath(os.path.join(file_location, os.pardir))
 
 
 def setup_module():
-    """"change directory to test directory"""
-    os.chdir(os.path.dirname(os.path.abspath(__file__)))
+    """"Make sure you are in the main local retriever directory"""
+    os.chdir(site_dir)
+    os.system('cp -r {0} {1}'.format(os.path.join(site_dir, "test/raw_data"), site_dir))
```
ethanwhite: Do we use the raw data in this test module? Since this is all unit tests I'd expect that we don't, in which case we could drop this line here.

ethanwhite:
Looks great @henrykironde. I made a minor comment or two in-line.

My one last question is about the naming of `site_dir`. This isn't a totally intuitive name to me. We basically mean the root working directory. I could see `root_work_dir` or `retriever_root_dir` or even just `workdir` (though I know we use that when changing to temp directories later. One option would be to use `workdir` and make the temp directories use `temp_workdir` instead. Pick what you think is best (or explain why `site_dir` makes more sense) and we'll get this merged.

test/test_retriever.py
henrykironde: I think `test_find_file_present()` uses that to test for the search path `"raw_data/{dataset}"`

henrykironde:
will change to `retriever_root_dir`

henrykironde:
Made the change of name from `site_dir` to `retriever_root_dir`.
We still copy the data. I edited the docstring to mention that.

```
def test_find_file_present():
    """Test if existing datafile is found

    Using the AvianBodySize dataset which is included for regression testing.
    We copy the raw_data directory to retriever_root_dir which is the current working directory.
    This enables the data to be in the DATA_SEARCH_PATHS.
    """
    test_engine.script.shortname = 'AvianBodySize'
    assert test_engine.find_file('avian_ssd_jan07.txt') == os.path.normpath(
        'raw_data/AvianBodySize/avian_ssd_jan07.txt')
```
Let me know what you think

ethanwhite:
LGTM. Will merge once the tests finish.

ethanwhite:
Nice work @henrykironde. Thanks.

Issue #628, opened 2016-08-30
Update Developer Docs for MySQL Installation


zhangcandrew:
**
For MySQL installation, while going through the inital setup and installation process, it is important to let the user know that access and file privileges must be granted to travis on their new database testdb on the localhost.

If this is not addressed, the following error will occur:

Couldn't create database ((1045, "28000Access denied for user 'root'@'localhost' (using password: YES)")). Trying to continue anyway.
Creating table BBS.species...
Couldn't create table ((1045, "28000Access denied for user 'root'@'localhost' (using password: YES)")). Trying to continue anyway.
(1045, "28000Access denied for user 'root'@'localhost' (using password: YES)")

**

ethanwhite:
Some of this information is currently in the [Continuous Intregration section of the docs](https://retriever.readthedocs.io/en/latest/developer.html#services-used). I agree that it would be better to move it up under [Testing](https://retriever.readthedocs.io/en/latest/developer.html#testing) since it really applies to local testing instead of CI. Would you like to try a PR moving that docs section up?

zhangcandrew:
Sure! I was going solely off the readMe in the retriever gitHub Repository initially, and only found out about the travis mySQL grants through talking to Henry. Just wondering though, what does PR stand for?

henrykironde:
:) PR is a Pull request. I was confused the first time I heard it.  you can call me to go through the workflow of that.

ethanwhite:
This was fixed in 0190f7824f4769d893134ca95b2cae8080a491a1. Thanks @zhangcandrew!

PR #631, opened 2016-09-01
Edited Developer docs

zhangcandrew:
Developer docs were edited.


henrykironde:
@zhangcandrew thanks for the PR, sorry for making you delete the first branch before we merged.
I will let @ethanwhite go through this before we merge.

ethanwhite:
Thanks for helping us improve the documentation @zhangcandrew! I have a few suggestions for improving the PR and some questions to help my understand particular decisions you've made. This is really common and it's part of why we (and lots of other projects) use the PR/review process to help make everything that gets merged be as good as it can be.

1. This change suggests that it is necessary to create a user before running the listed commands, but the `GRANT ALL PRIVILEGES ON testdb.* TO 'travis'@'localhost';` should create the user as part of the process. I tested this on Linux by first running `DROP USER 'travis'@'localhost';` and then running the retriever tests.
2. The retriever should create automatically create `testdb` as long as the privileges are in place. I tested this by first running `DROP DATABASE testdb` and then running the tests.

Can you try this and see if on your system it's actually necessary to separately create the user and the database?

The other big thing I think we want to do here is separate the `Continuous Integration` section from a more general `Testing` section. So what I was thinking was to move everything in the `Services Used` section starting at "After installing the servers..." to the beginning of the `Running tests locally section` and staring it something like "To run the test you will need to have all of the relevant database management systems and associated modules installed (see Setting up servers). You will then need to create the appropriate permissions for the tests to access the databases. You can do this by running the following commands in MySQL and PostgreSQL and creating the `.pgpass` file as described below:"

Depending on your answers to my questions above, if we need additional commands they should be added into these sections instead of stated more broadly. The goal is to make it so that when setting up a new system I can just cut and paste the relevant blocks into the DBMS and be ready to go.

Finally, we want the first line of commit messages to communicate as much information as possible about what the change accomplishes. In this case I'd probably use something like: "Move docs on local testing to better location".

ethanwhite:
Oh, and one other thing. You mentioned in #628 that you were originally just looking at the README. While we have a link to the docs in a badge at the top I think missing that will be common. So an additional change that might be useful either here or in a separate PR (probably a separate PR) is adding a small section to the README directing both users and developers to the documentation.

zhangcandrew:
I liked making my first pull request! it was really great using gitHub to a higher more dynamic workflow than what I usually had it for. 

So the problem I had brought up with Henry was that when I tried to run the Grant Privileges command:
    'GRANT ALL PRIVILEGES ON testdb.* TO 'travis'@'localhost';'

My version of MySQL returns this error: 
    'ERROR 1819 (HY000): Your password does not satisfy the current policy requirements'

So Henry had mentioned I could have run into some problems due to the travis user having not been created yet along with testdb, and recommended I look through and change some parts of the developer docs. I'm still working on curbing this problem however. Since it seems to be working for you, it might be a problem with my MySQL setup regarding something I have to change that I have not discovered yet. 

I can make the appropriate changes too for the doc, shifting around the sections and all, and I'll get to work on creating that new PR!

docs/developer.rst
```diff
@@ -36,6 +36,59 @@ You will also need the following modules:
...
+For MySQL specifically, you need to create a new user called travis, and also a new database called testdb. This new user should be separate from the default MySQL user of root. After this is fulfilled, we can run the following commands to grant privileges. 
+
+::
+
+  MySQL
+  -----
+  GRANT ALL PRIVILEGES ON testdb.* TO 'travis'@'localhost';
+  GRANT FILE ON *.* TO 'travis'@'localhost';
+  ​
+  Install MySQL on Mac
```
goelakash: @henrykironde We should probably not include the installation instructions for mysql/psql in the docs, or if we do, then we should mention all platforms.

henrykironde: SO I need to test this on a new system. In postgres we basically first create a new user by default `postgres`  with password `testpass`,
`psql -c "CREATE USER postgres WITH PASSWORD 'testpass'"`.
then we create a db `testdb` and grant all privileges to the user on testdb.
In MYSQL, we assume that running the two commands does the same

```
 GRANT ALL PRIVILEGES ON testdb.* TO 'travis'@'localhost';
 GRANT FILE ON *.* TO 'travis'@'localhost';

```
but I think the case may be different. On @zhangcandrew's machine the same failed and I think I had this issue once. We are still trying to figure it out.

On that note @zhangcandrew, could you try to run this command 
https://github.com/weecology/retriever/blob/master/appveyor.yml#L60
`mysql -e "CREATE USER 'travis'@'localhost';GRANT ALL PRIVILEGES ON *.* TO 'travis'@'localhost';GRANT FILE ON *.* TO 'travis'@'localhost';" -uroot`

ethanwhite:
> We should probably not include the installation instructions for mysql/psql in the docs, or if we do, then we should mention all platforms.

I agree that the installation instructions for MySQL using homebrew on OS X should be removed from here and I think this PR makes sense as a good place to do that. Thanks @goelakash.

ethanwhite:
Also - @zhangcandrew meet @goelakash, @goelakash meet @zhangcandrew.

Akash was a Google Summer of Code student with us this last summer and has made some great contributions to the project including Python 3 support (he's our resident expert on joint Python 2/3 compatible code) and moving us over to the datapackage.json standard for storing metadata about simple datasets. 

Andrew is an undergraduate at the University of Florida just starting his sophomore year who is planning on joining us to work on the Data Retriever to help get some experience working on an open source project.

goelakash:
We should definitely add instructions on how to create users on MySQL and PostgreSQL, for both Windows and Unix platforms. Let me recreate the users on Windows, so that @zhangcandrew  can add that to this PR as well.

zhangcandrew:
Hi @goelakash; thank you for the introduction @ethanwhite! So just to clarify for my own purposes, we should not include any instructions on initial installation of mysql and postgresql, but we should have instructions on creating travis as a user right? Commands to be run like: 

`CREATE USER 'travis'@'localhost'`.

docs/developer.rst
zhangcandrew: The command you gave me seemed to work fine for me @henrykironde! The only change I made to the command though was adding the argument -p since I had a password setup for my root user on mysql, but if the person is using the default settings, there should be no password setup initially for root, so I think it would work without that argument. Should I add that bit to what we've been writing?

ethanwhite:
> we should not include any instructions on initial installation of mysql and postgresql, but we should have instructions on creating travis as a user right?

Yes, exactly.

ethanwhite:
hey @zhangcandrew - I know this got a little complicated pretty quickly. Does everything here make sense at this point or would some clarification be useful?

zhangcandrew:
Hey @ethanwhite, I think I understand mostly what needs to be done, but clarification would be helpful. So on my list of things to add I have: 

Move "services used" section to beginning of "running tests locally" section
remove installation instructions for mysql to retain uniformity
Add commands to create users in mysql and postgresql (@goelakash is working on windows users I believe)


If I'm missing things, let me know, I want to get as much as I can in the next change so I don't clutter anything up!

ethanwhite:
@zhangcandrew I think you've got it all.

docs/developer.rst
```diff
@@ -149,6 +117,33 @@ After installing the servers we need to configure them by granting privileges to
...
 
+To run tests we use pytest.
+From the source top level directory, run
+
+.. code-block:: sh
+
+  $   py.test
+
+
+In case we want to run tests on a specific test category, we add the path of the test module, py.test [path]
```
ethanwhite: "In case we want to run tests on a specific test category, we add the path of the test module, py.test" -> "To run tests on a specific test category add the path of the test module to the end of the py.test command"

zhangcandrew: Just finished the appropriate changes!

ethanwhite:
Thanks @zhangcandrew. Looks great! In it goes.

ethanwhite:
I used a squash & merge here since this is a case where we really just need a single commit to describe the final changes, so you'll only see a single commit in the git history.

PR #633, opened 2016-09-03
Re-add edit_json feature

goelakash:
EDIT: Fails before running `retriever update` once.


henrykironde:
@goelakash thanks, will test this later today.

PR #636, opened 2016-09-06
Adding multi row insert to table.

henrykironde:
This feature (multi row insert )works with newer versions of sQlite.
3.7 and above. Most machines are running new version.

fixes: #312

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ python:
   - "3.5"
 addons:
   postgresql: "9.4"
+  apt:
+    sources:
+      - travis-ci/sqlite3
+    packages:
+      - sqlite3
 
 services:
   - postgresql
@@ -23,6 +28,7 @@ script:
 
 # database setup
 before_script:
+  - sqlite3 --version
   - mysql -e 'create database myapp_test;'
   - psql -c 'create database testdb;' -U postgres
 
--- a/engines/csvengine.py
+++ b/engines/csvengine.py
@@ -2,6 +2,7 @@ from builtins import str
 from builtins import object
 
 import os
+import csv
 
 from retriever.lib.models import Engine
 from retriever import DATA_DIR
@@ -45,6 +46,7 @@ class engine(Engine):
          "Format of table name",
          os.path.join(DATA_DIR, "{db}_{table}.csv")),
     ]
+    table_names = []
 
     def create_db(self):
         """Override create_db since there is no database just a CSV file"""
@@ -53,27 +55,23 @@ class engine(Engine):
     def create_table(self):
         """Create the table by creating an empty csv file"""
         self.auto_column_number = 1
-        self.output_file = open(self.table_name(), "w")
-        self.output_file.write(
-            ','.join(['"%s"' % c for c in self.table.get_insert_columns(join=False,create=True)]))
+        self.file = open(self.table_name(), 'w')
+        self.output_file = csv.writer(self.file, dialect='excel', lineterminator='\n', quoting=csv.QUOTE_MINIMAL)
+        self.output_file.writerow(self.table.get_insert_columns(join=False,create=True))
+        self.table_names.append((self.file, self.table_name()))
 
     def disconnect(self):
         """Close the last file in the dataset"""
-        try:
-            self.output_file.close()
-        except:
-            # when disconnect is called by app.connect_wizard.ConfirmPage to
-            # confirm the connection, output_file doesn't exist yet, this is
-            # fine so just pass
-            pass
+        for output_tuple in self.table_names:
+            output_tuple[0].close()
 
     def execute(self, statement, commit=True):
         """Write a line to the output file"""
-        self.output_file.write('\n' + statement)
+        self.output_file.writerows(statement)
 
     def format_insert_value(self, value, datatype):
         """Formats a value for an insert statement"""
-        v = Engine.format_insert_value(self, value, datatype, escape=False)
+        v = Engine.format_insert_value(self, value, datatype, escape=False, processed=True)
         if v == 'null':
             return ""
         try:
@@ -85,14 +83,18 @@ class engine(Engine):
 
     def insert_statement(self, values):
         """Returns a comma delimited row of values"""
-
         if not hasattr(self, 'auto_column_number'):
             self.auto_column_number = 1
-        insert_stmt = ','.join([str(value) for value in values])
+
         if self.table.columns[0][1][0][3:] == 'auto':
-            insert_stmt = str(self.auto_column_number) + "," + insert_stmt
-            self.auto_column_number += 1
-        return insert_stmt
+            newrows = []
+            for rows in values:
+                insert_stmt = [self.auto_column_number] + rows
+                newrows.append(insert_stmt)
+                self.auto_column_number += 1
+            return newrows
+        else:
+            return values
 
     def table_exists(self, dbname, tablename):
         """Check to see if the data file currently exists"""
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -102,11 +102,16 @@ class Engine(object):
 
         total = self.table.record_id + real_line_length
         pos = 0
+        count_iter = 1
+        insert_limit = 200
+        current = 0
+        types = self.table.get_column_datatypes()
+        multiple_values = []
         for line in real_lines:
             if not self.table.fixed_width:
                 # This replaces end of line characters that exist in a single line
                 # eg. "one \nline has multiple end of lines\n"
-                line = line.replace('\n', '').strip()
+                line.replace('\n', '').strip()
             if line:
                 self.table.record_id += 1
 
@@ -119,12 +124,12 @@ class Engine(object):
                     else:
                         line += real_lines[pos+1]
                         real_lines.pop(pos+1)
+                    real_line_length -= 1
                     val_list = (self.table.extract_values(line))
                 pos += 1
 
                 linevalues = self.table.values_from_line(line)
 
-                types = self.table.get_column_datatypes()
                 # Build insert statement with the correct # of values
                 try:
                     cleanvalues = [self.format_insert_value(self.table.cleanup.function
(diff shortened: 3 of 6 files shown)


henrykironde:
I have skipped the test for extra line ending because it is failing. I will get that fixed later.

.travis.yml
```diff
@@ -5,6 +5,11 @@ python:
...
 addons:
   postgresql: "9.4"
+  apt:
+    sources:
+      - travis-ci/sqlite3
+    packages:
+      - sqlite3
 
 services:
   - postgresql
```
henrykironde: Travis uses the older version of sqlite3. This enables the new installation of a more current version than the default version

ethanwhite:
This PR also includes significant changesets for the `csv`, `json`, and `xml` engines. Are these all tied together or can they be separated into separate PRs?

henrykironde:
They are all tied together. All engines change the way the insert statement is created. 
For flat file system, the changes are in the respective engines.

ethanwhite:
OK, thanks. I'll try to take a look this evening but will have to see what my stamina is like.

engines/csvengine.py
```diff
@@ -53,27 +55,23 @@ def create_db(self):
...
         """Close the last file in the dataset"""
-        try:
-            self.output_file.close()
-        except:
-            # when disconnect is called by app.connect_wizard.ConfirmPage to
-            # confirm the connection, output_file doesn't exist yet, this is
-            # fine so just pass
-            pass
+        for output_file_i, file_name in self.table_names:
+            output_file_i.close()
```
ethanwhite: Since this is a separate function I think `output_file` should be fine as a variable name instead of `output_file_i`.

engines/csvengine.py
```diff
@@ -53,27 +55,23 @@ def create_db(self):
     def create_table(self):
         """Create the table by creating an empty csv file"""
         self.auto_column_number = 1
-        self.output_file = open(self.table_name(), "w")
-        self.output_file.write(
-            ','.join(['"%s"' % c for c in self.table.get_insert_columns(join=False,create=True)]))
+        self.file = open(self.table_name(),'w')
```
ethanwhite: Missing space after `,`.

engines/csvengine.py
```diff
@@ -53,27 +55,23 @@ def create_db(self):
...
+        self.output_file = csv.writer(self.file, dialect='excel', lineterminator='\n', quoting=csv.QUOTE_MINIMAL)
+        self.output_file.writerow(self.table.get_insert_columns(join=False,create=True))
+        self.table_names.append((self.file, self.table_name()))
```
ethanwhite: Does the 2nd item in the tuple get used anywhere? If not, what about just a list of file names called `file_names`?

lib/engine.py
```diff
@@ -583,7 +589,11 @@ def format_insert_value(self, value, datatype, escape=True):
                     strvalue = self.escape_double_quotes(strvalue)
                 if hasattr(self.table, "escape_single_quotes") and self.table.escape_single_quotes:
                     strvalue = self.escape_single_quotes(strvalue)
-            return "'" + strvalue + "'"
+                return "'" + strvalue + "'"
+            if processed:
+                return strvalue
+            else:
+                return "'" + strvalue + "'"
```
ethanwhite: The new `processed` code needs an explanation in the docstring. It's not intuitive to me what's being accomplished here.

lib/engine.py
```diff
@@ -102,11 +102,16 @@ def source_gen():
...
+        insert_limit = 200
+        current = 0
+        types = self.table.get_column_datatypes()
+        multiple_values = []
         for line in real_lines:
             if not self.table.fixed_width:
                 # This replaces end of line characters that exist in a single line
                 # eg. "one \nline has multiple end of lines\n"
-                line = line.replace('\n', '').strip()
+                line = line.strip()
```
ethanwhite: Can you walk me through why we're removing this here? Since we're pulling the `replace` the comment should also be removed since it is just explaining the `replace`.

lib/engine.py
```diff
@@ -135,32 +139,34 @@ def source_gen():
...
+                        if current > real_line_length:
+                            print(str(real_line_length) + " rows inserted into " + self.table_name() + ": ")
+                        else:
+                            print(str(current) + " rows inserted into " + self.table_name() + ": ")
+                    except:
+                        print(insert_stmt)
+                        raise
+                else:
+                    multiple_values.append(cleanvalues)
+                count_iter += 1
```
ethanwhite: This issues dates back before either this PR or @henrykironde, but `gen_source` is definitely too long and it makes reviewing changes to it hard work. Definitely something to target for refactoring sometime soon.

engines/csvengine.py
```diff
@@ -53,27 +55,23 @@ def create_db(self):
...
 
     def execute(self, statement, commit=True):
         """Write a line to the output file"""
-        self.output_file.write('\n' + statement)
+        self.output_file.writerows(statement)
 
     def format_insert_value(self, value, datatype):
         """Formats a value for an insert statement"""
-        v = Engine.format_insert_value(self, value, datatype, escape=False)
+        v = Engine.format_insert_value(self, value, datatype, escape=False, processed=True)
```
ethanwhite: If I understand things correctly, this is causing the values to be quoted if they are strings. Is that right? That seems like something that `csv.write` should be able to handle using `quoting=csv.QUOTE_NONNUMERIC`. I'm sure it's more complicated than I'm seeing, so could you walk me through the issue?

lib/engine.py
```diff
@@ -654,15 +664,21 @@ def insert_statement(self, values):
...
-        insert_stmt = insert_stmt.rstrip(", ") + ");"
-        n = 0
-        while len(values) < insert_stmt.count("%s"):
-            values.append(self.format_insert_value(None, types[n]))
-            n += 1
-        insert_stmt %= tuple([str(value) for value in values])
+        insert_stmt += " VALUES " 
+        vals = ""
+        for rows in values:
+            valk= rows
```
ethanwhite: Missing space. The meaning of `valk` isn't immediately clear to me.

lib/engine.py
```diff
@@ -654,15 +664,21 @@ def insert_statement(self, values):
...
-        for i in range(0, columncount):
-            insert_stmt += "%s, "
-        insert_stmt = insert_stmt.rstrip(", ") + ");"
-        n = 0
-        while len(values) < insert_stmt.count("%s"):
-            values.append(self.format_insert_value(None, types[n]))
-            n += 1
-        insert_stmt %= tuple([str(value) for value in values])
+        insert_stmt += " VALUES " 
+        vals = ""
```
ethanwhite: Unclear what this line is doing. Is `valk` below supposed to be `vals`? If so I don't see the need to set it to `""` here since it will get overwritten by `valk = rows`.

lib/engine.py
```diff
@@ -654,15 +664,21 @@ def insert_statement(self, values):
...
+            insert_stmt2 = " ("
+            for i in range(0, columncount):
+                insert_stmt2 += "%s, "
+            insert_stmt2 = insert_stmt2.rstrip(", ") + "),"
+            n = 0
+            while len(valk) < insert_stmt.count("%s"):
+                valk.append(self.format_insert_value(None, types[n]))
+                n += 1
+            insert_stmt2 %= tuple([str(value) for value in valk])
+            insert_stmt += insert_stmt2
```
ethanwhite: So, if I'm reading things right, `insert_stmt` at this point is:

`"INSERT INTO " + self.table_name() + " (" + columns + ")" + " VALUES "  + insert_stmt2`

What do you think about calling `insert_values`, deleting all of the `insert_stmt` code before this point, and the changing this line to:

`insert_stmt = "INSERT INTO {} ({}) VALUES {}".format(self.table_name(), columns, insert_values)`

My goal with this is to decrease the complexity and increase the readability.

lib/tools.py
```diff
@@ -1,4 +1,4 @@
-"""Data Retriever Tools
+"""EcoData Retriever Tools
```
ethanwhite: Looks like an accidental switch back to `EcoData`.

lib/tools.py
```diff
@@ -325,7 +325,7 @@ def sort_csv(filename):
     csv_reader_infile = csv.reader(input_file)
 
     # The first entry is the header line
-    infields = next(csv_reader_infile)
+    infields = ""
```
ethanwhite: The goal of the original line was to grab the first line and store it for later. As such, it seems like if we're just going to assign `infields` `""` then we can just get rid of the line and associated comment entirely. That said, from a first read it seems as if this just accomplished the same thing as the changes below now accomplish, which is using the first row as the header and not writing it. Is this change addressing one of the Python 2/3 csv differences? If so it should be documented with a comment. If not can you explain what the change is doing? Thanks.

test/test_integration.py
```diff
@@ -41,7 +41,7 @@
                  'script': "shortname: extra_newline\ntable: extra_newline, http://example.com/extra_newline.txt",
                  'expect_out': 'col1,col2,col3\n1,2,3\n'}
 
-tests = [simple_csv, autopk_csv, crosstab, autopk_crosstab, skip_csv, extra_newline]
+tests = [simple_csv, autopk_csv, crosstab, autopk_crosstab, skip_csv]
```
ethanwhite: Is the test giving failing when it shouldn't or did the new changes trigger a change in behavior? Either way I'd prefer to get this fixed as part of the PR rather than turning off the test.

ethanwhite:
Hey @henrykironde - looks really good overall. It's a complicated set of changes so I've left a bunch of questions about things I didn't understand.

engines/csvengine.py
goelakash: We should make `autopep8` as a recommended practice before submitting a PR.

henrykironde: :) True: I may add that in the pre-commit version

lib/tools.py
henrykironde: Yes we do not need it, It's not compatible with the p3 version of csv engine. I will test it to document the error produced.

lib/engine.py
henrykironde: added doc strings

lib/engine.py
henrykironde: I think an issue for this will be good

engines/csvengine.py
henrykironde: processed meaning that the Engine has the data processed, CSV module has quoting so is the data processed by the engine (processed =True)

lib/tools.py
ethanwhite: So the use of `next()` isn't compatible with Python 3? If so, that's fine, but then we can still get rid of `infields = ""` right?

henrykironde: Sure sorry about that.

lib/engine.py
```diff
@@ -654,15 +668,20 @@ def insert_statement(self, values):
...
-        insert_stmt += "VALUES ("
-        for i in range(0, columncount):
-            insert_stmt += "%s, "
-        insert_stmt = insert_stmt.rstrip(", ") + ");"
-        n = 0
-        while len(values) < insert_stmt.count("%s"):
-            values.append(self.format_insert_value(None, types[n]))
-            n += 1
-        insert_stmt %= tuple([str(value) for value in values])
+        insert_stmt += " VALUES "
```
henrykironde: For this part I will create a new issue that addresses the same issues  that you mentioned in the comment.

```
So, if I'm reading things right, insert_stmt at this point is:

"INSERT INTO " + self.table_name() + " (" + columns + ")" + " VALUES " + insert_stmt2

What do you think about calling insert_values, deleting all of the insert_stmt code before this point, and the changing this line to:

insert_stmt = "INSERT INTO {} ({}) VALUES {}".format(self.table_name(), columns, insert_values)

My goal with this is to decrease the complexity and increase the readability.
```

henrykironde: give me a minute to confirm this

ethanwhite:
OK, I think this PR is fine to go. Let me know when you've finished checking what you're looking at.

henrykironde:
@ethanwhite, Sorry if I took long than I expected, I will create an issue to refactor that part

ethanwhite:
No problem. So is this all ready to go in?

PR #641, opened 2016-09-11
Add scripts from EA archive 2000 to 2011

henrykironde:


henrykironde:
These scripts are still undergoing testing. Feel free to jump in and comment or test some

ethanwhite:
I haven't look at these more closely, but is there a reason we need the `.script` files?

henrykironde:
@ethanwhite I thought we needed them, will go ahead and remove them

ethanwhite:
Sorry for the confusion. Since these are new scripts the old versions of the retriever will all ignore them anyway, so nothing to worry about there.

goelakash:
@henrykironde Were these JSON scripts made using the CLI, or converted from .script files by the parser function?

henrykironde:
They were converted from .script files by the parser function.
They are not ready for merging because they still have unicode issues.

PR #656, opened 2016-09-18
Add value to replace in column

henrykironde:
replace # with "" in column names.

### DIFF
--- a/lib/table.py
+++ b/lib/table.py
@@ -58,6 +58,7 @@ class Table(object):
             ("&", "and"),
             ("\xb0", "degrees"),
             ("?", ""),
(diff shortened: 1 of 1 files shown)


lib/table.py
```diff
@@ -58,6 +58,7 @@ def clean_column_name(self, column_name):
             ("&", "and"),
             ("\xb0", "degrees"),
             ("?", ""),
+            ("#", ""),
```
henrykironde: this was discovered while testing the new scripts that we want to add

ethanwhite:
The things people do to data 😄. As soon as appveyor signs off I'll get this in.

PR #658, opened 2016-09-18
Add code of conduct

### LINKED ISSUE #654: Add a code of conduct
I keep planning on doing this and not getting around to it so I'm opening an issue. As usual we can probably just copy @ctb: https://github.com/dib-lab/khmer/blob/master/CODE_OF_CONDUCT.rst

henrykironde:
This code of conduct has been adapted from the
Contributor Covenant, version 1.4: http://contributor-covenant.org/version/1/4.

This commit also sets the encoding for for the conf.py module
and adds __future__ libraries required for python3
fixes #654

### DIFF
--- a/docs/code_of_conduct.rst
+++ b/docs/code_of_conduct.rst
@@ -1,0 +1,54 @@
+Contributor Code of Conduct
+===========================
+
+
+Our Pledge
+^^^^^^^^^^
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+Our Standards
+^^^^^^^^^^^^^
+
+Examples of behavior that contributes to creating a positive environment include:
+
+Using welcoming and inclusive language
+Being respectful of differing viewpoints and experiences
+Gracefully accepting constructive criticism
+Focusing on what is best for the community
+Showing empathy towards other community members
+Examples of unacceptable behavior by participants include:
+
+The use of sexualized language or imagery and unwelcome sexual attention or advances
+Trolling, insulting/derogatory comments, and personal or political attacks
+Public or private harassment
+Publishing others' private information, such as a physical or electronic address, without explicit permission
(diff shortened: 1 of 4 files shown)


docs/code_of_conduct.rst
```diff
@@ -0,0 +1,51 @@
...
+Focusing on what is best for the community
+Showing empathy towards other community members
+Examples of unacceptable behavior by participants include:
+
+The use of sexualized language or imagery and unwelcome sexual attention or advances
+Trolling, insulting/derogatory comments, and personal or political attacks
+Public or private harassment
+Publishing others' private information, such as a physical or electronic address, without explicit permission
+Other conduct which could reasonably be considered inappropriate in a professional setting
+Our Responsibilities
```
ethanwhite: This should be a header row.

docs/conf.py
```diff
@@ -1,3 +1,13 @@
+from __future__ import print_function
+from __future__ import absolute_import
+from imp import reload
+import sys
+# sys removes the setdefaultencoding method at startup; reload to get it back
+reload(sys)
+if hasattr(sys, 'setdefaultencoding'):
+    # set default encoding to latin-1 to decode source text
+    sys.setdefaultencoding('latin-1')
```
ethanwhite: Do the docs need the `sys` and default encoding related imports?

ethanwhite:
Oh, one more thing. From a little quick looking around it seems like it's pretty standard to capitalize this file name just like with LICENSE, CONTRIBUTING, etc. to indicate that it's this class of files that is somewhat separate from the software itself.

docs/conf.py
henrykironde: yes because they create scripts and have to read them in, so if a script as some characters non ASCII chars, then the encoding will help in that

ethanwhite: OK, thanks for the clarification.

ethanwhite:
I've also just seen it called `CONDUCT.md` as well.

henrykironde:
I will keep it as `CODE_OF_CONDUCT.rst` as in the original link since we want to include it in readthedocs

henrykironde:
This is ready for merge, unless I missed something

ethanwhite:
I think that the header row correction for "Our responsibilities" still needs to be take care of.

ethanwhite:
Thanks @henrykironde.

PR #661, opened 2016-10-02
Change to string before concatenating value with '"' delimiters in compile_json

### LINKED ISSUE #652: Wrap retriever_minimum_version value in str() in compile
On line 241 of `compile.py` the call to `value` should be changed to `str(value)`. It is natural for people to think of version numbers as numbers and if they accidentally enter `2.0` instead of `"2.0"` this causes a type failure.

### LINKED ISSUE #653: Stop bad scripts from causing errors
Currently when `compile_json` gets run if something goes wrong the retriever errors out even on commands not running the script (e.g., `retriever ls`). What it should do is ignore the bad script and possibly report back that there is an issue with the script but keep running normally otherwise.

goelakash:
Fixes #652.

### DIFF
--- a/lib/compile.py
+++ b/lib/compile.py
@@ -92,7 +92,7 @@ def compile_script(script_file):
                     tables[last_table][key] = "'" + str(e) + "'"
             else:
                 # general script attributes
-                values[key] = '"' + value + '"'
+                values[key] = '"' + str(value) + '"'
 
     if 'shortname' not in list(values.keys()):
         try:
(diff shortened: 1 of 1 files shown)


ethanwhite:
Thanks @goelakash. This should definitely take care of #652 and I'm going to go ahead and merge it to take care of that. I'd still like a more general solution to #653, so I'm going to remove #653 from your fix description so that it doesn't auto-close. I'll comment over there separately in a minute.

goelakash:
@ethanwhite By a general solution, do you mean improved exception handling in such a case ?

ethanwhite:
> @ethanwhite By a general solution, do you mean improved exception handling in such a case ?

Yes, exactly. A bad script shouldn't cause the user to experience an error. The should receive a warning related to that script and everything else should work normally. Basically it should work like it was working for you, but across platforms and systems which doesn't appear to be the case at the moment for some reason.

Issue #168, opened 2014-02-12
SQL INSERT statements use custom quoting rather than parameter binding 


hlapp:
Any reason you can't use parameter binding for binding values to SQL statements such as  `INSERT (...) VALUES (...)`? For most SQL engines, it's both faster and safer than using custom quoting and direct interpolation of values. It's certainly a lot less code (because the SQL engine does most of the type conversion and quoting).

goelakash:
Does this mean that we should use a database wrapper with ORM instead? Because currently we do have parameter binding (Using `?` in `INSERT` statements instead of python's `.format{}` formatting).

ethanwhite:
I think @hlapp is talking about this:
https://stackoverflow.com/questions/775296/python-mysql-parameterized-queries

We currently do the first version.

henrykironde:
I guess after we have created the table with the right data types we just pass all data to the database as strings  %s in the[ insert statement function](https://github.com/weecology/retriever/blob/master/lib/engine.py#L608) 
from a quick look I think we need to change the lines from here to give the database  the %s  and then values .https://github.com/weecology/retriever/blob/master/lib/engine.py#L617.


```
cursor.execute ("""
            INSERT INTO Songs (SongName, SongArtist, SongAlbum, SongGenre, SongLength, SongLocation)
            VALUES
                (%s, %s, %s, %s, %s, %s)

        """, (var1, var2, var3, var4, var5, var6))
```

Deborah-Digges:
@ethanwhite @henrykironde I'd like to take this up if nobody else is working on it.

henrykironde:
@Deborah-Digges, Please do no one is working on it.
:+1:

goelakash:
@ethanwhite That is completely dependent on the API of the database wrappers that we use (whether they take a single `str` argument or multiple). 
I am also of the opinion that since none of the data that retriever uses is hosted separately or sensitive in nature, we would not a 'safer' method to insert data in the DBs. But anything that @Deborah-Digges suggests or can improve upon should be valuable.

Deborah-Digges:
@goelakash @ethanwhite @henrykironde based on @goelakash's observation, do you think that these changes would add value to the project?

ethanwhite:
In general the database API wrappers all behave the roughly same way in Python. This makes it relatively easy to switch between database management systems and means that this should work. Examples for all three database APIs we interact with [1, 2, 3] show that the all take and prefer the:

[...]

PR #664, opened 2016-10-06
Refactor insert_statment for readability

### LINKED ISSUE #647: Refactor insert_statement in lib/engine
current status:

```
    def insert_statement(self, values):
        """Returns a SQL statement to insert a set of values."""
        columns = self.table.get_insert_columns()
        types = self.table.get_column_datatypes()
        columncount = len(self.table.get_insert_columns(join=False, create=False))
        insert_stmt = "INSERT INTO {} ({}) VALUES ".format(self.table_name(), columns)
        for rows in values:
            vals = rows
            insert_stmt2 = " ("
            for i in range(0, columncount):
                insert_stmt2 += "%s, "
            insert_stmt2 = insert_stmt2.rstrip(", ") + "),"
            n = 0
            while len(vals) < insert_stmt.count("%s"):
                vals.append(self.format_insert_value(None, types[n]))
                n += 1
            insert_stmt2 %= tuple([str(value) for value in vals])
            insert_stmt += insert_stmt2
        insert_stmt = insert_stmt.rstrip(", ") + ";"
        if self.debug:
            print(insert_stmt)
        return insert_stmt
```

This function could be refactor to improve readability

goelakash:
Fixes #647

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -670,18 +670,13 @@ class Engine(object):
         types = self.table.get_column_datatypes()
         columncount = len(self.table.get_insert_columns(join=False, create=False))
         insert_stmt = "INSERT INTO {} ({}) VALUES ".format(self.table_name(), columns)
-        for rows in values:
-            vals = rows
-            insert_stmt2 = " ("
-            for i in range(0, columncount):
-                insert_stmt2 += "%s, "
-            insert_stmt2 = insert_stmt2.rstrip(", ") + "),"
-            n = 0
-            while len(vals) < insert_stmt.count("%s"):
-                vals.append(self.format_insert_value(None, types[n]))
-                n += 1
-            insert_stmt2 %= tuple([str(value) for value in vals])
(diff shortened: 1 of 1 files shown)


goelakash:
@henrykironde Looks good?

henrykironde:
After reviewing the old version and the new version, both do not prevent sql injections since we are not letting sql do the processing of place holders.

henrykironde:
A good implementation would be as shown, http://bobby-tables.com/python.html
MySQL or PostgreSQL, use` %s`  and SQLite uses `?`

henrykironde:
We should also look at this #168

goelakash:
@henrykironde This is a bulk insert statement, so in order to pass parameters to this query as argument to a `cursor` object to insert data in the table, the data will have to be available there as well, which will need refactoring of `add_to_table` and `execute` functions as well.

henrykironde:
@goelakash, thanks for this PR

PR #684, opened 2016-10-14
Vertnet work

KristinaRiemer:


KristinaRiemer:
The fishes dataset has been removed from the script because the file at the link is corrupted and doesn't work.

henrykironde:
Thanks @KristinaRiemer, tested and works fine

henrykironde:
Note: This may fail on some systems, that's not an error from the script.

PR #686, opened 2016-10-17
Distribute and update scripts on install

### LINKED ISSUE #659: Distribute and update scripts on install
We want the software to be immediately functional on install without having to run any extra commands. Currently following install it is necessary to run `retriever update` to download all of the scripts.

Future releases should do two things:

1. Auto-populate `.retriever/scripts` with the scripts at the state of the release being installed.
2. Automatically run `retriever update` (or something equivalent) at the end of installation to update any scripts that have changed since the release.

This combination makes sure that:

1. The retriever is ready to go immediately following the installation
2. The user has the most up-to-date scripts immediately following installation
3. Installation is as fast as possible because instead of separately downloading all of the scripts only scripts that have changed since the last release need to be updated.

This issue is a detailed treatment of one of the issues in #318.

henrykironde:
fixes #659

### DIFF
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -1,18 +1,16 @@
 """Checks the repository for updates."""
 from __future__ import print_function
 from future import standard_library
-standard_library.install_aliases()
-
-
 import os
 import sys
-import urllib.request, urllib.parse, urllib.error
+import urllib.request
+import urllib.parse
+import urllib.error
 import imp
-from hashlib import md5
-from inspect import getsourcelines
-from threading import Thread
-from retriever import REPOSITORY, VERSION, MASTER_BRANCH, REPO_URL, SCRIPT_WRITE_PATH, HOME_DIR
+from pkg_resources import parse_version
+from retriever import REPOSITORY, SCRIPT_WRITE_PATH, HOME_DIR
 from retriever.lib.models import file_exists
+standard_library.install_aliases()
 
 global abort, executable_name
 abort = False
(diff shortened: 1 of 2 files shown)


ethanwhite:
Thanks @henrykironde. This works well, but leads to a slow install because all of the scripts are being manually downloaded from GitHub. This might be an OK starting point, but if we go this route we should add a print statement to let the user know that this is expected to be slow. Something like "Downloading dataset scripts. This may take a while...".

Ideally we'd have all of these files included in the tarball that `pip` downloads and just move them over (which will be much quicker). I know we can include them in the tarball and have them put in the same directory as the retriever's code (https://python-packaging.readthedocs.io/en/latest/non-code-files.html), but don't know how we would find them and move them over.

henrykironde:
`Ideally we'd have all of these files included in the tarball that pip downloads and just move them over (which will be much quicker)`

This sounds good. Will go ahead and change this

ethanwhite:
The downloading on install works great.

I tested the updating more generally by changing the version number in `MoM2003.json` to `0.9`, which should have caused it to update but it didn't. Unfortunately I can't see the issue from a quick read of the code.

Also, there is an extra layer of indenting throughout `check_for_updates()`.

lib/repository.py
```diff
@@ -29,123 +26,79 @@ def download_from_repository(filepath, newpath, repo=REPOSITORY):
...
+            if not file_exists(path_script_name):
+                download_from_repository("scripts/" + script_name,
+                                         os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
+
+            need_to_download = False
+            try:
+                file_object, pathname, desc = imp.find_module(''.join(script_name.split('.')[:-1]), [SCRIPT_WRITE_PATH])
+                new_module = imp.load_module(script_name, file_object, pathname, desc)
+                m = str(new_module.SCRIPT.version)
+                need_to_download = str(script_version) > m
```
ethanwhite: Should this comparison be done on the floating point values instead of the string versions to avoid issues like:

```
In [3]: "2.0" > "11.2"
Out[3]: True
```

henrykironde: I have just seen this but was rectifying that. 
I have used the version parser to take care of that.
It should be able to take care of this "2.0" > "11.2"

ethanwhite:
Nice use of `parse_version`. Definitely the best answer there. As soon as the test pass I'll get this in.

setup.py
```diff
@@ -25,14 +25,13 @@
...
 __version__ = 'v2.0.dev'
-with open("_version.py","w") as version_file:
+with open("_version.py", "w") as version_file:
     version_file.write("__version__ = " + "'" + __version__ + "'\n")
     version_file.close()
 
+
 def clean_version(v):
-    if v == 'master':
-        return '1.0.0'
```
ethanwhite: Yeah, that definitely seems like cruft. Good find.

Issue #668, opened 2016-10-08
Python 3  `bytes` vs `str` errors for most `.py` dataset scripts


ethanwhite:
E.g.,

```
ethan@gandalf:~/RetrieverCode (master *)$ retriever install csv Gentry
=> Installing Alwyn H. Gentry Forest Transect Dataset
Progress: 226 / 226 rows inserted into ./Gentry_sites.csv: 

write() argument must be str, not bytes
```

and a slightly different error

```
ethan@gandalf:~/RetrieverCode (master *)$ retriever install csv leda
=> Installing A database on the life history traits of the Northwest European flora
Downloading releasing_height.txt...
a bytes-like object is required, not 'str'
```

Datasets include:
- First error: baad, Gentry, BBS, BBS50, 
- Second error: leda, EA_petraitis2008

These datasets fail on all engines.

ethanwhite:
Errors in leda, EA_petraitis2008, and baad were all fixed by #691.

goelakash:
This issue with `str` has been fixed. I think we should track the new errors that occur with BBS and BBS50 installation.

ethanwhite:
Last I checked Gentry was still having issues as well. I'm happy with closing this and moving on to more specific issues.

goelakash:
Yes, my thoughts exactly.

henrykironde:
Don't close this issue, It's actually bigger than it looks, working on it and will give you the details. Almost there.

goelakash:
The title is misleading and actually follows a different error, which has been fixed by previous commits. We should file separate bug reports for different bugs.

henrykironde:
Okey I will close this over the weekend when I am cleaning all the issues. Thank you so much 
@goelakash

ethanwhite:
This is fixed now.

PR #691, opened 2016-10-21
Fix installation of LEDA dataset

goelakash:
Fixes some issues in #668

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -463,7 +463,7 @@ class Engine(object):
                 fileloc = self.format_filename(os.path.join(archivebase,
                                                             os.path.basename(filename)))
 
-                unzipped_file = open(fileloc, 'w')
+                unzipped_file = open(fileloc, 'wb')
                 for line in open_archive_file:
                     unzipped_file.write(line)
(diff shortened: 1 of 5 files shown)


ethanwhite:
Thanks for this @goelakash! For the three datasets with remaining issues do you think it's similar sorts of changes that we need to track down or something different?

Also, I forgot to mention this to you, but we don't merge our own PRs unless absolutely critical (e.g., reverting a commit that broke the build). Henry and I don't merge our own stuff either. This provides the opportunity for thorough review on bigger changes and quick sanity checks on simple changes like this one. In this case it would have ended up being useful because of a not yet documented change we've made to handling scripts that requires incrementing the included `version` value any time we update a script. No problem on this one and sorry I forgot to mention it earlier. Thanks again!

goelakash:
@ethanwhite I thought I'd let it be reviewed, but since it was a minor change for Python 2/3 compatibility, I merged it, which in hindsight, should have been avoided. I should have also paid attention to the new version number requirements though. I'll keep your suggestions in mind the next time.

Issue #694, opened 2016-10-22
Increment versions on BAAD, leda, EA_petraitis2008, BBS, and BBS50


ethanwhite:
#691 updated 5 scripts but didn't update their version numbers (since we haven't documented this requirement yet). The version numbers in these five scripts all need to be upped to 1.2.

@zhangcandrew - if you have a time this would be something fairly straightforward to help out with. The values for `version` in each of those 5 JSON files just need to be updated.

zhangcandrew:
Am I upping the versions in their respective .py files? I can't seem to find a .json for each of these in the scripts folder.

ethanwhite:
Yes, that's right. For datasets requiring more complicated cleanup and restructuring the information can't be captured in JSON files so the core script is the Python file.

PR #696, opened 2016-10-23
Changed version number from 1.0 to 1.2 on EA_petraitis, baad, bbs,

zhangcandrew:
bbs50, and leda scripts

changed the version number on these 5 scripts from 1.0 to 1.2. This is a
cleaner commit since my other ones were a bit off

address issues #694. For the final time


ethanwhite:
This is totally my fault, but those version numbers should actually be 1.1. In general, any time we change the script we will want to increment the minor version number by a single value.

ethanwhite:
Sorry I didn't get that right in the issue. Not sure what was going through my head.

Issue #699, opened 2016-10-31
new_json function requires values during script creation which are recommended on the scripts documentation


zhangcandrew:
Looking at the Scripts documentation, there are 5 fields which are strongly recommended when using the new_json retriever function. If I remember correctly though, when we tested out the new_json function, those fields were required. I will address these issues by making those values not required.

PR #700, opened 2016-10-31
new_json function no longer requires values which are recommended

zhangcandrew:
the new_json function's documentation states that certain values are only recommended, but when running the function, they are required. This pull request mends that by making the values optional, though highly recommended still .

addresses issue #699


henrykironde:
Thanks, @zhangcandrew

Issue #695, opened 2016-10-22
Document requirement to increment version when updating script


ethanwhite:
Any time a script gets updated we are going to increment the minor version number to track provenance and allow for reproducibility. This needs to be added to the developer documentation.

ethanwhite:
@zhangcandrew - this would be another good PR to work on as you get used to things. I sent you an invite to join the project as a collaborator. Once you accept you should be able to assign yourself to issues. If you're planning on tackling one please go ahead and do that so we know you're planning on taking care of it. Thanks.

zhangcandrew:
I can add this to the documentation pretty easily! Just to be clear though, am I changing the version number within the script like I did for #694, or is this purely within the documentation?

Also, I still seem unable to assign issues to myself, but I'm not sure why. I don't have the checkbox next to issues, so I am unable to give myself issues.

ethanwhite:
Awesome. Yes, this is just about a change to the documentation so that contributors know to increment the version numbers.

zhangcandrew:
Looking at the Scripts documentation, there are 5 fields which are strongly recommended when using the new_json retriever function. If I remember correctly though, when we tested out the new_json function, those fields were required. Should I get rid of the Strongly Recommended header and move those fields into "Required", or change datapackage.py so that their stored dictionary values don't rely on the clean_input function which requires a non-empty input?

henrykironde:
I think we could go ahead and change that to work for recommended instead of  mandatory

PR #701, opened 2016-11-02
Edited scripts.rst so that developers know to increment the version

zhangcandrew:
added a line in scripts.rst which lets developers know to edit the
script version every time they update a script.

addresses issue #695


henrykironde:
Nice :)

Issue #671, opened 2016-10-08
Data too long for column error in ` Palmer2007`


ethanwhite:
Python 3.

```
ethan@gandalf:~$ retriever install postgres Palmer2007
=> Installing North Carolina forest plants (Ecological Archives 2007)
Creating database Palmer2007...
Couldn't create database (schema "palmer2007" already exists
). Trying to continue anyway.
Creating table Palmer2007.trees...
[...]

PR #709, opened 2016-11-14
Use csv module for parsing lines

henrykironde:
Add data types to columns in FrayJorge and EltonTraits data sets
Add test for extra new lines.

Extra new lines are created when there are is a line end in a cell.
To view this scenario you can have a look at the EltonTraits
(http://esapubs.org/archive/ecol/E095/178/MamFuncDatSources.txt).

`
Ref_34     "Emmons L.H. & F. Feer. 1997. Neotropical Rain Forest Mammals, a Field Guide. 2da ed. The
University of Chicago Press. 281 pp"
`
The above example could be confused for two lines and becomes
complicated when more columns are invloved

When opened as txt the files rendering fails, however when opened
in csv handling applications, The right rendereing is obtained.

### DIFF
--- a/engines/postgres.py
+++ b/engines/postgres.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from builtins import str
 from retriever.lib.models import Engine, no_cleanup
 
@@ -70,6 +71,7 @@ class engine(Engine):
     def insert_data_from_file(self, filename):
         """Use PostgreSQL's "COPY FROM" statement to perform a bulk insert."""
         self.get_cursor()
+        self.set_engine_encoding()
         ct = len([True for c in self.table.columns if c[1][0][:3] == "ct-"]) != 0
         if (([self.table.cleanup.function, self.table.delimiter,
               self.table.header_rows] == [no_cleanup, ",", 1])
@@ -82,7 +84,7 @@ class engine(Engine):
 COPY """ + self.table_name() + " (" + columns + """)
 FROM '""" + filename.replace("\\", "\\\\") + """'
 WITH DELIMITER ','
-CSV HEADER"""
+CSV HEADER;"""
             try:
                 self.execute("BEGIN")
                 self.execute(statement)
@@ -121,6 +123,10 @@ CSV HEADER"""
             except:
                 pass
         return Engine.format_insert_value(self, value, datatype)
+
+    def set_engine_encoding(self):
+        self.execute("SET CLIENT_ENCODING TO '{0}';".format(
+            sys.getdefaultencoding()))
 
     def get_connection(self):
         """Gets the db connection."""
--- a/engines/sqlite.py
+++ b/engines/sqlite.py
@@ -1,6 +1,7 @@
 from builtins import range
 import os
-import platform
+import io
+import csv
 from retriever.lib.models import Engine, no_cleanup
 from retriever import DATA_DIR
 
@@ -36,61 +37,6 @@ class engine(Engine):
         """Escapes single quotes in the line"""
         return line.replace("'", "''")
 
-    def get_bulk_insert_statement(self):
-        """Get insert statement for bulk inserts
-
-        This places ?'s instead of the actual values so that executemany() can
-        operate as designed
-
-        """
-        columns = self.table.get_insert_columns()
-        types = self.table.get_column_datatypes()
-        columncount = len(self.table.get_insert_columns(False))
-        insert_stmt = "INSERT INTO " + self.table_name()
-        insert_stmt += " (" + columns + ")"
-        insert_stmt += " VALUES ("
-        for i in range(0, columncount):
-            insert_stmt += "?, "
-        insert_stmt = insert_stmt.rstrip(", ") + ");"
-        return insert_stmt
-
-    def insert_data_from_file(self, filename):
-        """Use executemany to perform a high speed bulk insert
-
-        Checks to see if a given file can be bulk inserted, and if so loads
-        it in chunks and inserts those chunks into the database using
-        executemany.
-
-        """
-        CHUNK_SIZE = 1000000
-        self.get_cursor()
-        ct = len([True for c in self.table.columns if c[1][0][:3] == "ct-"]) != 0
-        if (([self.table.cleanup.function, self.table.header_rows] == [no_cleanup, 1])
-            and not self.table.fixed_width
-            and not ct
-            and (not hasattr(self.table, "do_not_bulk_insert") or not self.table.do_not_bulk_insert)
-            ):
-            columns = self.table.get_insert_columns()
-            filename = os.path.abspath(filename)
-            try:
-                bulk_insert_statement = self.get_bulk_insert_statement()
-                line_endings = set(['\n', '\r', '\r\n'])
-                with open(filename, 'r') as data_file:
-                    data_chunk = data_file.readlines(CHUNK_SIZE)
-                    data_chunk = [line.rstrip('\r\n') for line in data_chunk if line not in line_endings]
-                    del(data_chunk[:self.table.header_rows])
-                    while data_chunk:
-                        data_chunk_split = [row.split(self.table.delimiter)
-                                            for row in data_chunk]
-                        self.cursor.executemany(bulk_insert_statement, data_chunk_split)
-                        data_chunk = data_file.readlines(CHUNK_SIZE)
-                self.connection.commit()
-            except:
-                self.connection.rollback()
-                return Engine.insert_data_from_file(self, filename)
-        else:
-            return Engine.insert_data_from_file(self, filename)
-
     def table_exists(self, dbname, tablename):
         """Determine if the table already exists in the database"""
         if not hasattr(self, 'existing_table_names'):
@@ -104,6 +50,9 @@ class engine(Engine):
     def to_csv(self):
         self.connection.text_factory = str
         Engine.to_csv(self)
+
+    def set_engine_encoding(self):
+        self.connection.text_factory = str
 
     def get_connection(self):
         """Gets the db connection."""
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -8,6 +8,7 @@ from builtins import input
 from builtins import zip
 from builtins import next
 from builtins import str
+
 import sys
 import os
 import getpass
@@ -15,6 +16,7 @@ import zipfile
 import gzip
 import tarfile
 import csv
+import re
 import io
 import time
 from urllib.request import urlretrieve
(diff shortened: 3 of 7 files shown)


henrykironde:
@zhangcandrew Andrew can you use this branch for testing scripts.

henrykironde:
@ethanwhite, @goelakash, @zhangcandrew, review when ever you can, Thank you

.travis.yml
```diff
@@ -34,6 +34,12 @@ before_script:
 
 after_success:
   - codecov
+  - retriever  install postgres FrayJorge -u postgres -d testdb
```
henrykironde: I added some lines to test direct retriever installation. 
We can remove them if not needed

engines/postgres.py
```diff
@@ -70,6 +71,7 @@ def escape_single_quotes(self, value):
     def insert_data_from_file(self, filename):
         """Use PostgreSQL's "COPY FROM" statement to perform a bulk insert."""
         self.get_cursor()
+        self.set_engine_encoding()
```
henrykironde: new encoding function to set encoding when installing in the database systems

engines/sqlite.py
```diff
@@ -41,17 +44,17 @@ def get_bulk_insert_statement(self):
 
         This places ?'s instead of the actual values so that executemany() can
         operate as designed
-
         """
+        self.set_engine_encoding()
         columns = self.table.get_insert_columns()
         types = self.table.get_column_datatypes()
-        columncount = len(self.table.get_insert_columns(False))
+        column_count = len(self.table.get_insert_columns(False))
```
henrykironde: only name refactor

lib/engine.py
```diff
@@ -66,70 +69,31 @@ def get_connection(self):
...
         for line in real_lines:
             if not self.table.fixed_width:
                 # This replaces end of line characters that exist in a single line
                 # eg. "one \nline has multiple end of lines\n"
-                line.replace('\n', '').strip()
+                line = line.replace('\n', '').strip()
             if line:
                 self.table.record_id += 1
-
-                # Check for single row distributed over multiple lines
```
henrykironde: We removed this part, It is handled by python's csv reader

lib/engine.py
```diff
@@ -159,9 +124,12 @@ def source_gen():
                         self.execute(insert_stmt, commit=False)
                         current += insert_limit
                         if current > real_line_length:
-                            prompt = "Progress: " + str(real_line_length) + " / " + str(total) + " rows inserted into " + self.table_name() + ": "
```
henrykironde: no change, only code clean up

lib/engine.py
```diff
@@ -159,9 +124,12 @@ def source_gen():
...
+                            prompt = "Progress: " + str(real_line_length) + " / " + str(
+                                total) + " rows inserted into " + self.table_name() + ": "
                         else:
-                            prompt = "Progress: " + str(current) + " / " + str(total) + " rows inserted into " + self.table_name() + ": "
```
henrykironde: no change, only code clean up

lib/engine.py
```diff
@@ -236,7 +228,8 @@ def auto_get_datatypes(self, pk, source, columns, column_values):
                         value = values[i]
 
                         if self.table.cleanup.function != no_cleanup:
-                            value = self.table.cleanup.function(value, self.table.cleanup.args)
```
henrykironde: no change, only code clean up

lib/engine.py
```diff
@@ -246,7 +239,7 @@ def auto_get_datatypes(self, pk, source, columns, column_values):
                                 try:
                                     value = int(value)
                                     if column_types[i][0] == 'int' and hasattr(self, 'max_int') and value > self.max_int:
-                                        column_types[i] = ['bigint',]
+                                        column_types[i] = ['bigint', ]
                                 except:
```
henrykironde: no change, only code clean up

lib/engine.py
```diff
@@ -623,7 +616,8 @@ def get_input(self):
                 else:
                     prompt = opt[1]
                     if opt[2]:
-                        prompt += " or press Enter for the default, %s" % opt[2]
```
henrykironde: no change, only code clean up

lib/engine.py
```diff
@@ -668,18 +661,23 @@ def insert_statement(self, values):
         """Returns a SQL statement to insert a set of values."""
         columns = self.table.get_insert_columns()
         types = self.table.get_column_datatypes()
-        columncount = len(self.table.get_insert_columns(join=False, create=False))
```
henrykironde: no change, only code clean up

test/test_integration.py
```diff
@@ -200,12 +200,25 @@
             'expect_out': 'b,c\n2,3\n5,6\n'}
```
henrykironde: The only change in this file is the addition of the extra_newline values,
The rest is code style.

test/test_regression_cli.py
```diff
@@ -12,10 +12,13 @@
 from retriever.lib.tools import getmd5
```
henrykironde: no change in the code logic but only style

lib/engine.py
```diff
@@ -66,70 +69,31 @@ def get_connection(self):
     def add_to_table(self, data_source):
         """This function adds data to a table from one or more lines specified
         in engine.table.source."""
+        lines = gen_from_source(data_source)
+
         if self.table.columns[-1][1][0][:3] == "ct-":
             # cross-tab data
-
-            lines = gen_from_source(data_source)
             real_lines = []
```
ethanwhite: Excellent call moving this into it's own function. I think that the creation of `real_lines` should be moved inside the function as well.

lib/engine.py
```diff
@@ -66,70 +69,31 @@ def get_connection(self):
...
         else:
-            # this function returns a generator that iterates over the lines in
-            # the source data
-            def source_gen():
-                return (line for line in gen_from_source(data_source)
-                        if line.strip('\n\r\t '))
-
-            # use one generator to compute the length of the input
-            real_lines, len_source = source_gen(), source_gen()
+            real_lines, len_source = tee(lines)
```
ethanwhite: So much better than doing this ourselves. Nice find!

lib/engine.py
```diff
@@ -173,6 +141,30 @@ def source_gen():
         print ("\n")
         self.connection.commit()
 
+    def get_ct_data(self, lines, real_lines):
+        """Creates cross tab data"""
```
ethanwhite: As noted about I think that `real_lines = []` should be the first line of this function rather than doing it outside the function and passing `real_lines` in.

lib/engine.py
```diff
@@ -668,18 +661,23 @@ def insert_statement(self, values):
...
+                row.append(self.format_insert_value(
+                    None, types[row_length + i]))
 
-            insert_stmt += " (" + ", ".join([str(val) for val in row]) +"), "
+            insert_stmt += " (" + ", ".join([str(val) for val in row]) + "), "
         insert_stmt = insert_stmt.rstrip(", ") + ";"
         if self.debug:
             print(insert_stmt)
+        # print (type(insert_stmt))
+        # exit()
```
ethanwhite: Debugging lines that should be removed?

lib/engine.py
```diff
@@ -698,16 +696,29 @@ def table_name(self, name=None, dbname=None):
...
     def warning(self, warning):
-        new_warning = Warning('%s:%s' % (self.script.shortname, self.table.name), warning)
+        new_warning = Warning('%s:%s' % (
+            self.script.shortname, self.table.name), warning)
         self.warnings.append(new_warning)
 
 
+def load_data(filename, delimiter="\t"):
+    reg = re.compile("\\r\\n|\n|\r")
+    with io.open(filename,  newline='', encoding='latin-1') as dataset_file:
```
ethanwhite: Should we be hard coding the `encoding` or doing this programmatically?

lib/engine.py
```diff
@@ -698,16 +696,29 @@ def table_name(self, name=None, dbname=None):
         return self.opts["table_name"].format(db=dbname, table=name)
 
     def warning(self, warning):
-        new_warning = Warning('%s:%s' % (self.script.shortname, self.table.name), warning)
+        new_warning = Warning('%s:%s' % (
+            self.script.shortname, self.table.name), warning)
         self.warnings.append(new_warning)
 
 
+def load_data(filename, delimiter="\t"):
```
ethanwhite: I don't understand what this function is doing. It would definitely benefit from a doc string describing what it is doing and why?

lib/engine.py
```diff
@@ -698,16 +696,29 @@ def table_name(self, name=None, dbname=None):
...
+
+
 def skip_rows(rows, source):
     """Skip over the header lines by reading them before processing."""
     lines = gen_from_source(source)
     for i in range(rows):
         next(lines)
-    return lines
+    # return lines
+    return source
```
ethanwhite: I'm confused as to what's going on here. Should we really be returning `source`? Can you explain the choice and document it in the docstring?

test/test_integration.py
```diff
@@ -200,12 +200,25 @@
...
+Ref_34	"Emmo The
+University pp"
+Ref_36	"Fidi and
+Nancy J.d
+Eulemur st,
+southeast"
+Ref_37	"Emmons H.
+A new genus.
+American Museum 1999"
+""",
```
ethanwhite: I'm guessing this was an actual failing line, but for the integration tests it would seem more useful to distill this to a simpler more generic case like:

```
"""col1,col2,col3
"ab", "cd", "e\nf"
"""

File: test/test_integration.py [hunk 21]
```diff
@@ -219,9 +232,12 @@
                         }\n
                     }\n
                     """,
-                 'expect_out': 'col1,col2,col3\n1,2,3\n'}
+                 'expect_out': 'ref_id,full_reference\nRef_34,Emmo The University pp'
```
ethanwhite: If following recommended change above this would become something like `'col1,col2,col3\n"ab","cd","ef"'`.

.travis.yml
henrykironde: There is really no important reason for these to get run. So I will remove them

engines/sqlite.py
```diff
@@ -60,34 +63,45 @@ def insert_data_from_file(self, filename):
...
                     while data_chunk:
-                        data_chunk_split = [row.split(self.table.delimiter)
-                                            for row in data_chunk]
-                        self.cursor.executemany(bulk_insert_statement, data_chunk_split)
+                        data_chunk_split = []
+                        for row in data_chunk:
+                            row_values = row.split(self.table.delimiter)
+                            if len(row_values) != column_len:
+                                row_values = [None] * \
+                                    (len(column_len) - len(row_values))
```
henrykironde: `So if the number of columns in the row doesn't make the number of columns we're converting everything to nulls?` yes based on some files the right most elements may lack values and  sometimes may not have any delimeters.
eg. 
```
a,b,c,d
1,2,3,4
2,4,
3,1,2,3
```
` will it break databases that have a primary key field?.`
At this point we should not be having any issues with any  but i will cross check .

ethanwhite: Is this the same behavior that we have for all the other engines? My naive assumption is that with data structured like this the missing fields are nulls, but the rest of the data is valid and should be included.

lib/engine.py
```diff
@@ -701,6 +678,37 @@ def warning(self, warning):
...
 
+    def load_data(self, filename):
+        """Generator uses the csv module to open a file and returns a list.
+
+        Opens a file and gets and sets the delimiter.
+        opens a file using the csv module and removes any extra line endings in the
+        fields
+        opens fixed width files and returns the list of values using the fixed width as the
+        delimiter
+        """
```
ethanwhite: Cleaned up docstring:

```
        """Generator returning lists of values from lines in a data file

        1. Works on both delimited (csv module) fixed width data (extract_fixed_width)
        2. Identifies the delimiter if not known
        3. Removes extra line endings

        """
```

lib/engine.py
```diff
@@ -701,6 +678,37 @@ def warning(self, warning):
...
+        reg = re.compile("\\r\\n|\n|\r")
+        if not self.table.delimiter:
+            with io.open(filename, newline='', encoding='latin-1') as dataset_file:
```
ethanwhite: Why do we use `''` for `newline` here? (I'm sure it's right I just don't understand it so want to know for the future.

lib/engine.py
```diff
@@ -701,6 +678,37 @@ def warning(self, warning):
...
+                self.auto_get_delimiter(dataset_file.readline())
+        with io.open(filename, newline='', encoding='latin-1') as dataset_file:
+            if self.table.fixed_width :
```
ethanwhite: Extra space before `:`

lib/engine.py
```diff
@@ -701,6 +678,37 @@ def warning(self, warning):
...
+                for row in dataset_file:
+                    yield self.extract_fixed_width(row)
+            else:
+                for row in csv.reader(dataset_file, delimiter=self.table.delimiter, escapechar="\\"):
+                    yield [reg.sub(" ", values) for values in row]
+
+    def extract_fixed_width(self, line):
+        """Splits a line based on the fixed width and returns a list of the values
+        """
```
ethanwhite: The trailing `"""` goes on the same line as the text for one line docstrings.

lib/engine.py
henrykironde: Newline controls how universal newline characters.
In this case ""  for Universal mode is used when empty "".
Otherwise we can set it to something different

ethanwhite: Great. Thanks for the explanation.

henrykironde:
This PR fixes #671 `retriever install postgres Palmer2007`  by adding 100 to the length of the auto detected field length.

henrykironde:
Already merged a copy

PR #718, opened 2016-11-28
Changed spacing in warning messages in tools.py to make them clearer

zhangcandrew:
Added line breaks after each sentence in the warning messages so that
they would be clearer for the user when displayed.


henrykironde:
@zhangcandrew, http://esapubs.org is up again. Could you commit amend and force a push for both PRs after cleaning up as suggested by Ethan, to have the test run before we merge.

PR #719, opened 2016-11-28
Added the Iris dataset to retriever

zhangcandrew:
Added an Iris.json file which adds functionality for the Iris dataset from information retrieved from UC Irvine.


zhangcandrew:
Sure! I changed the columns to their correct names and added the dataset citation, using and for now. I'm not sure why version.txt was rearranged when I ran python verson.py, but I'll get the issue up.

scripts/Iris.json
```diff
@@ -0,0 +1,88 @@
+{
+    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.
+
+",
```
ethanwhite: Looks like there are a couple of newlines that shouldn't be here. They prevent this from working for me (it doesn't show up in `retriever ls`).

henrykironde:
Thanks all

ethanwhite:
Ah, looks like @henrykironde merged this while I was writing. @zhangcandrew can you submit a follow up PR with the associated cleanup. Thanks.

scripts/Iris.json
henrykironde: So about that.  On my system, I had a copy which showed up and I mainly focused on `version.txt`.

ethanwhite: Interesting. I just confirmed that it doesn't show up on my machine if just the current JSON script is there. Ubuntu 16.04, Python 3.5.2. This is true both running from inside the repo and for outside the repo after running `retriever update` to get the current script that's now in `master`.

zhangcandrew:
I'm not sure how those extra lines popped up. Just created a new PR with the adjusted changes! 
Should the version of the new Iris script be 1.0 or 1.1?

ethanwhite:
Editors do funny things sometimes. This makes it worth spending a minute with the diff after you've made even seemingly minor changes to catch them as much as possible. Some will still slip though of course, and that's exactly what reviews like this can help with.

PR #722, opened 2016-12-01
Add progress to download_file

henrykironde:
When downloading large files, the user may think that the software
is not responding. This enables the user to know that the retriever
is busy fetching the files.


ethanwhite:
Nice addition!

PR #724, opened 2016-12-02
Follow up PR: Fixed formatting in Iris.json and removed redundant data

zhangcandrew:
Fixed some formatting in Iris.json. Also the script only downloads the
bezdekIris dataset now, which is the corrected dataset.

Should the version number for the script be 1.0 or 1.1?

### DIFF
--- a/scripts/Iris.json
+++ b/scripts/Iris.json
@@ -1,8 +1,6 @@
 {
-    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.
-
-",
-    "description": "Famous dataset from R. A. Fisher",
+    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.",
+    "description": "Famous dataset from R. A. Fisher. This dataset has been corrected. Information Source: Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.",
     "homepage": "http://mlr.cs.umass.edu/ml/datasets/Iris",
     "keywords": [
         "Taxon > Plants",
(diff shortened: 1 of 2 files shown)


scripts/Iris.json
```diff
@@ -15,45 +13,12 @@
...
-                "delimiter": ",",
-                "header_rows": 0
-            },
             "name": "bezdekIris",
             "schema": {
                 "fields": [
                     {
                         "name": "sepal_length",
-                        "type": "double"
+                        "type": "double"
```
ethanwhite: Remove extra whitespace at end of line.

scripts/Iris.json
```diff
@@ -15,45 +13,12 @@
...
-                ]
-            },
-            "url": "http://mlr.cs.umass.edu/ml/machine-learning-databases/iris/iris.data"
-        }, 
-        {
-            "dialect": {
-                "delimiter": ",",
-                "header_rows": 0
-            },
             "name": "bezdekIris",
```
ethanwhite: Let's just call this Iris now since it's just the correct version of the Iris data.

scripts/Iris.json
```diff
@@ -1,7 +1,5 @@
 {
-    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.
-
-",
+    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.",
     "description": "Famous dataset from R. A. Fisher",
```
ethanwhite: Let's add a note here pointing out that this is the corrected version of the data and cite the source for that information.

scripts/Iris.json
```diff
@@ -81,7 +46,6 @@
     "retriever_minimum_version": "2.0.dev",
     "title": "Iris Plants Database",
     "urls": {
-        "Iris": "http://mlr.cs.umass.edu/ml/machine-learning-databases/iris/iris.data",
         "bezdekIris": "http://mlr.cs.umass.edu/ml/machine-learning-databases/iris/bezdekIris.data"
```
ethanwhite: As the the `name` field I think this should just be `Iris` now.

scripts/Iris.json
```diff
@@ -1,8 +1,7 @@
 {
-    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.
-
-",
+    "citation": "R. A. Fisher. 1936. The Use of Multiple Measurements in Taxonomic Problems. and Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.",
     "description": "Famous dataset from R. A. Fisher",
+    "addendum": "This dataset has been corrected. Information Source: Asuncion, A. & Newman, D.J. (2007). UCI Machine Learning Repository [http://www.ics.uci.edu/~mlearn/MLRepository.html]. Irvine, CA: University of California, School of Information and Computer Science.",
```
zhangcandrew: We included an addendum parameter because in the python scripts, it was still being used, but we did not have it in compile.py

ethanwhite: Thanks @zhangcandrew. I'd completely forgotten that we had an `addendum` category. My initial thought on its continued use is that since there is no mechanism for exposing that information to the user that we're better off putting the information in the description. What do you think?

 `addendum` is also not in the datapackage spec.

zhangcandrew: Since I wasn't here for a large part of "addendum" 's use in the old scripts, I'm not sure if there were any unusual but useful reasons for using it. Off the top of my head, it seems good to have to show extra information or changes made to the data sets, but since it isn't exposed to the user currently, I agree with putting the information for this set in the description for now. I think it is useful for us to have the "addendum" section around though, since we can reference it to see any important changes made to scripts without cluttering up anything displayed to the user that doesn't need to be. It would be an optional parameter for the scripts.

ethanwhite: Sounds good. Go ahead and make the change and we'll get this in. I really like the idea of documenting changes to the scripts (or changes in the underlying data during the time of coverage by the retriever). Maybe open up an issue for further discussion.

ethanwhite:
Thanks Andrew!

PR #726, opened 2016-12-02
added a path variable for the home_dir in tools.py

zhangcandrew:
added a path variable which is set to the home_dir in tools.py so that
when the home_dir is displayed in the terminal, it is more standardized
across systems.


henrykironde:
Looks ready to merge. Thanks @zhangcandrew

ethanwhite:
👍

PR #725, opened 2016-12-02
Version.py now lists scripts in alphabetical order, ignoring case

zhangcandrew:
Version.py updates and returns version.txt with the scripts in
alphabetical order, regardless of case.

PR #727, opened 2016-12-02
Repair GWDD dataset

henrykironde:
Generates the files so that the engine can use the files for installing


henrykironde:
The script is fine but with current state of retriever, the script will fail.

ethanwhite:
Thanks @henrykironde. Once we merge #709 we can test and get this in. It will need to be rebased due to changes in #725.

henrykironde:
changes included in current PRs

PR #731, opened 2016-12-08
Adding the regression tests

henrykironde:


henrykironde:
Looks like we are winning :)
Two engines to go

PR #735, opened 2016-12-12
New json changes

zhangcandrew:
When we create a new json scripts, the version number is automatically added and
set to 1.0 now, and all table urls are compiled and placed in the urls
dictionary at the end of the scripts, just like in other normal scripts.
Null values previously when inputted in the new_json function would
cause an error during compilation if the null value was a string instead of an int. Null values can now handle both types. This was remedied in my changes to
compile.py


ethanwhite:
Looks like the tests are failing. One of the failures is reproducible locally and clearly related to this change. I don't know what the regressions tests are failing on CI, but they're still working locally for me.

zhangcandrew:
I added a change which allows null values to also be strings since I was having a problem compiling scripts with string type null values before. I think this is causing the problem for the add_dialect test and possibly the rest of the tests. I will keep looking into this.

lib/compile.py
```diff
@@ -30,8 +30,12 @@ def add_dialect(table_dict, table):
...
-            table_dict[
-                'cleanup'] = "Cleanup(correct_invalid_value, nulls=" + str(val) + ")"
+            try: 
+                table_dict[
+                    'cleanup'] = "Cleanup(correct_invalid_value, nulls=" + int(val) + ")"
+            except: 
+                table_dict[
+                    'cleanup'] = "Cleanup(correct_invalid_value, nulls=\"" + str(val) + "\")"
 
         elif key == "delimiter":
```
zhangcandrew: I looked into this and it seems like this isn't an easy fix, either test will have to be changed, or some other parts of how null values are handled, so for now I've set compile.py back to its original form, and I'll create an issue for this.

ethanwhite:
LGTM. In it goes. Thanks @zhangcandrew.

Issue #737, opened 2016-12-14
MOM2003 needs property "header_rows": 0


henrykironde:
Currently we are not downloading the file with the right number of rows.
This change will change the MD5 sums  used in testing. (should be 5731 instead of 5730)

zhangcandrew:
Hey Henry, I committed a change to one of my branches which fixes this issue, but I don't think the tests will pass, so I've held off on putting a merge request in for now.

henrykironde:
Nice @zhangcandrew, the Md5sum will definitely change. Go a head and create that branch only for mom and change the `Md5sum(92bf63eb5b36b777c600d0a95229222c)` with the new one

```
db_md5 = [
    # ('DelMoral2010', '0'),
    ('AvianBodySize', '79680888f7768474479e70c87cd36c9d'),
    ('MoM2003', '92bf63eb5b36b777c600d0a95229222c')
]
```

PR #738, opened 2016-12-16
Added header_rows value to Mom2003.json and changed regression test

zhangcandrew:
MoM200 now downloads the correct amount of rows.
Changed the regression test to reflect the change in the correct number
of downloaded rows.

addresses issue #737


henrykironde:
Thanks @zhangcandrew, looks great. This will go in now.

PR #740, opened 2016-12-21
Development features for 2.0 release

henrykironde:
1.Add CSV module to reprocess data. 
```
Previously the initial read of the raw data was done line by line reading plain text.
In cases where a newline occurred inside enclosing quotes this led to the line being
inappropriately split into two lines. When encountering values with internal new lines
the retriever now replaces them with a space.
This change required substantial restructuring because the csv module returns list of
strings instead of one string from reading the source directly.
```
2.Add Cross platform IO readers and writers on Posix and Windows
```
This adds cross platform support for reading writing files on multiple operating
systems running different python versions. 2 and 3
```
3.Enables Postgresql tests on Appveyor.
`We changed the Test password required to run local tests for Postgres to 'Password12!'`

### DIFF
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -18,7 +18,7 @@ import csv
 import io
 import time
 from urllib.request import urlretrieve
-from retriever import DATA_SEARCH_PATHS, DATA_WRITE_PATH
+from retriever import DATA_SEARCH_PATHS, DATA_WRITE_PATH, open_fr, open_fw, open_csvw
 from retriever.lib.cleanup import no_cleanup
 from retriever.lib.warning import Warning
 
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -77,7 +77,7 @@ class Engine(object):
         total = self.table.record_id + real_line_length
         pos = 0
         count_iter = 1
-        insert_limit = 200
+        insert_limit = 400
         current = 0
         types = self.table.get_column_datatypes()
         multiple_values = []
--- a/lib/engine.py
+++ b/lib/engine.py
@@ -15,6 +15,7 @@ import zipfile
 import gzip
 import tarfile
 import csv
+import re
 import io
 import time
 from urllib.request import urlretrieve
@@ -65,37 +66,12 @@ class Engine(object):
         """This function adds data to a table from one or more lines specified
         in engine.table.source."""
         if self.table.columns[-1][1][0][:3] == "ct-":
-            # cross-tab data
-
             lines = gen_from_source(data_source)
-            real_lines = []
-            for line in lines:
-                split_line = self.table.extract_values(line)
-                initial_cols = len(self.table.columns) - (3 if hasattr(self.table, "ct_names") else 2)
-                # add one if auto increment is not set to get the right initial columns
-                if not self.table.columns[0][1][0] == "pk-auto":
-                    initial_cols += 1
-                begin = split_line[:initial_cols]
-                rest = split_line[initial_cols:]
-                n = 0
-                for item in rest:
-                    if hasattr(self.table, "ct_names"):
-                        name = [self.table.ct_names[n]]
-                        n += 1
-                    else:
-                        name = []
-                    real_lines.append(
-                        self.table.combine_on_delimiter(begin + name + [item]))
-            real_line_length = len(real_lines)
+            # cross-tab data
+            real_line_length, real_lines = self.get_ct_data(lines)
         else:
-            # this function returns a generator that iterates over the lines in
-            # the source data
-            def source_gen():
-                return (line for line in gen_from_source(data_source)
-                        if line.strip('\n\r\t '))
-
-            # use one generator to compute the length of the input
-            real_lines, len_source = source_gen(), source_gen()
+            real_lines = gen_from_source(data_source)
+            len_source = gen_from_source(data_source)
             real_line_length = sum(1 for _ in len_source)
 
         total = self.table.record_id + real_line_length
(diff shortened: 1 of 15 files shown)


henrykironde:
I am testing this PR. If any one wants to run `try_install_all `, please use this file `https://raw.githubusercontent.com/henrykironde/retriever/try_install_all/try_install_all.py`
Otherwise I am testing all datasets.

`.py` scripts using open file will fail on python3 like Gentry and GWDD. PRs for these two are on the way.

Scripts using `insert_data_from_archive` will fail on python3.
Since some files have some corrupt data like vertnet and BBS, The python3 Zipfile module does not handle this. Python2 handles corrupt files and amends the data. These dataset will not be available on Python3 until I find another solution. (The zip will be downloaded well, but if you try to extract, python 3 errors. I have tried using io.ByteIO module to reload the data as a new byte object but still could not extract all the data)

henrykironde:
@ethanwhite This should be ready for review. The `.py` scripts using open statements will come in after

ethanwhite:
@henrykironde OK, great. I'll try to take a look today. Tomorrow at the latest.

lib/engine.py
```diff
@@ -211,52 +189,50 @@ def auto_create_table(self, table, url=None, filename=None, pk=None):
...
                                     column_types[i] = ['double', ]
                             if column_types[i][0] == 'double':
                                 try:
-                                    value = float(value)
-                                    if "e" in str(value) or ("." in str(value) and
-                                                             len(str(value).split(".")[1]) > 10):
-                                        column_types[i] = ["decimal", "50,30"]
+                                    val = float(val)
+                                    if "e" in str(val) or ("." in str(val) and len(str(val).split(".")[1]) > 10):
+                                        column_types[i] = ["decimal", "30,20"]
```
ethanwhite: Looks like this is a regression from e2e35d7 and should be set back to "50,30".

lib/engine.py
```diff
@@ -695,10 +661,62 @@ def table_name(self, name=None, dbname=None):
...
+        self.disconnect()
+
     def warning(self, warning):
         new_warning = Warning('%s:%s' % (self.script.shortname, self.table.name), warning)
         self.warnings.append(new_warning)
 
+    def load_data(self, filename):
+        """Generator returning lists of values from lines in a data file
+
+        1. Works on both delimited (csv module) fixed width data (extract_fixed_width)
```
ethanwhite: Add an "and" before "fixed width".

lib/engine.py
```diff
@@ -695,10 +661,62 @@ def table_name(self, name=None, dbname=None):
...
+        2. Identifies the delimiter if not known
+        3. Removes extra line endings
+
+        """
+        if not self.table.delimiter:
+            self.set_table_delimiter(filename)
+
+        reg = re.compile("\\r\\n|\n|\r")
+        dataset_file = open_fr(filename)
```
ethanwhite: Seems like settings my the delimiter and defining reg and only used in non-fixed-width data. Should the go in the else?

lib/tools.py
```diff
@@ -291,8 +232,16 @@ def getmd5(data, data_type='lines', mode='rb'):
...
-        lines = open(file_path, mode)
-        for line in lines:
+        # don't use open_fr to keep line endings consistent across OSs
+        if sys.version_info >= (3, 0, 0):
+            if os.name == 'nt':
+                input_file = io.open(file_path, 'r', encoding='ISO-8859-1')
+            else:
+                input_file = open(file_path, 'r', encoding='ISO-8859-1')
+        else:
+            input_file = io.open(file_path, encoding='latin-1')
```
ethanwhite: Make encoding consistent in the else.

lib/tools.py
```diff
@@ -371,7 +314,11 @@ def create_file(data, output='output_file'):
 def file_2string(input_file):
     """return file contents as a string"""
     input_file= os.path.normpath(input_file)
```
ethanwhite: Add space before =.

lib/tools.py
```diff
@@ -371,7 +314,11 @@ def create_file(data, output='output_file'):
...
-    with open(input_file, 'rU') as obs_out_file:
-        obs_out = obs_out_file.read()
+    if sys.version_info >= (3, 0, 0):
+        input = io.open(input_file, 'rU')
+    else:
+        input = io.open(input_file, encoding='latin-1')
```
ethanwhite: Use the ISO version for consistency.

test/test_integration.py
```diff
@@ -1,23 +1,36 @@
...
+reload(sys)
+if hasattr(sys, 'setdefaultencoding'):
+    sys.setdefaultencoding('latin-1')
 import pytest
 from retriever.lib.compile import compile_json
 from retriever import HOME_DIR, ENGINE_LIST
 from retriever.lib.tools import file_2string
 from retriever.lib.tools import create_file
 
+if os.name == "nt":
```
ethanwhite: Add a comment that we're doing this due to the different CI systems.

test/test_integration.py
```diff
@@ -200,12 +278,15 @@
             'expect_out': 'b,c\n2,3\n5,6\n'}
 
 extra_newline = {'name': 'extra_newline',
-                 'raw_data': 'col1,col2,col3\n1,2\n,3\n',
+                 'raw_data': """col1,col2,col3\nab,"e\nf",cd""",
```
ethanwhite: Can we use ' instead of """?

test/test_regression_cli.py
```diff
@@ -12,22 +12,27 @@
 from retriever.lib.tools import getmd5
 from retriever import ENGINE_LIST
 
+if os.name == "nt":
+    os_password = "Password12!"
+else:
+    os_password = ""
+
```
ethanwhite: Add same comment about CIs as above.

ethanwhite:
Awesome work on this one @henrykironde. I know it's been a bit of a grind, but it looks great. In it goes!

PR #742, opened 2017-01-05
Repair excel data sets

henrykironde:


henrykironde:
@ethanwhite, Could you have look at the data from Gentry and the script in general.

ethanwhite:
I've run the requested test code and uploaded it to chat.

`line` and `species_id` should both remain as `int`. `line` comes from the data as an `int` and we generate `species_id` as a unique integer ID.

PR #743, opened 2017-01-05
Repair BBS, BBS50Stops, Leda and fia

henrykironde:
Use cross platform open (open_fr, open_fw)


henrykironde:
Currently under tests

ethanwhite:
This changes all look good and testing this PR seems to have solved all of the issues in #668.

PR #753, opened 2017-01-11
Repair machine leaning dataset

henrykironde:
The dataset dialect values need to be redefined
For example header_rows is given as an digit not string

### DIFF
--- a/scripts/Abalone.json
+++ b/scripts/Abalone.json
@@ -7,7 +7,7 @@
     "resources": [
         {
             "dialect": {
-                "header_rows": "0"
+                "header_rows": 0
             },
             "name": "data",
             "schema": {
@@ -59,5 +59,5 @@
     "urls": {
         "data": "http://archive.ics.uci.edu/ml/machine-learning-databases/abalone/abalone.data"
     },
-    "version": 1.0
+    "version": 1.1
 }
--- a/scripts/Adult.json
+++ b/scripts/Adult.json
@@ -11,8 +11,8 @@
         {
             "dialect": {
                 "delimiter": ",",
-                "header_rows": "0", 
-                "nulls": "?"
+                "header_rows": 0,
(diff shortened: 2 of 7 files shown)


henrykironde:
These files where passing before I merged to master, I may have had some old copies of the scripts or we committed a wrong copy initially. Sorry about that

henrykironde:
@zhangcandrew, could you cross check these files. I think we also don't need to do any thing with the "?" values as we thought previously

ethanwhite:
Since we're making changes here I think a couple of changes to give more descriptive names would be good. The [spec](http://specs.frictionlessdata.io/data-packages/#required-fields) has now also changed to the following "This MUST be lower-case and contain only alphanumeric characters along with “.”, “_” or “-“ characters." 

I'd propose: abalone-age, adult-income, breast-cancer-wisconsin, car-evaluation, poker-hands, wine-quality. I'm happy to add a commit for this if we all agree with the change.

zhangcandrew:
@henrykironde I checked them again, I think we the addition of the brackets around the '?', the problem was fixed. When creating new jsons with the new_json function however, single value nulls are automatically stripped of the brackets, and that's where our previous problems arose, since I was making the new scripts through the new json function. This is something we don't do with delimiters, and I don't think we should do it with nulls either. I can go ahead and change that in the code real quick if it's ok. 

Other than that, I agree with the name changes that Ethan proposes.

zhangcandrew:
@henrykironde Yeah, we don't need to do anything anymore with the "?" s. I think we're good there. 

Also, I agree with the naming mechanic Ethan proposes.

ethanwhite:
OK, so it sounds like this is ready to go in. I'll go ahead and merge it and then handle the renaming as I deal with #755.

ethanwhite:
Just added a fix here to the wine quality data. One of the urls was missing the file name.

PR #757, opened 2017-01-11
Fix JSON and XML encoding issues on *nix Python 3

ethanwhite:
During disconnect files were being read without latin-1 encoding causing
an error before when they were read before disconnected.


ethanwhite:
Scripts that this fixed on Python 3 on Ubuntu include:
* PlantTaxonomy
* GWDD
* leaf_herbivory
* ArcticBreedBird

PR #759, opened 2017-01-12
Deleted Adult script

zhangcandrew:
Deleted the adult script due to the numerous problems caused by the
script itself.


ethanwhite:
Thanks @zhangcandrew.

PR #760, opened 2017-01-12
Correct auto get data type and add tests for char data type

henrykironde:


henrykironde:
This will solve the baad installation error.

PR #767, opened 2017-01-16
Fix for missing values in XML engine.

bendmorris:
This fixes an issue with Palmer2007 which currently fails with an IndexError. The issue is that the raw data file is missing values for "check" and "notes" in many rows.

The fix pads the row with empty values when a row is shorter than the header row. This happens to work for this dataset (where missing values are always at the end of the row) and is consistent with the sqlite engine's behavior, which fills in NULLs. However, there does seem to be some danger of masking legitimate data issues that the current strictness would help catch.

### DIFF
--- a/engines/xmlengine.py
+++ b/engines/xmlengine.py
@@ -108,15 +108,11 @@ class engine(Engine):
         else:
             newrows = values
 
-        open_tag = '\n<row>\n'
-        end_tag = '</row>'
-        xml_lines = []
-        for line_data in newrows:
-            write_data = ""
-            for i in range(len(keys)):
-                write_data += '    ' + '<' + str(keys[i]) + '>' + str(line_data[i]) + '</' + str(keys[i]) + '>' + "\n"
-            xml_lines.append(open_tag + write_data + end_tag)
+        xml_lines = ['\n<row>\n{}</row>'.format(self._format_single_row(keys, line_data)) for line_data in newrows]
         return xml_lines
+
+    def _format_single_row(self, keys, line_data):
+        return ''.join('    <{key}>{value}</{key}>\n'.format(key=key, value=value) for key, value in zip(keys, line_data))
 
     def table_exists(self, dbname, tablename):
         """Check to see if the data file currently exists"""
--- a/lib/table.py
+++ b/lib/table.py
@@ -128,6 +128,12 @@ class Table(object):
                 # too many values for columns; ignore
                 pass
             column += 1
+
+        # make sure we have enough values by padding with None
+        keys = self.get_insert_columns(join=False, create=False)
+        if len(linevalues) < len(keys):
+            linevalues.extend([None for _ in range(len(keys) - len(linevalues))])
+
         return linevalues
 
     def get_insert_columns(self, join=True, create=False):


henrykironde:
@bendmorris Thanks, if am not mistaken at that point we have the right number of values. I think the error with Palmer is the escape character "\"
A simple fix may be to include a custom clean function to remove a field with only one character "\"

bendmorris:
Here's an example of what I'm seeing in Oosting_Trees_1998.txt:

```python
['583', 'COFL', '   33', ' 113', '3.2', '1', '  4.9', '6', '  1', '  1', '  1', '  3', '  .', '  33', ' 2513', '   33', ' 113', '.', '  ', '']
['584', 'QUAL', '   36', ' 51', ' 44.9', '   1', '  .', '  2', '  1', '  4', '  1', '  1', '  .', '  36', ' 2451', '   36', ' 51', ' .']
```

The first line has two more values than the second (20 vs. 18).

henrykironde:
Could we have this changed in the add_to_table rather than in a single engine?

bendmorris:
AFAICT this issue only affects the XML engine. The default Engine.insert_statement used by the SQL engines accounts for this already:

```python
            # Add None with appropriate value type for empty cells
            for i in range(columncount - row_length):
                row.append(self.format_insert_value(None, types[row_length + i]))
```

I checked CSV and JSON as well.

henrykironde:
I will go through this, because that shouldn't happen for only one engine. I am thinking there is a bug hiding somewhere

ethanwhite:
@bendmorris - thanks for this. It's great to see you around. @henrykironde & I looked at this this morning and JSON also has issues. We thought the best solution would be to move this padding code into `lib.table.values_from_line()` so that it's automatically handled early in the process and covered for all engines. What do you think?

bendmorris:
Sounds good to me.

ethanwhite:
I'm going to re-open this as a reminder. @bendmorris - if you decide to to a crack at the more general route just assign this to yourself.

bendmorris:
Whoops, Github automatically closed it when I pushed again, and I can't reopen.

I moved my fix into values_from_line. While I was testing this, I noticed that the XML engine is fairly slow and spends most of its time in insert_statement. Replacing some inefficient string concatenation with a list comprehension, I'm seeing 4x speed increases.

Not sure what's up with the tests. Running locally I see a lot of similar failures, even on master.

bendmorris:
(Think I got everything figured out.)

henrykironde:
Thanks @bendmorris, It would be really cool if we could spot more areas where we can improve the time complexity. May be include an issue for such cases and anyone would take a look. For this case, I think the time is increased instead(let me know how you go about measuring the time) but we have a clean code base. great fix :+1:

bendmorris:
I identified insert_statement as a bottleneck with cProfile: `python -m cProfile -s cumtime __main__.py install xml plant-occur | less`

From master:

```
time retriever install xml plant-occur
=> Installing Oosting Natural Area (North Carolina) plant occurrence (Palmer et al. 2007)

Downloading species_codes.txt...
100%  0 seconds Progress: 224 / 224 rows inserted into ./plant_occur_oosting_species.xml totaling 224:


Downloading Oosting_pres2.txt...
100%  0 seconds Progress: 21253 / 21253 rows inserted into ./plant_occur_oosting_pres.xml totaling 21253:


Downloading Oosting_Trees_1998.txt...
100%  1 seconds Progress: 12973 / 12973 rows inserted into ./plant_occur_oosting_trees.xml totaling 12973:


Downloading Oosting_env.txt...
100%  0 seconds Progress: 289 / 289 rows inserted into ./plant_occur_oosting_env.xml totaling 289:

Done!

real	0m14.978s
user	0m9.938s
sys	0m0.331s
```

From 540a15736317fd26d23d4cd675de11aa38801c85 (immediately prior to my commit):

```
time retriever install xml plant-occur
=> Installing Oosting Natural Area (North Carolina) plant occurrence (Palmer et al. 2007)
Progress: 224 / 224 rows inserted into ./plant_occur_oosting_species.xml totaling 224:

Progress: 21253 / 21253 rows inserted into ./plant_occur_oosting_pres.xml totaling 21253:

Progress: 12973 / 12973 rows inserted into ./plant_occur_oosting_trees.xml totaling 12973:

Progress: 289 / 289 rows inserted into ./plant_occur_oosting_env.xml totaling 289:

Done!

real	0m36.191s
user	0m35.524s
sys	0m0.481s
```

bendmorris:
The fix in https://github.com/weecology/retriever/commit/540a15736317fd26d23d4cd675de11aa38801c85 results in some slowdown since it's calling `get_insert_columns` once more per row. This could probably be mitigated by caching the list of columns on the first call.

PR #768, opened 2017-01-16
[WIP]:repaire Npn data

henrykironde:
The xml file is processed, separating taxa plantae and animalia to differnet files
The data ranges from start_year to the current data of the end_year

fixes#693

### DIFF
--- a/scripts/npn.py
+++ b/scripts/npn.py
@@ -1,18 +1,21 @@
 #retriever
 """Retriever script for National Phenology Network data
+The dataset contains observation data retrieved from start date to current date
+date format YYYY-MM-DD
+Data having a value -9999 or "-9999" are considered 'null' or 'empty
+Data from the API is xml having both taxa(plantae and animalia)
 """
 from future import standard_library
+
 standard_library.install_aliases()
-from builtins import range
+from builtins import str
 
-import os
-import urllib.request, urllib.parse, urllib.error
-import zipfile
-from decimal import Decimal
+import time
+import xml.etree.ElementTree as ET
+
 from retriever.lib.templates import Script
-from retriever.lib.models import Table, Cleanup, no_cleanup, correct_invalid_value
-
-VERSION = '0.5'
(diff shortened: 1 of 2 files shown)


henrykironde:
Need to test it. currently the server seems to be down or blocking requests

scripts/npn.py
```diff
@@ -21,54 +27,132 @@ def __init__(self, **kwargs):
         self.name = "USA National Phenology Network"
         self.shortname = "NPN"
         self.retriever_minimum_version = '2.0.dev'
-        self.version = '1.0'
+        self.version = '1.1'
```
ethanwhite: Needs to go to 3 digit version -> `1.1.0`

scripts/npn.py
```diff
@@ -21,54 +27,132 @@ def __init__(self, **kwargs):
...
+                   ("intensity_value", ("char",)),
+                   ("abundance_value", ("char",))
+                   ]
+        while start_year <= end_year:
+            if start_year == end_year:
+                # get the data in the current year to current day (sometimes the recent day's data is empty)
+                data_url = base_url.format(startYear=start_year, endYear_date=str(time.strftime("%Y-%d-%m")),
+                                           request_src=request_src)
+            else:
+                data_url = base_url.format(startYear=start_year, endYear_date=str(start_year + 1) + "-01-01",
```
ethanwhite: `endYear_date=str(start_year) + "-12-31"` may avoid the potential for data on January 1st to get duplicated.

scripts/npn.py
ethanwhite: Actually, since this is such a major change I think this is the right situation to increment the major version number, so use `2.0.0` here instead.

henrykironde:
This data is very large and would need smaller iterations like a month or 3 months.
xml parse reads in entire source, hence running up on memory.

ethanwhite:
Should this be closed in favor of #827?

PR #763, opened 2017-01-13
Update scripts to match datapackage spec

ethanwhite:
Updates dataset files to match datapackage specification.

1. Makes names lowercase and more descriptive
2. Using `missingValues` instead of `nulls` in json files
3. Move to three level semantic versioning for script numbers
4. Use `keywords` instead of `tags` and switch to non-categorized keywords
5. General cleanup of names, descriptions, and titles

Fixes #518, #754, #755, #756.

PR #770, opened 2017-01-16
Clean up special characters

henrykironde:
Homeranges has fields with single special characters.
This fails on mysql engine.
Add escape character to list of values to clean as null


ethanwhite:
This is going to conflict with #763. My preference would be to get that PR reviewed and in and then rebase these more minor changes.

PR #771, opened 2017-01-16
Make script search paths work in correct order

ethanwhite:
The design is to look for scripts first in ., then in ./scripts, and finally
in ~/.retriever/scripts. The first script found should be loaded and any additional
copies of the same script ignored. The previously implementation looked good, but had
issues with aliasing, so even though copies of scripts from later directories weren't
explicitly added, they were replacing the correct scripts in the list. This separates
the check for whether the script has already been found to avoid the aliasing issue.

This design allows for easy development and customization. When developing modified
scripts in the repository are used by default. When customizing scripts added to
the working directory are used preferentially. When no other scripts appear the core
scripts are used.

This solves the painful development issue of needing to repeatedly run `retriever reset scripts` to avoid potentially testing outdated scripts in `~/.retriever` instead of the modified scripts in the repository.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -124,6 +124,7 @@ def to_str(object, object_encoding=sys.stdout):
 def MODULE_LIST(force_compile=False):
     """Load scripts from scripts directory and return list of modules."""
     modules = []
+    loaded_scripts = []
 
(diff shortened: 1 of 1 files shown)


henrykironde:
Will review it later.

Issue #764, opened 2017-01-14
Add test to check if all .json files are valid json


ethanwhite:

ethanwhite:
Closed by 71f4639. Thanks again @ShivamNegi!

PR #772, opened 2017-01-16
Test to check if all .json files are valid

ShivamNegi:
#764

### DIFF
--- a/test/test_json.py
+++ b/test/test_json.py
@@ -4,6 +4,5 @@ from glob import glob
 def test_valid_json():
 	"""Test if all json files are valid"""
 	for filename in glob("./scripts/*.json"):
-		if filename[-5:] == ".json":
-			with open(filename, "r") as fin:
-				assert json.load(fin)
+		with open(filename, "r") as fin:
+			assert json.load(fin)
--- a/test/test_json.py
+++ b/test/test_json.py
@@ -1,15 +1,9 @@
 import json
-import os
+from glob import glob
 
-HOMEDIR = os.path.expanduser('~')
-file_location = os.path.dirname(os.path.realpath(__file__))
-retriever_root_dir = os.path.abspath(os.path.join(file_location, os.pardir))
-
-def json_test_module():
+def test_valid_json():
 	"""Test if all json files are valid"""
-	os.chdir(retriever_root_dir + '/scripts')
-	for filename in os.listdir("."):
+	for filename in glob("./scripts/*.json"):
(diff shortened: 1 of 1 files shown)


test/test_json.py
```diff
@@ -0,0 +1,15 @@
+import json
+import os
+
+HOMEDIR = os.path.expanduser('~')
+file_location = os.path.dirname(os.path.realpath(__file__))
+retriever_root_dir = os.path.abspath(os.path.join(file_location, os.pardir))
```
ethanwhite: I don't think we need to explicitly track down the root directory here since the tests will always run from that directory. I think we should be able to remove these three lines and just work with `./scripts` below.

test/test_json.py
```diff
@@ -0,0 +1,15 @@
...
+
+def json_test_module():
+	"""Test if all json files are valid"""
+	os.chdir(retriever_root_dir + '/scripts')
```
ethanwhite: Instead of changing the directory, using `os.listdir`, and subsetting based on the last 5 characters of the filename I think using glob will be a little clearer. E.g.,:

```
from glob import glob
...
for filename in glob("./scripts/*.json"):
...
```

test/test_json.py
```diff
@@ -0,0 +1,15 @@
...
+	for filename in os.listdir("."):
+		if filename[-5:] == ".json":
+			file_content = ""
```
ethanwhite: It doesn't look like this line is currently being used and so it can be removed.

test/test_json.py
```diff
@@ -0,0 +1,15 @@
+import json
+import os
+
+HOMEDIR = os.path.expanduser('~')
+file_location = os.path.dirname(os.path.realpath(__file__))
+retriever_root_dir = os.path.abspath(os.path.join(file_location, os.pardir))
+
+def json_test_module():
```
ethanwhite: The word `test` needs to go first in the function name so that `pytest` will detect it. So either `test_json_module` or `test_valid_json` would be good names.

ethanwhite:
Thanks for the changes @ShivamNegi. They look great. There's one more thing which is that I don't think you need the `if filename[-5:] == ".json":` anymore since `glob("./scripts/*.json")` will only return file names ending in `.json`. Once you remove that line and the tests pass we can go ahead and get this merged.

ShivamNegi:
Thanks for the guidance.

ethanwhite:
Thanks for contributing! This looks great now. In it goes.

PR #776, opened 2017-01-20
Reset home-ranges version to proper value

ethanwhite:
A change to the home-ranges version number in version.txt crept in in
852830e even though now change to the script or its version was made.
This resets the version in version.txt to the proper value.

### DIFF
--- a/version.txt
+++ b/version.txt
@@ -20,7 +20,7 @@ forest-plots-michigan.json,1.1.0
 forest-plots-wghats.json,1.1.0
 fray-jorge-ecology.json,1.2.0
 gentry-forest-transects.py,1.3.0
-home-ranges.json,1.2.0
+home-ranges.json,1.1.0
 intertidal-abund-me.py,1.3.0
 iris.json,1.2.0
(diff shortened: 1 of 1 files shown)


henrykironde:
I do not have windows with me, could anyone help in testing home-ranges on mysql

ethanwhite:
> I do not have windows with me, could anyone help in testing home-ranges on mysql

It is failing, at least on Python 2, but I don't think that influences this PR which is just fixing a version mismatch between the current script and `version.txt`.

Here's the error:

```
$ retriever install mysql home-ranges -u root -p bats1e
=> Installing Database of Vertebrate Home Range Sizes - Tamburello et al., 2015
Creating database home_ranges...


Downloading Tamburelloetal_HomeRangeDatabase.csv...
100%  0 seconds Creating table home_ranges.ranges...
Progress: 569 / 569 rows inserted into home_ranges.ranges totaling 569:

Creating table home_ranges.home_ranges...
'charmap' codec can't encode character u'\x96' in position 33895: character maps to <undefined>
```

Issue #707, opened 2016-11-11
Add developer tools to assure script versions get update


ethanwhite:
Whenever a script is changed its `version` value needs to be incremented. This sort of thing is easy to forget and we should have some tools to help avoid missing it.

- [ ] Add a pre-commit hook that checks if there are changes in ./scripts/ and if so runs `version.py` and runs `git add version.txt`
- [ ] Update `version.py` to check scripts that have been changed (i.e., that have diffs) to see if the script version numbers have been changed, and if not, increment them.

This issue takes over the last two pieces of #318.

zhangcandrew:
Closing this for now after #849 was merged in! This can be reopened if any problems arise with the git hook.

PR #785, opened 2017-01-27
[WIP]:Git hooks for committing functionality

zhangcandrew:
I updated version.py to updated script versions if there are changes and
created a test.py pre-commit hook check scripts before committing. These
both need to be tested still on other platforms. To make the hook
work, change the name of test.py to pre-commit and put it in the
.git/hooks folder.

Addresses issue #707


zhangcandrew:
Hey @henrykironde if you have some time, could you review this code precommit hook code. I made some modifications to it based on the code you sent me. 

-version numbers are now objects called LooseVersions which can be more easily compared
-setup.py makes a check to see if a git hook exists currently (not sure what to do yet if one does)
-version files are checked against the last cached commit. 

Thanks

hooks/pre-commit
```diff
@@ -0,0 +1,59 @@
...
+    version_dict[key_values[0]] = key_values[1]
+
+
+def get_script_version(file_name):
+    file_name = os.path.split(file_name)[1]
+    return LooseVersion(version_dict[file_name])
+
+
+# check if script directory changes
+status, output = commands.getstatusoutput("git diff --name-only origin/script2 scripts")
```
zhangcandrew: This was unchanged from before. I was unsure what we were trying to do with this. Were we trying to see if the scripts had been moved to a different directory?

hooks/pre-commit
```diff
@@ -0,0 +1,59 @@
...
+
+# get_version_text_files_from_master:
+# discus if we need the version files in master of in last commit
+# this is tricky
+version_files = subprocess.check_output("git diff --cached --name-only", shell = True).decode("UTF-8")
+
+versions_not_changed =[]  # if script changes we expect to change the version
+versions_updated = True
+for files in scripts_changed:
+    if files in version_files:
```
henrykironde: I think we should change the logic here, If a script has changes, read in the file and use match = re.search() to get the version number.

update  the regex and make sure it gets the right version number 

mm is (a string having version number) the string below tries to cover all possible occurrences of version in both json and py files.


```
mm =  """
self.version = '1.1.0'
    "retriever": "True",
	"retriever_minimum_version": "2.0.0dev"
    "title": "Abalone Age and Size Data",
    "urls": {
        "data": "http://archive.ics.uci.edu/ml/machine-learning-databases/abalone/abalone.data"
    },
    "version": "1.2.1"
"""

```


```
>>> pattern = re.compile(r'[."]version["\s':=]+(\d+\.\d+.\d+)')
>>> match = re.search(pattern, mm)
>>> match.group(1)
'1.2.1'
```

hooks/pre-commit
```diff
@@ -0,0 +1,59 @@
...
+# check if script directory changes
+status, output = commands.getstatusoutput("git diff --name-only origin/script2 scripts")
+
+# get list of changed scripts
+scripts_changed = output.split("\n")
+
+# get_version_text_files_from_master:
+# discus if we need the version files in master of in last commit
+# this is tricky
+version_files = subprocess.check_output("git diff --cached --name-only", shell = True).decode("UTF-8")
```
henrykironde: add a commet "get the changes staged for the next commit relative to HEAD"

We shall have to test the results of `subprocess.check_output` when using `.decode("UTF-8")` on python3 and python2.

hooks/pre-commit
henrykironde: This should be `git diff --name-only upstream/master scripts`
We test out current branch with what is in the upstream/master.

We assume everyone will have upstream as the remote name value

```
upstream	https://github.com/weecology/retriever.git (fetch)
upstream	https://github.com/weecology/retriever.git (push)

```

We can also add a check for this value using `git remote -v ` and if it is not in the results , we could exit(1) and tell the user to run `git remote add upstream	https://github.com/weecology/retriever.git`

hooks/pre-commit
```diff
@@ -0,0 +1,59 @@
...
+import os
+import sys
+import urllib.request, urllib.parse, urllib.error
+from distutils.version import LooseVersion
+import re
+
+pattern = re.compile(r'script_version["\s:=]+(\d+\.\d+)')
+version_dict = {}
+versionfile = urllib.request.urlopen(
+    "https://raw.githubusercontent.com/henrykironde/retriever/1f18008838cecf7f8ac208d57cd2a578e50e1c2c/version.txt")
```
henrykironde: This should be the https://raw.githubusercontent.com/weecology/retriever/master/version.txt

henrykironde:
Hi @zhangcandrew, I have made some comments, and for the set up file, Please update the current one. Take a look at the current and incorporate those changes into it.

hooks/pre-commit
zhangcandrew: I've updated the regex to match the new version numbers. It seems like the new logic you want to implement is already in the python file though, with the regex search and match. Maybe I'm missing something though.

zhangcandrew:
Thanks Henry, I've made the appropriate changes, and commented on some other logic.

hooks/pre-commit
```diff
@@ -0,0 +1,62 @@
...
+
+
+def get_script_version(file_name):
+    file_name = os.path.split(file_name)[1]
+    return LooseVersion(version_dict[file_name])
+
+
+# check if script directory changes
+remotes = subprocess.check_output("git remote -v")
+if("upstream    https://github.com/weecology/retriever.git" in remotes):
```
henrykironde: remove "( )" and change the printed string
```
if "upstream    https://github.com/weecology/retriever.git" in remotes:
    status, output = commands.getstatusoutput("git diff --name-only upstream/master scripts")
else:
    print("Please first add upstream as a remote.\n"
          "git remote add upstream https://github.com/weecology/retriever.git")
    exit(1)
```

hooks/pre-commit
```diff
@@ -0,0 +1,62 @@
...
+    status, output = commands.getstatusoutput("git diff --name-only upstream/master scripts")
+else:
+    print("Please add weecology/retriever as a remote. \n git remote add upstream https://github.com/weecology/retriever.git")
+    exit(1)
+# get list of changed scripts
+scripts_changed = output.split("\n")
+
+# get the changes staged for the next commit relative to HEAD
```
henrykironde: Remove # Still needs testing on python3 and python2
Rename version_files to staged_scripts
```
scripts_changed = output.split("\n")

# get the changes staged for the next commit relative to HEAD
staged_scripts = subprocess.check_output("git diff --cached --name-only", shell = True).decode("UTF-8")

versions_not_changed =[]  # if script changes we expect to change the version
versions_updated = True
for files in scripts_changed:
    if files in staged_scripts:
```

setup.py
```diff
@@ -4,10 +4,37 @@
...
+import subprocess
 
 current_platform = platform.system().lower()
 extra_includes = []
+if current_platform == "darwin":
+  if(subprocess.check_output("[ ! -f .git/hooks/pre-commit ] && echo \"Not Found\"") == "Not Found"):
+    os.system("chmod +x hooks/pre-commit")
+    os.system("cp -p hooks/pre-commit .git/hooks/pre-commit")
+    # os.system("chmod +x hooks/pre-commit")
+    # os.system("ln -s hooks/pre-commit .git/hooks/pre-commit")
```
henrykironde: Remove these lines 
 
```
+    # os.system("chmod +x hooks/pre-commit")
 +    # os.system("ln -s hooks/pre-commit .git/hooks/pre-commit")
```

This set up file is old .

henrykironde:
Everything looks fine for the precommit file if those changes are made, we need to use the latest `setup.py` file.

zhangcandrew:
Just fixed it up Henry! I think it's using the latest setup.py file. I pulled from upstream and added it to the branch, then made the appropriate changes.

henrykironde:
I think it will be nice to move the changes to a new branch and create a new PR. 
Don't worry about my commits earlier.

henrykironde:
And I am not sure what these lines are doing.

```
else:
# os.system("ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit")
```

zhangcandrew:
That line was just in case the developer already had git hooks in place. We create a symbolic link from our hook to theirs instead of overwriting their file so that they can keep their own git hooks along with the ones we have for them. I'm not sure if it's too useful, but I added it as a precaution

PR #790, opened 2017-02-10
Bump version in preparation for next release

ethanwhite:


ethanwhite:
This should probably go ahead and be merged now that #794 is in. It doesn't commit us to not adding any more commits, but it seems like we're basically at that point now anyway.

PR #794, opened 2017-02-17
Use pyinstaller for creating exe for windows and app for mac

henrykironde:
Switch from Py2exe, Reason; not enough documentation
for the tool when it comes to py2to3 support

Create a spec file from the source using pyinstaller with __main__.py
Renamed the __main__.spec(output) to pyinstaller.spec

On mac options used
	pyinstaller  --onefile --icon=icon.ico  --windowed
for windows
	pyinstaller  --windowed is enough for creating a default spec file

Win32com is not available on unix machines so the import failure reported
will not affect the application(--log-level ERROR removes warnings)

Note: we still use the setup.py option "py2app" to build the debian package

### DIFF
--- a/build_mac
+++ b/build_mac
@@ -1,13 +1,7 @@
 export DYLD_LIBRARY_PATH=/usr/local/mysql/lib/
 sudo rm -rf mac
+sudo mkdir mac
 sudo python setup.py install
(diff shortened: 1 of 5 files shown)


ethanwhite:
Awesome work @henrykironde!

PR #799, opened 2017-02-21
Rename the files using python name convention

henrykironde:
Convert - to _ in both the file names and script name
Update the test files


ethanwhite:
I definitely agree with the switch to `_` in filenames. I'm conflicted about the use for the `name` attribute. Keeping it consistent with the filename is nice, but it's also generally inconsistent with naming in other package managers. apt primarily uses `-`, as does homebrew (see https://github.com/Homebrew/homebrew-core/tree/master/Formula).

I guess I lean towards `-` for `name` and `_` for the filenames, but I'm open to use `_` for both.

What do other folks think?

henrykironde:
I am okey with changing the names back to `-`. I will wait for opinions.

PR #803, opened 2017-02-23
Add default command

henrykironde:
This enables users to see the default values for each engine


henrykironde:
This is going to be used in the docs to explain the commands better to users

ethanwhite:
LGTM. :+1:

Issue #134, opened 2013-12-05
Allow overriding the use of locally cached data


ethanwhite:
By default if the Retriever has already downloaded a dataset it will use a locally cached version of the files instead of redownloading them. In some cases this may not be desirable. For example, if the newest year of Breeding Bird Survey data has been released the user may want to rebuild the database with the new data.

Solution: add a `--not-cached` that deletes the locally cached data before proceeding.

ShivamNegi:
I am working on this issue.

ShivamNegi:
I want some help here. I have updated the ```./lib/templates.py``` to incorporate ```--not-cached``` argument. But there are still many scripts which are not using the ```class BasicTextTemplate``` for downloading. These scripts have their own download functions. I can't seem to find any other way than to go in each file and change its download function. Any thought on how to go about incorporating the ```--not-cached``` functionality?

ethanwhite:
Implemented in bc0c2318a854de6c844bace6a28807f54328eae8. Thanks @ShivamNegi!

ShivamNegi:
@ethanwhite after bc0c231. Running ```install``` output is shown below.
```
➜  retriever git:(vertnet-dataset) ✗ retriever install csv vertnet
=> Installing vertnet:
download() got an unexpected keyword argument 'use_cache'
Done!
```
This can be fixed by adding a ```use_cache``` default argument in the ```download``` function of ```vertnet```.

ethanwhite:
Ah yes, thanks for catching that. This will actually cause issues for a number of files:

```
scripts/la_selva_trees.py:    def download(self, engine=None, debug=False):
scripts/gentry_forest_transects.py:    def download(self, engine=None, debug=False):
scripts/breed_bird_survey.py:    def download(self, engine=None, debug=False):
scripts/forest_inventory_analysis.py:    def download(self, engine=None, debug=False):
scripts/prism_climate.py:    def download(self, engine=None, debug=False):
scripts/breed_bird_survey_50stop.py:    def download(self, engine=None, debug=False):
scripts/biomass_allometry_db.py:    def download(self, engine=None, debug=False):
scripts/bioclim.py:    def download(self, engine=None, debug=False):
scripts/plant_life_hist_eu.py:    def download(self, engine=None, debug=False):
scripts/vertnet.py:    def download(self, engine=None, debug=False):
scripts/wood_density.py:    def download(self, engine=None, debug=False):
[...]

PR #811, opened 2017-03-01
Fixes 683: restructure project by moving core codes in sub-directory

kunalpal:

PR #810, opened 2017-03-01
Removed default encoding in refernce to #716

jainamritanshu:
Sir I have removed the hard coded assignments for encoding and added a field of **encoding**. I haven't edited the existing scripts for the existing data. Should I do them manually? Kindly review the code and tell me any changes needed for the code.
@ethanwhite @henrykironde

PR #817, opened 2017-03-03
Option to clear local cache

ShivamNegi:
fix for #134 
```download_file``` function in  ```./lib/engine.py```  is being called twice. So this particular fix downloads the data file twice when there is no prior data file present and we run ```retriever install --not-cached csv iris```.
For the general case when the data file is already present and we need install by clearing local cache it works correctly.

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -27,7 +27,6 @@ from retriever.lib.tools import choose_engine, name_matches, reset_retriever
 from retriever.lib.get_opts import parser
 from retriever.lib.datapackage import create_json, edit_json
 
-
 def main():
     """This function launches the Data Retriever."""
     if len(sys.argv) == 1:
@@ -173,6 +172,11 @@ def main():
             debug = False
             sys.tracebacklimit = 0
 
+        if hasattr(args, 'debug') and args.not_cached:
+            use_cache = False
+        else:
+            use_cache = True
+
         if args.dataset is not None:
             scripts = name_matches(script_list, args.dataset)
         else:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ShivamNegi** [comment]
@henrykironde @ethanwhite travis CI is failing tests over piece of code I have not altered could you please look into it.

**Comment by ethanwhite** [comment]
Thanks for the contribution.

`insert_data_from_url` is overridden in `engines/download_only.py`, so you need to add the new `cache=True` there as well. I suspect that will take care of things based on the error message.

**Comment by ShivamNegi** [comment]
I just saw ```forest_inventory_analysis.py```, ```breed_bird_survey.py``` and ```npn.py``` have their own specific download functions. I am updating the PR.

File: __main__.py [hunk 1]
```diff
@@ -173,6 +172,11 @@ def main():
             debug = False
             sys.tracebacklimit = 0
 
+        if hasattr(args, 'not_cached') and args.not_cached:
```
henrykironde: Do we need the two checks here.

ShivamNegi: I don't think so the ```hasattr(args, 'not_cached')``` is required.
Though it has been added while checking for the ```--debug``` argument in ```__main__.py```

henrykironde: I think argparse takes care of all that. Have it cleaned and ready for review

lib/get_opts.py
```diff
@@ -38,6 +38,7 @@
 reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections', choices=['all', 'scripts', 'data', 'connections'])
 install_parser.add_argument('--compile', help='force re-compile of script before downloading', action='store_true')
 install_parser.add_argument('--debug', help='run in debug mode', action='store_true')
+install_parser.add_argument('--not-cached', help='clears local cache before installation', action='store_true')
```
ethanwhite: Technically this implementation overwrites the local cache, so maybe something like `overwrites local cache of raw data`

henrykironde:
@ShivamNegi, Thanks for the great work, we have this merged in locally to avoid conflicts by #811  and #810

Issue #792, opened 2017-02-14
Incorporate the PREDICTS database release


dmcglinn:
This is an impressive global database on species richness and abundance - it seems well suited for the `retriever`

http://data.nhm.ac.uk/dataset/902f084d-ce3f-429f-a6a5-23162c73fdf7

ethanwhite:
Totally agree.

I tried to do this right when it came out but there was no direct download link. I've been corresponding with the folks in charge and they provided this link:
http://data.nhm.ac.uk/dataset/the-2016-release-of-the-predicts-database/resource/78dac1a9-6aa0-4dcb-9750-50df04f8ca6e/download

We are allowed to use it for the retriever, but we're not supposed to advertise it more broadly.

dmcglinn:
ah yea I see now that they want your email address to download the files. Nice job getting that direct link!

jainamritanshu:
@ethanwhite I would like to proceed on this. Shall I?

henrykironde:
@jainamritanshu, go for it

jainamritanshu:
@ethanwhite @henrykironde Sorry if I am missing something. I have created the json script of PREDICTS but the python script has not been created. I saw the comment in `def compile_json`(file `lib/compile.py`) that the command to do so is `retriever create_json <script name>`. When I do so, the cli tells me that it is an invalid argument. The exact error I got -
`retriever: error: argument command: invalid choice: 'create_json'`. Please help?

ShivamNegi:
I think that the argument you are searching for here is ```new_json```. You can view all the options using ```retriever -h```.

jainamritanshu:
@ShivamNegi I made the .json file using `new_json`. As I stated

> I have created the json script of PREDICTS but the python script has not been created.

But the .py file script wasn't created and as you can see in the screenshot -
![screenshot from 2017-03-03 21-22-35](https://cloud.githubusercontent.com/assets/15127027/23557865/931d4344-0057-11e7-98e7-1921bc68f471.png)

I saw the comment that stated that I need to execute the command `retriever create_json <script name>`. So I got confused. Thanks anyways. :)

henrykironde:
Sorry, we forgot to change the docs. You could PR the changes to correct that.

jainamritanshu:
@henrykironde Will do that :)

PR #821, opened 2017-03-03
Made script for PREDICTS Database in reference to #792

jainamritanshu:
@ethanwhite @henrykironde I made .json script in order to incorporate PREDICTS Database new release


henrykironde:
Nice @jainamritanshu, since this data is in a zipped folder we shall create a python file instead of the json file.  
An example in shown in script [biomass_allometry_db](https://github.com/weecology/retriever/blob/master/scripts/biomass_allometry_db.py) 



citation:
 ```
Lawrence N Hudson; Tim Newbold; Sara Contu; Samantha L L Hill et al. (2016). Dataset: The 2016 release of the PREDICTS database. http://dx.doi.org/10.5519/0066354
Retrieved: 21:39 03 Mar 2017 (GMT)
```
Once you download from archive `engine.download_files_from_archive` the csv file will be extracted.
we create a table based on the fields 
check [vertnet script](https://github.com/weecology/retriever/blob/master/scripts/vertnet.py#L42) on how to provide the fields 

Then use `engine.insert_data_from_file`

let me know in case you need clarification.

**Comment by jainamritanshu** [comment]
@henrykironde I got it. Thank you so much for the clarification. Will commit the changes asap. :)

**Comment by jainamritanshu** [comment]
@henrykironde Sorry for the delayed action on this PR. I have made a python script for incorporating the PREDICTS Database. Kindly go through the code and suggest any changes required.

File: scripts/predicts.py [hunk 1]
```diff
@@ -0,0 +1,109 @@
...
+                                ("Species", ("char")),
+                                ("Best_guess_binomial", ("char")),
+                                ("Higher_taxa", ("char")),
+                                ("Higher_taxon", ("char")),
+                                ("Measurement", ("float")),
+                                ("Effort_corrected_measurement", ("float"))]
+
+            engine.table = table
+            if not os.path.isfile(engine.format_filename(filename)):
+                engine.download_files_from_archive(self.urls[tablename], [filename], filetype="zip", archivename="predicts_latest_" + str(tablename))
```
ethanwhite: Since we only have one table we can keep this simple:

`"predicts_latest_" + str(tablename)"` -> `"predicts_data"`

scripts/predicts.py
```diff
@@ -0,0 +1,109 @@
...
+            engine.create_table()
+            engine.insert_data_from_file(engine.format_filename(str(filename)))
+
+SCRIPT = main()
```
ethanwhite: Add a new line at the end of the file.

scripts/predicts.py
ethanwhite: I think we should only need one `download_files_from_archive` call. This second one is probably better, but it should either be this one or the one above unless I'm missing something.

jainamritanshu:
@ethanwhite I have made a couple of changes. Kindly give it a read.

ethanwhite:
Thanks @jainamritanshu - I've read through this and it looks great.

This won't run right now since it's dependent on #822 and encounters issues with the fix @ShivamNegi is working on in relation to #835, so I'll wait for those two to come in so we can test this properly.

The is one thing left to do on your end, which is to run `python version.py` and add the updated `version.txt` to this commit.

ethanwhite:
#835 is now fixed, so just waiting on #822 to go in and for `version.txt` to be updated.

jainamritanshu:
Sure will update version.txt asap! :smile:

henrykironde:
Looks good, Will be tested further after #822

ethanwhite:
Now that #897 is in (Thanks again @jainamritanshu!) this is almost ready to go. The one thing that it's lacking that got added later in the #822/#897 process was the conditional to handle 2.0 vs 2.1 internal naming. Once that's added this should be ready to go.

jainamritanshu:
@ethanwhite I have made the changes for version handling similar to other scripts

scripts/predicts.py
```diff
@@ -0,0 +1,111 @@
+#retriever
+
+"""Retriever script for direct download of data data"""
+from retriever.lib.models import Table, Cleanup, correct_invalid_value
+from retriever.lib.templates import Script
+from retriever import VERSION
+from pkg_resources import parse_version
+
```
ethanwhite: Missing the `os` import needed for Line 106.

scripts/predicts.py
```diff
@@ -0,0 +1,111 @@
...
+        if parse_version(VERSION) <= parse_version("2.0.0"):
+            self.shortname = self.name
+            self.name = self.title
+
+    def download(self, engine=None, debug=False):
+        Script.download(self, engine, debug)
+        engine = self.engine
+
+        filename = "predicts_data.csv"
```
ethanwhite: It looks to me like the file the in archive is actually `database.csv` not `predicts_data.csv`. I get an error using the later, but things execute using the former.

henrykironde:
ref #937

Issue #765, opened 2017-01-14
Update internals to match datapackage spec


ethanwhite:
After updating our language to match the datapackage spec our internal naming doesn't match the naming in our datapackage files. We should update this to make everything easier to understand. Specifically

* tags -> keywords
* nulls -> missingValues
* name -> title
* shortname -> name

Possibly a few others I'm not thinking of.

In many cases this will basically be find & replace, but being careful to watch out for potential complexities.

jainamritanshu:
@ethanwhite I am working on this

ethanwhite:
This is complete as of the merging of #897. Thanks again @jainamritanshu!

PR #822, opened 2017-03-03
Update internals in reference to issue #765

jainamritanshu:
@ethanwhite @henrykironde 
I have made the following changes

- name -> title
- shortname -> name
- tags -> keywords
- nulls -> missingValues 

I wanted to ask to change `missingValues` to `missing_values` as the original one seems to be in camel case and not in [pep8 naming convention](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) , if you allow I would change it.
I am still finding such internal name cases which could be updated.
I guess the code is clean. Kindly go through it once and if there are any suggestion I would start working on them


henrykironde:
@jainamritanshu, we have some tests using the the old variable names in this file https://github.com/weecology/retriever/blob/master/test/test_retriever.py

jainamritanshu:
@henrykironde I have made the changes as suggested and all the tests have been passed. Just one more thing, should we change `missingValues` -> `missing_values` as the camel case seems more to be a javascript naming convention and `missing_values` is according to [pep8 naming convention](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)

henrykironde:
@jainamritanshu lets leave it as missingValues because we are following the Json datapackage standards. I believe that is the format they follow.

jainamritanshu:
@henrykironde Sure! I didn't realize that.

scripts/plant_life_hist_eu.py
```diff
@@ -9,8 +9,8 @@
 class main(Script):
     def __init__(self, **kwargs):
         Script.__init__(self, **kwargs)
-        self.name = "A database on the life history traits of the Northwest European flora" 
-        self.shortname = "plant-life-hist-eu"
+        self.title = "A database on the life history traits of the Northwest European flora" 
+        self.title = "plant-life-hist-eu"
```
ethanwhite: This should be `name` instead of `title`.

jainamritanshu: @ethanwhite Sorry, my bad! Fixed it now

ethanwhite:
LGTM.

@henrykironde - make the final call on `missingValues` vs `missing_values` and then I think this is probably ready to go.

Thanks @jainamritanshu!

henrykironde:
Good lets go for missing_values.

henrykironde:
@jainamritanshu, As you finally come to the end of this issue when cleaning up the commits, 
explain clearly what you have changed. A good example of a commit for such huge changes can be seen here[ 1](https://github.com/weecology/retriever/commit/a9208eaec63107e5216543523e4bcded547f231d) [2](https://github.com/weecology/retriever/commit/4702ed05c081af938baa12309f2e8bb3f52964ad) [3](https://github.com/weecology/retriever/commit/7cfa9a72d269630cb3952ed2cfb6d475ca7fe373)

jainamritanshu:
@henrykironde I have done the required changes. Please go through it and I would be happy to work on any other changes required :smile:

henrykironde:
Working on it.

lib/cleanup.py
```diff
@@ -13,9 +13,9 @@ def floatable(value):
 def correct_invalid_value(value, args):
     """This cleanup function replaces null indicators with None."""
```
henrykironde: lets add missing values to 
"""This cleanup function replaces null indicators with None.""">> """This cleanup function replaces missing value indicators with None."""

lib/datapackage.py
```diff
@@ -53,17 +53,17 @@ def get_replace_columns(dialect):
             continue
 
 
-def get_nulls(dialect):
+def get_missing_values(dialect):
     """Get list of strings that denote null in the dataset"""
```
henrykironde: Edit the doc string the reflect the changes

henrykironde:
@jainamritanshu Some more changes are in the docs folder
``https://github.com/weecology/retriever/blob/master/docs/conf.py#L44``

Due to the recent commits, the directory structure has changed and many files have changes causing a lot of conflicts. If you could update your changes using the current master. that would make merging easy

jainamritanshu:
@henrykironde Thank you for waiting. I have made the required changes according to the new directory structure and implemented your latest suggestions. Kindly go through the code once.

henrykironde:
@ethanwhite, do you have some time to glance through this. I think its ready to go

docs/conf.py
```diff
@@ -41,7 +41,7 @@
                                                "\n| **reference:**    {}"
                                                "\n| **citation:**    {}"
                                                "\n| **description:**   {}"
-                                               "\n\n\n".format(script.name, script.shortname, reference_link,
+                                               "\n\n\n".format(script.title, script.name, reference_link,
```
henrykironde: "\n\n| **shortname:**    {}" >> "\n\n| **name:**    {}"

jainamritanshu: @ethanwhite @henrykironde Made the change.

ethanwhite:
I'll give this a read right now.

retriever/lib/datapackage.py
```diff
@@ -54,16 +54,16 @@ def get_replace_columns(dialect):
...
     if val == "" or val == []:
         # return and dont add key to dialect dict if empty val
         return
-    dialect['nulls'] = val
+    dialect['missing_values'] = val
     # change list to single value if size == 1
-    if len(dialect['nulls']) == 1:
-        dialect['nulls'] = dialect['nulls'][0]
+    if len(dialect['missing_values']) == 1:
+        dialect['missing_values'] = dialect['missing_values'][0]
```
ethanwhite: I'm guessing that these 3-4 lines are what ends up setting the values to be written to the JSON (@henrykironde - can you confirm). Is so then these should all be set to `missingValues`.

jainamritanshu:
@ethanwhite @henrykironde sorry for the typos. I have fixed those. Waiting for the confirmation for the change in `missing_values` variable.

retriever/lib/datapackage.py
```diff
@@ -54,16 +54,16 @@ def get_replace_columns(dialect):
 
 
 def get_nulls(dialect):
-    """Get list of strings that denote null in the dataset"""
-    val = clean_input("nulls (separated by ';') (press return to skip): ",
+    """Get list of strings that denote missing value in the dataset"""
+    val = clean_input("missing-values (separated by ';') (press return to skip): ",
```
henrykironde: Could we change this`missing-values`  to Either missing_values or missing values

retriever/lib/datapackage.py
henrykironde: yes

retriever/lib/datapackage.py
ethanwhite: That's OK with me. I'd recommended it be changed to it's current form to be consistent with 'table-url' etc. but it's just the user input request so I'm happy with whatever you (@henrykironde) think is best.

jainamritanshu:
@henrykironde @ethanwhite I have made the changes suggested. Kindly have a look :smile:

ethanwhite:
Thanks @jainamritanshu - I'm really sorry about the confusion here, but I didn't want you to change them all back to `missing_values`, just the ones that would end up involved with the JSON files, i.e., the ones I specifically mentioned in the inline comments.

So, the goal is to use `missing_values` everywhere except in the JSON files. Does that make sense? If not let me know and I can PR some changes into your branch to be clear about what we're going for. I probably wouldn't be able to get to that until Monday.

jainamritanshu:
@ethanwhite I apologize for no completely understanding what you meant. I will make the required changes asap.

jainamritanshu:
@ethanwhite I made the required changes. I guess everything works fine now. Kindly give it a read.

ethanwhite:
@jainamritanshu - terrible sorry to leave this PR sitting. We realized that there were some issues with doing this that hadn't occurred to us originally and needed a chance to thing through them.

The challenge for these changes is that since that all 2.x versions of the software need to work with the scripts in `master`, that's how we update scripts when something in the data changes. This means that we can't make changes to scripts that 2.0.0 doesn't know how to handle and that is the case for the changes to the Python scripts in this PR. This is something we should have realized earlier. Sorry about that.

So, how do we deal with this. I think the best (slightly hacky) answer is to have the `.py` scripts check the version of the retriever that is using them and use the appropriate values. Here are two examples:

For `scripts/bioclim.py`:

```
...
from retriever import VERSION

     def __init__(self, **kwargs):
        Script.__init__(self, **kwargs)
        if VERSION < 2.1.dev:
            self.name = "Bioclim 2.5 Minute Climate Data"
            self.shortname = "bioclim"
            self.tags = ["climate"]
        else:
            self.title = "Bioclim 2.5 Minute Climate Data"
            self.name = "bioclim"
            self.keywords = ["climate"]
        self.ref = "http://worldclim.org/bioclim"
        self.urls = {"climate": "http://biogeo.ucdavis.edu/data/climate/worldclim/1_4/grid/cur/bio_2-5m_bil.zip"}
        self.description = "Bioclimatic variables that are derived from the monthly temperature and rainfall values in order to generate more biologically meaningful variables."
        self.citation = "Hijmans, R.J., S.E. Cameron, J.L. Parra, P.G. Jones and A. Jarvis, 2005. Very high resolution interpolated climate surfaces for global land areas. International Journal of Climatology 25: 1965-1978."
        self.version = '1.1.0'
```

An alternative and probably approach would be something like:

```
...
from retriever import VERSION

     def __init__(self, **kwargs):
        Script.__init__(self, **kwargs)
        self.title = "Bioclim 2.5 Minute Climate Data"
        self.name = "bioclim"
        self.ref = "http://worldclim.org/bioclim"
        self.keywords = ["climate"]
        self.urls = {"climate": "http://biogeo.ucdavis.edu/data/climate/worldclim/1_4/grid/cur/bio_2-5m_bil.zip"}
        self.description = "Bioclimatic variables that are derived from the monthly temperature and rainfall values in order to generate more biologically meaningful variables."
        self.citation = "Hijmans, R.J., S.E. Cameron, J.L. Parra, P.G. Jones and A. Jarvis, 2005. Very high resolution interpolated climate surfaces for global land areas. International Journal of Climatology 25: 1965-1978."
        self.version = '1.1.0'

        if VERSION < 2.1.dev:
            self.shortname = self.name
            self.name = self.title
            self.tags = self.keywords
```

This makes the code for the fix a quick copy-paste into all the .py scripts.

This solves all of the cases where the changes are just in member variables, but still leaves the probably of passing `missing_values` instead of `nulls` inside `auto_create_table` in some scripts. Those calls will also have to be wrapped in conditionals like those above.

None of this is particularly elegant, but I don't see any other way around it at the moment (if you do, certainly let me know). To test that everything is backwards compatible you can run:

```
pip install retriever
```

to get 2.0.0 from pypi. Then move into the repo and checkout your development branch and run the retriever for these datasets. It will use your dev scripts, but the 2.0.0 version of the retriever. If everything works that way you should be in good shape.

Apologies again for the added complexity and confusion here. We should have implemented this before releasing 2.0, but now we just need to work around it.

jainamritanshu:
@ethanwhite no problem at all. The second approach seems to be a clean one, and I will surely make the required changes accordingly asap.

jainamritanshu:
@ethanwhite I have made some changes as suggested by you to handle the previous versions. Kindly give it a read. If every thing is alright, I will squash the commits right away.

henrykironde:
Looks good but I am still testing this PR.

henrykironde:
@jainamritanshu, please run some thing like `retriever ls` or `retriever -v` track the errors.

scripts/wood_density.py
```diff
@@ -37,6 +37,11 @@ def __init__(self, **kwargs):
         2. Provide recognition of the efforts of this group in the assembly of the data by using the citation for the database above.
         3. Recognize that these data were assembled by the group for various analyses and research questions. If any of these uses overlap with your interests, you recognize that group has precedence in addressing these questions."""
 
+        if VERSION < 2.1.dev:
```
henrykironde: So when we compare versions we can use these packages `from pkg_resources import parse_version` , `from distutils.version import LooseVersion`
I am not sure which is better, but here are some examples  
1

```
from pkg_resources import parse_version
if parse_version(VERSION) < parse_version('2.1.dev'):
```



2
```
import sys
from distutils.version import LooseVersion

a, b = LooseVersion(VERSION), LooseVersion('2.1.dev')

if a > b:
    print a
elif b > a:
    print b
elif a == b:
    print 'equal'


```

jainamritanshu:
@henrykironde Sure will do the same asap!

scripts/wood_density.py
henrykironde: I am not sure which method of the two is better. But you can test them as see what you want to be included.

jainamritanshu: @henrykironde I am looking into this.

jainamritanshu:
@henrykironde I made this PR before the merged PR of adding more vertnet scripts like vertnet_birds.py and more. So should I update with the master branch and then update?

scripts/wood_density.py
jainamritanshu: @henrykironde I looked into this for a bit. I guess we could be better off with `parse_version`.
Since `LooseVersion` can be easily tricked for some complex version use cases like - 
```
>>> LooseVersion('1.4') > LooseVersion('1.4-rc1')
False
```
With `parse_version` - 
```
>>> parse_version('1.4') > parse_version('1.4-rc2')
True
```
I know that we are not using such complex versions, but still I guess `parse_version` would be a more better choice. Kindly let me know if there are any changes required with the comparison of versions.

jainamritanshu:
@henrykironde I have made a couple of fixes which I couldn't catch previously. Sorry for the delay in the final changes. I guess this works fine now. Could you please review it?

ethanwhite:
Thanks @jainamritanshu! Have you tested the scripts in this PR against 2.0.0? From a quick glance it looks like the `.py` scripts that use `missingValues`/`nulls` haven't handled that change in a way that won't break 2.0.0, but maybe I'm missing something.

scripts/plant_life_hist_eu.py
```diff
@@ -68,7 +68,7 @@ def download(self, engine=None, debug=False):
             new_data.close()
             self.engine.auto_create_table(Table(key,
                                                 cleanup=Cleanup(correct_invalid_value,
-                                                                nulls=[-999.9])),filename=str("new" + key))
+                                                                missingValues=[-999.9])),filename=str("new" + key))
```
jainamritanshu: @ethanwhite I used `retriever ls` to test the scripts. It worked fine, but I a too had this doubt about missingValues in .py scripts, just to be clear are you referring to something similar to this

ethanwhite: I would try doing an install with at least one of them. Best case scenario I suspect that it's running, but the null values aren't getting replaced.

jainamritanshu: This maybe a possibility. I will commit the changes for handling the use of missingValues for version < 2.0.0

henrykironde:
@jainamritanshu, It looks like the merge did not turn out as I expected. I will create a new PR, feel free to use any.

jainamritanshu:
Sure @henrykironde

ethanwhite:
This has been replaced by #897.

PR #823, opened 2017-03-04
Test-Install new scripts

### LINKED ISSUE #741: Add Test New Script functionality to tests
As we scale up the number of datasets a common issue will be testing new scripts both locally and through CI. A nice feature would be a test that detects new scripts (by comparing the local script list against those listed in `version.txt` on `master`) and performs an equivalent to the `try_run_all.py` code for those scripts.

For small to medium datasets this should be runable on CI. For large datasets we might need to add a `donottest` or `slow` flag that allows control over whether they are run or not (using pytest's `--runslow` option).

ShivamNegi:
Fix #741
try_install_modified.py tests for new scripts and attempts to install them. There is no separate test script under ./test which checks for new scripts. The task of checking for new scripts and installing them have been done in the same try_install_modified.py script. Please review it.

### DIFF
--- a/try_install_modified.py
+++ b/try_install_modified.py
@@ -1,0 +1,163 @@
+"""Attempt to update datasets in all database management systems.
+This module, when run, attempts to install datasets from modified Retriever
+scripts in the /scripts folder (except for those listed in IGNORE), for each
+engine in ENGINE_LIST() from __init__.py. In other words, it trys to install
+using all possible combinations of database platform and script and checks
+to see if there are any errors. It does not check the values in the database.
+"""
+from __future__ import print_function
+from __future__ import absolute_import
+from future import standard_library
+
+standard_library.install_aliases()
+import os
+import sys
+import urllib.request
+import urllib.parse
+import urllib.error
+from distutils.version import LooseVersion
+from imp import reload
+from retriever.lib.tools import choose_engine
+from retriever import MODULE_LIST, ENGINE_LIST, SCRIPT_LIST
+from version import get_module_version
+
+reload(sys)
+if hasattr(sys, 'setdefaultencoding'):
+    sys.setdefaultencoding('latin-1')
+if os.name == "nt":
+    os_password = "Password12!"
+else:
+    os_password = ""
+
+MODULE_LIST = MODULE_LIST()
+ENGINE_LIST = ENGINE_LIST()
+SCRIPT_LIST = SCRIPT_LIST()
+TEST_ENGINES = {}
+IGNORE = [
+    "forest-inventory-analysis",
+    "bioclim",
+    "prism-climate",
+    "vertnet",
+    "NPN",
+    "mammal-super-tree"
+]
+IGNORE = [dataset.lower() for dataset in IGNORE]
+UPSTREAM_VERSIONS = {}
+
+
+def to_string(value_to_str):
+    if sys.version_info >= (3, 0, 0):
+        return value_to_str.decode("UTF-8")
+    else:
+        return value_to_str
+
+
+def get_modified_scripts():
+    """Get modified script list, using version.txt in repo and master upstream"""
+    modified_list = []
+    version_file = urllib.request.urlopen("https://raw.githubusercontent.com/weecology/retriever/master/version.txt")
+    local_repo_scripts = get_module_version()  # local repo versions
+
+    version_file.readline()
+    for line in version_file.readlines():
+        master_script_name, master_script_version = to_string(line).lower().strip().split(",")
+        UPSTREAM_VERSIONS[master_script_name] = master_script_version
+
+    for item in local_repo_scripts:
+        local_script, local_version = item.lower().split(",")
+        # check for new scripts or a change in versions for present scripts
+        # repo script versions compared with upstream.
+        if local_script not in UPSTREAM_VERSIONS.keys():
+            modified_list.append(os.path.basename(local_script).split('.')[0])
+        else:
+            if LooseVersion(local_version) != UPSTREAM_VERSIONS[local_script]:
+                modified_list.append(os.path.basename(local_script).split('.')[0])
+
+    return modified_list
+
+modified_scripts = get_modified_scripts()
+
+if modified_scripts is None:
+    print("No new scripts found. Database is up to date.")
+    sys.exit()
+
+# If engine argument, tests are only run on given engines
+if len(sys.argv) > 1:
+    ENGINE_LIST = [
+        e for e in ENGINE_LIST
+        if e.name in sys.argv[1:] or
+        e.abbreviation in sys.argv[1:]
+        ]
+
+if os.path.exists("test_modified"):
+    os.system("rm -r test_modified")
+os.makedirs("test_modified")
+os.chdir("test_modified")
+
+dbfile = os.path.normpath(os.path.join(os.getcwd(), 'testdb.sqlite'))
+
+engine_test = {
+    "postgres": {'engine': 'postgres',
+                 'user': 'postgres',
+                 'password': os_password,
+                 'host': 'localhost',
+                 'port': 5432,
+                 'database': 'postgres',
+                 'database_name': 'testschema',
+                 'table_name': '{db}.{table}'},
+
+    "mysql": {'engine': 'mysql',
+              'user': 'travis',
+              'password': '',
+              'host': 'localhost',
+              'port': 3306,
+              'database_name': 'testdb',
+              'table_name': '{db}.{table}'},
+
+    "xml": {'engine': 'xml',
+            'table_name': 'output_file_{table}.xml'},
+
+    "json": {'engine': 'json',
+             'table_name': 'output_file_{table}.json'},
+
+    "csv": {'engine': 'csv',
+            'table_name': 'output_file_{table}.csv'},
+
+    "sqlite": {'engine': 'sqlite',
+               'file': dbfile, 'table_name': '{db}_{table}'}
+}
+
+for engine in ENGINE_LIST:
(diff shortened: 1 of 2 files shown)


try_install_modified.py
```diff
@@ -0,0 +1,147 @@
...
+                        module.SCRIPT.shortname
+                    )
+                )
+print('')
+if errors:
+    print("Engine, Dataset, Error")
+    for error in errors:
+        print(error)
+else:
+print("All tests passed")
```
henrykironde: print wrong indent

henrykironde:
@ShivamNegi could you help me explain how you tested this functionality, I have tried to test it and still not working.

henrykironde:
I recommend looking at how the version number are obtained and compared in #844

ShivamNegi:
@henrykironde, I apologise for the mess I have updated the pull request.
```./retriever/version.py``` has some functions which are being used in ```try_install_modified``` that is the reason for adding it under the ```./retriever``` folder. I know the placement is not right. Need help in placing it. 
Leaving that aside I tested ```try_install_modified```, I think it works.

File: retriever/try_install_modified.py [hunk 2]
```diff
@@ -0,0 +1,139 @@
...
+to see if there are any errors. It does not check the values in the database.
+"""
+from __future__ import print_function
+from __future__ import absolute_import
+import os
+import sys
+from imp import reload
+from retriever.lib.tools import choose_engine
+from retriever import MODULE_LIST, ENGINE_LIST, SCRIPT_LIST
+from retriever.version import get_module_version, update_version_file
```
zhangcandrew: I think the line 

```
from version import get_module_version, update_version_file
```
should fix this problem, so you can get rid of the redundant version.py

ShivamNegi: ```version.py``` is in the parent directory, so directly importing will not work.
Using ```os```, ```sys``` ```version``` could be imported changing the relative path. But I doubt it is a good way to implement. Can functions ```get_module_version```, ```update_version_file``` be added in ```retriever``` somehow? Because a function like ```update_version_file``` could be very handy when implementing a python interface.

zhangcandrew: I see. Adding update_version_file might be beneficial, but I don't think it's absolutely necessary. That would need the approval of @henrykironde though. I've also looked over the code again, and added a few more suggestions which completely circumvent the issues you were facing earlier with this import. It seems to work after those, so test it out and let me know.

retriever/try_install_modified.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def get_modified_scripts():
+    """Get modified script list compared to version.txt."""
+    modified_scripts = []
+    script = get_module_version()
+    with open("version.txt", "r") as version_file:
+        version_file.readline()
+        for counter, line in enumerate(version_file.readlines()):
```
zhangcandrew: Test out this version of def get_modified_scripts():

```
def to_string(value):
    if sys.version_info >= (3, 0, 0):
        return value.decode("UTF-8")
    else:
        return value

def get_modified_scripts():
    subprocess.check_output(
    'git fetch https://github.com/weecology/retriever.git master 2>/dev/null',
    shell=True)

    output = subprocess.check_output(
                "git diff --name-only FETCH_HEAD ../scripts",
                shell=True)

    # get list of changed scripts
    scripts_changed = to_string(output).splitlines()

    return scripts_changed
```

This will grab all modified scripts relative to upstream/master, like we do in hooks/precommit, and relieves us from having to call get_module_version() with slightly more readable code. Having a method like get_modified_scripts() could be very useful though in the future, but we should run it through @henrykironde first before we make changes to the retriever __init__ file.

retriever/try_install_modified.py
```diff
@@ -0,0 +1,139 @@
...
+            else:
+                errors.append((key, "No connection detected......" + module.SCRIPT.shortname))
+
+if errors:
+    print("Engine, Dataset, Error")
+    for error in errors:
+        print(error)
+else:
+    print("All tests passed. All scripts are updated to latest version.")
+    update_version_file()
```
zhangcandrew: I've replaced this line with:
```
subprocess.check_output(
                "python ../version.py",
                shell=True)
```
which just runs the version.py file from the prior directory.

zhangcandrew:
@ShivamNegi I noticed that the code doesn't update the actual version numbers in the version.txt or in the scripts itself. 

I don't think having the functionality to update the scripts is necessary since it will be implemented in a PR very soon, so maybe solidifying the idea of just testing those scripts that have been modified would be good for this PR, instead of testing the scripts and then trying to update version.txt. Let me know what you think!

retriever/try_install_modified.py
ShivamNegi: @zhangcandrew, line 139 is calling the function which should update the script version numbers in ```version.txt```.

ShivamNegi:
Yes, I agree on making this issue related to only installing the modified scripts .
@henrykironde, now I see you had referenced #844 before but I misunderstood it for something else.
@zhangcandrew, thanks! I am looking into the changes mentioned.

retriever/try_install_modified.py
zhangcandrew: I know. Running python version.py should accomplish the same thing though. I might be missing something?

henrykironde:
@ShivamNegi, It is totally understandable.

retriever/try_install_modified.py
ShivamNegi: @zhangcandrew , ```git diff --name-only FETCH_HEAD ../scripts``` is showing ```error: Could not access 'FETCH_HEAD'``` for me.

zhangcandrew: I might have gotten the relative directories wrong. Can you try running this sequence of statements in the terminal: 

```
git fetch https://github.com/weecology/retriever.git master 2>/dev/null
git diff --name-only FETCH_HEAD scripts
```
to see what gets outputted? The first one should output nothing, and the second one should output changed scripts, so you can test it by modifying some scripts and then running those two commands. If that works, change the lines to that in the file and try running it again. That should work.

ShivamNegi: Yes, now it works.

henrykironde:
@ShivamNegi, Let me know when the PR is ready for testing and review

ShivamNegi:
@zhangcandrew, In your implementation of ```get_modified_scripts```, why are the script version numbers are not being checked with the ```version.txt``` file? From what I understand it is a ```git diff``` of the ```scripts``` folder.

zhangcandrew:
Right so from my understanding of your implementation of try_install_all_modified, we want to test out and install all modified scripts. In that case, we don't explicitly need to check version.txt, just finding those scripts which we have modified for testing and installing purposes would be good for us. 

There may also be cases where when modifying scripts, we forget to change the version number, so running the prior implementation of try_install_all_modified would not test this script since version number in the script and version.txt would not be changed, however, we can circumvent this problem by just looking for any script which has been modified anywhere to be tested and installed. 

Let me know if I might have missed anything that you covered.

ShivamNegi:
Okay, so we have gone completely off track from #741. Which asked to check the ```version.txt``` against the installed scripts.
Now ```get_modified_script``` identifies changed scripts only which have been ```commited```. Instead of running a ```try_install_modified``` can't the user just run a ```retriever update``` which downloads the changed scripts.

henrykironde:
I will review this right now and let you know

zhangcandrew:
@ShivamNegi Our new get_modified_script now returns any scripts which have been changed locally in reference to upstream/master not only those that have been committed. So what is returned after the commands are run is a list of changed local scripts.

I think this is a better implementation of finding those scripts which have been changed locally vs checking through version.txt, and I think it achieves the same end goal, but I could be wrong.

henrykironde:
@ShivamNegi made some changes and made a pr to your branch. All you have to do is merge it 👍

henrykironde:
@ShivamNegi Clean up the commits

try_install_modified.py
```diff
@@ -0,0 +1,165 @@
...
+    for line in version_file.readlines():
+        master_script_name, master_script_version = to_string(line).lower().strip().split(",")
+        UPSTREAM_VERSIONS[master_script_name] = master_script_version
+
+    for item in local_repo_scripts:
+        local_script, local_version = item.lower().split(",")
+        # check for new scripts or a change in versions for present scripts
+        # repo script versions compared with upstream.
+        if local_script not in UPSTREAM_VERSIONS.keys():
+            modified_list.append( os.path.basename(local_script).split('.')[0])
```
ethanwhite: Remove space after `(`

try_install_modified.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            if LooseVersion(local_version) == UPSTREAM_VERSIONS[local_script]:
+                pass
+            else:
+                modified_list.append( os.path.basename(local_script).split('.')[0])
```
ethanwhite: What about shortening Lines 72-76 to:

```
elif LooseVersion(local_version) != UPSTREAM_VERSIONS[local_script]:
    modified_list.append( os.path.basename(local_script).split('.')[0])
```

This is shorter, removes a layer of nesting, and makes is clearer that we're doing two checks "Is it new" and "Has it changed" and adding it to `modified_list` if either of those is true.

henrykironde:
@ShivamNegi, Could be get this updated ad requested. Let me know when you are done

ShivamNegi:
Could we close this PR for now. I am running into import errors due to the conflicting file structure. I will open up a new issue for incorporating what @ethanwhite said.

henrykironde:
Closed as requested

PR #828, opened 2017-03-06
Added a pre-commit hooks for developer use

zhangcandrew:
Automatically updates version.txt when committing if scripts staged for
changes have not been updated within the script or within version.txt.

Hook acts before a commit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,21 @@ from __future__ import absolute_import
 from setuptools import setup
 from pkg_resources import parse_version
 import platform
+import sys
+import warnings
 import os
+import subprocess
 
 current_platform = platform.system().lower()
 extra_includes = []
+
+if current_platform == "windows":
+    os.system("chmod +x hooks/pre-commit")
+    os.system("mklink hooks/pre-commit .git/hooks/pre-commit")
+
+else:
+    os.system("chmod +x hooks/pre-commit")
+    os.system("ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit")
 
 __version__ = 'v2.0.0'
 with open(os.path.join("retriever", "_version.py"), "w") as version_file:
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
+
+import os
+import sys
+import urllib.request, urllib.parse, urllib.error
+from distutils.version import LooseVersion
+import re
+import subprocess
+
+pattern = re.compile(r'[."]version["\s":=]+(\d+\.\d+.\d+)')
+version_dict = {}
+versionfile = urllib.request.urlopen(
+    "https://raw.githubusercontent.com/weecology/retriever/master/version.txt")
+versionfile.readline()
+
+for lines in versionfile:
+    key_values = lines.decode("UTF-8").split(",")
+    version_dict[key_values[0]] = key_values[1]
+
+
+def get_script_version(file_name):
+    file_name = os.path.split(file_name)[1]
+    return LooseVersion(version_dict[file_name])
+
+
+# check if script directory changes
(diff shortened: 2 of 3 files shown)


zhangcandrew:
@henrykironde I switched over to this new PR. 

in response to: 
```
else:
# os.system("ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit")
```

I was using it just in case the developer already had git hooks in place. We create a symbolic link from our hook to theirs instead of overwriting their file so that they can keep their own git hooks along with the ones we have for them. I'm not sure if it's too useful, but I added it as a precaution.

setup.py
```diff
@@ -4,10 +4,35 @@
...
 
 current_platform = platform.system().lower()
 extra_includes = []
+if current_platform == "darwin":
+    if(subprocess.check_output("[ ! -f .git/hooks/pre-commit ] && echo \"Not Found\"") == "Not Found"):
+      os.system("chmod +x hooks/pre-commit")
+      os.system("cp -p hooks/pre-commit .git/hooks/pre-commit")
+    else:
+      os.system("ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit")
+    try:
```
henrykironde:
```
    try:
         import py2app
     except ImportError:
         pass
     extra_includes = []
```

not needed

setup.py
```diff
@@ -4,10 +4,35 @@
...
+        import py2app
+    except ImportError:
+        pass
+    extra_includes = []
+elif current_platform == "windows":
+    os.system("chmod +x hooks/pre-commit")
+    os.system("mklink hooks/pre-commit .git/hooks/pre-commit")
+    try:
```
henrykironde:
```
     try:
         import py2exe
     except ImportError:
         pass
     import sys
     extra_includes = ['pyodbc', 'inspect']
     sys.path.append(
         "C:\\Windows\\winsxs\\x86_microsoft.vc90.crt_1fc8b3b9a1e18e3b_9.0.21022.8_none_bcb86ed6ac711f91")
```
not used any more

henrykironde:
I think with those changes we shall have be ready to test :+1:

setup.py
```diff
@@ -4,10 +4,23 @@
...
 from pkg_resources import parse_version
 import platform
+import sys
+import warnings
 import os
+import subprocess
 
 current_platform = platform.system().lower()
 extra_includes = []
+if current_platform == "darwin":
```
henrykironde: If Linux and OS x have the same procedures 
we can say 
```
if window:
    do all the windows checks and commands
else : 
     do command for both mac os x and linux
```

henrykironde:
let me know what you think

setup.py
zhangcandrew: If OS X and Linux have the same procedures, that would work. I have to test it on Linux still, so I will circle back to you on that!

zhangcandrew: Yeah, it seems to be working on Linux now. Another problem we now face though is that the commands module does not run in python 3.5. I might need to switch the entire pre-commit over to subprocess

henrykironde: go for it

setup.py
```diff
@@ -4,10 +4,23 @@
...
+    if(subprocess.check_output("[ ! -f .git/hooks/pre-commit ] && echo \"Not Found\"") == "Not Found"):
+      os.system("chmod +x hooks/pre-commit")
+      os.system("cp -p hooks/pre-commit .git/hooks/pre-commit")
+    else:
+      os.system("ln -s -f ../../hooks/pre-commit .git/hooks/pre-commit")
+
+elif current_platform == "windows":
```
zhangcandrew: I think the mklink serves the same purpose in setup.py, but I could be wrong.

zhangcandrew:
Just updated pre-commit and setup.py. I'm currently using the git hook, and I have tested it with a few scripts. Everything seems to be ok for now on Ubuntu. Let me know how things go on your side!

henrykironde:
Closing because I have added the changes to a new pr

Issue #819, opened 2017-03-03
Add option to install Vernet datasets separately


KristinaRiemer:
Currently when someone installs the Vertnet dataset, it automatically downloads all five datasets (Amphibian, Reptile, Mammal, Bird, and Fish). Because they're so large, it would be helpful to have the option to specify downloading only one of the five datasets at a time, especially because some people will only need/use one of them.

zhangcandrew:
@henrykironde Do you think it'd be better if we made 5 different scripts instead of 1 with the option to choose from any of the 5 included?

ethanwhite:
Seems like we have a couple of choices here:

1. Keep the current scripts and add additional scripts for each separately `vertnet-amphibians`, `vertnet-reptiles`...
2. Just have separate scripts for each separately `vertnet-amphibians`, `vertnet-reptiles`..., without an "everything script
3. Leave things they way they are

I don't see any harm to (1), except that we'll need to remember to update 6 scripts when there is a new release.

A bigger project would be to create a dataset type that is a combination of multiple datasets, so that `vertnet` just becomes a list of the 5 datasets that make it up. That would be a cool functionality to add at some point. but is probably a bit far off to be a solution to this (and it would require creating the separate files anyway).

What are folks thoughts on the 3 options above?

henrykironde:
I will go with, Keep the current scripts and add additional scripts for each separately vertnet-amphibians, vertnet-reptiles...

If anyone wants to work on this, tag your self or comment.

zhangcandrew:
I thinks that's a good way to handle the situation, since they will still have the option to download data sets from vertnet if they would like to. In the future, would it be worth it to add a feature that lets users download multiple datasets in one command? 

Something like: 
```
retriever install csv iris portal pantheria
```
to install csvs for the iris, portal, and pantheria datasets all at once?

ethanwhite:
> In the future, would it be worth it to add a feature that lets users download multiple datasets in one command?

Yes, that would be great!

ShivamNegi:
I would like to work on this. 
So 5 new scripts need to be made namely ```vertnet-amphibians```, ```vertnet-birds```, ```vertnet-fishes```, ```vertnet-mammals```, ```vertnet-reptiles```.
@zhangcandrew that would be a very useful addition.

henrykironde:
Yes

ShivamNegi:
How do I add a new script to Data Retriever list?
```
[...]

PR #829, opened 2017-03-06
Adding vertnet amphibians script

ShivamNegi:
#819 
@henrykironde I couldn't test this script. As I am unable to add it to the retriever data set list.


scripts/vertnet_amphibians.py
```diff
@@ -0,0 +1,237 @@
+# -*- coding: latin-1 -*-
+# retriever
```
henrykironde: no space `# retriever` > `#retriever`

henrykironde: That will make it work, and update the version file, let me know if it works

henrykironde:
@ShivamNegi lets make one commit message
 ```
Adding vertnet individual dataset support

Adding vertnet birds script
Adding vertnet fishes script
Adding vertnet reptiles script
Adding vertnet mammals script
Adding vertnet amphibians script
```
Additionally the version numbers for these scripts should be "1.0.0" since they are new scripts
The version file is corrupt. Always note the script changes, in this case the script is corrupted.

**Comment by henrykironde** [comment]
You need to clean your repo.
I suggest running

```
retriever reset all
rm -r scripts
git reset --hard
python version.py
```

This will give you the right version.txt to add and commit

**Comment by ShivamNegi** [comment]
@henrykironde I get this exception on running ```python version.py```
```
➜  retriever git:(vertnet-dataset) ✗ python version.py 
Failed to load script: forest_fires_portugal (scripts)
Exception: Non-ASCII character '\xc3' in file scripts/forest_fires_portugal.py on line 9, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details (forest_fires_portugal.py, line 9) 
```

**Comment by henrykironde** [comment]
could you give me your system details, I am trying to replicate that error.

**Comment by ShivamNegi** [comment]
HP ProBook 440
8 GB ram, Intel i5 quad core processor
Running Ubuntu 16.04 LTS

running ```python version.py``` produces error in ```Python 2.7.12```. Works fine with ```Python 3.5.2```.

**Comment by ShivamNegi** [comment]
@henrykironde I think this works.

**Comment by henrykironde** [comment]
I tested the installation on windows and it looks like we fail to install. However this does not have anything to do with the scripts. I feel the scripts are fine and ready to go in. It has something to do with the data.

**Comment by ShivamNegi** [comment]
Should multiple tags be added or a single tag is fine?
```self.tags = [u'taxon', u'amphibians']```
I am saying this because all the scripts that are being intialized using ```BasicTextTemplate``` have at least 2 tags.

Also what is the citation for other files? 
Will changing ```[*]``` with ```Aves```, ```Fishes```, ```Mammals```, ```Reptilia``` do?
```
self.citation = "Bloom, D., Wieczorek J., Russell, L. (2016).  VertNet_[Amphibia]_Sept. 2016. CyVerse Data Commons. http://datacommons.cyverse.org/browse/iplant/home/shared/commons_repo/curated/VertNet_[Amphibia]_Sep2016"
```

ethanwhite:
> Should multiple tags be added or a single tag is fine?

A single tag is fine, but it should still be in a list. Multiple tags provides more information so that's why a lot of datasets have them.

> Will changing [*] with Aves, Fishes, Mammals, Reptilia do?

Yes.

ShivamNegi:
@ethanwhite, please review.

ethanwhite:
Looks great. Since @henrykironde's already signed off, in it goes. Thanks!

Issue #833, opened 2017-03-08
Allow overriding cache option for specific scripts


ShivamNegi:
```--not-cached``` option doesn't work properly with some scripts: 

```
scripts/la_selva_trees.py:    def download(self, engine=None, debug=False):
scripts/gentry_forest_transects.py:    def download(self, engine=None, debug=False):
scripts/breed_bird_survey.py:    def download(self, engine=None, debug=False):
scripts/forest_inventory_analysis.py:    def download(self, engine=None, debug=False):
scripts/prism_climate.py:    def download(self, engine=None, debug=False):
scripts/breed_bird_survey_50stop.py:    def download(self, engine=None, debug=False):
scripts/biomass_allometry_db.py:    def download(self, engine=None, debug=False):
scripts/bioclim.py:    def download(self, engine=None, debug=False):
scripts/plant_life_hist_eu.py:    def download(self, engine=None, debug=False):
scripts/vertnet.py:    def download(self, engine=None, debug=False):
scripts/wood_density.py:    def download(self, engine=None, debug=False):
scripts/intertidal_abund_me.py:    def download(self, engine=None, debug=False):
scripts/npn.py:    def download(self, engine=None, debug=False):
```
These scripts are using their own ```download``` functions.

ShivamNegi:
@ethanwhite , I will take this issue myself as a continuation of #134 .

ethanwhite:
Fixed in 3ac063c850c69d016ea45fbbce917ae387065546. Thanks @ShivamNegi!

PR #835, opened 2017-03-08
Fixing '--not-cached' option in specific scripts

ShivamNegi:
#833 
Added ```--not-cached```options for all the mentioned scripts.


henrykironde:
If scripts changed update the version number and run `python version.py`.
This will update version.txt.

ShivamNegi:
How should I update the version number. Should the last digit of the version number be changed or the first?

ethanwhite:
In this case I'd say the last number, or the equivalent of a patch level version bump. We're still trying to figure out how to thing about semantic versioning for datapackage information more generally.

ShivamNegi:
@henrykironde @ethanwhite updated script version numbers and squashed commits.

henrykironde:
Version.txt not fixed yet

ShivamNegi:
Oh! I forgot to run ```python version.py```. Updated the commit.

ethanwhite:
Thanks for this @ShivamNegi. This is how I was thinking we would solve it when you first mentioned it, but I've been looking at this for a while now and the need to make these changes to every `.py` script when this doesn't really have anything to do with individual scripts (we will never want to override this flag for a particular script) has me wondering if the initial implementation in bc0c2318a854de6c844bace6a28807f54328eae8 was the right way to go). I didn't see it at the time because it didn't jump out that we had to change every download function to accomplish this.

It seems like ideally we'd create a member variable for one of the classes that carries this information around and can be used by the relevant download functions. I played around with this is bit and I think the best option would be to do this with the instance of the `engine` class that is created in `__main__.py`.

This would involve a change like this to `__main__.py`:

```
        if hasattr(args, 'debug') and args.not_cached:f
            engine.use_cache = False
        else:
            engine.use_cache = True
```

And then changing all of the functions that started taking the `use_cache` argument in bc0c2318a854de6c844bace6a28807f54328eae8 with changes like 

```
    def download_file(self, url, filename):
        if not self.find_file(filename) or not self.use_cache:
```

This would allow us to get rid of all of the `use_cache` arguments and just attached the flag once at the beginning. It would then automatically apply to all scripts.

What do you think about this approach?

@henrykironde - what are your thoughts on this alternate approach to bc0c2318a854de6c844bace6a28807f54328eae8 and these associated changes?

henrykironde:
@ShivamNegi, I will need more time to review this before I make any suggestion. Let me finish up on one PR and will get to this in like 2 hours.

ShivamNegi:
@ethanwhite I think this would be the right approach to making the code more understandable and easy for debugging. Seeing the large number of individual download functions for scripts, creating ```use_cache``` variable in the parent class would certainly will be the right way to go.

ethanwhite:
@ShivamNegi - OK, great. Did you want to take a go at changing over to that approach?

ShivamNegi:
Yes sure!. I'll get it done.

ethanwhite:
Awesome! Thanks! I'm going to go ahead and close this since it will be easier to start with a fresh branch from `master`.

PR #836, opened 2017-03-08
Remove unicode from forest-fires-portugal citation

ethanwhite:
This was causing errors on Python 2 due to how the encoding ending up in
the compiled Python file.


henrykironde:
updates to` version.txt`

ethanwhite:
Fixed. Sorry. Definitely excited for #828 :smile:

Issue #421, opened 2016-02-26
Check MySQL and Postgres credential files


ethanwhite:
In addition to allowing users to directly provide their MySQL and PostgreSQL credentials, it should also be possible for them to store these credentials in the usual places.

We should check information given by the user to the retriever first, and then fall back on the configuration files for usernames and passwords if they are not provided.

For PostgreSQL this is `~/.pgpass` with the format:

```
hostname:port:database:username:password 
```

See: https://wiki.postgresql.org/wiki/Pgpass. `*`s can be used in place of any of the `:` separated values.

For MySQL this is `~/.my.cnf` with the format:

```
[client]
user = root
password = yourpassword
```

See: https://dev.mysql.com/doc/refman/5.5/en/option-files.html. `.my.cnf` can contain a lot of additional configuration information so we'll need to look explicitly for `user =` and `password =`.

jainamritanshu:
@ethanwhite @henrykironde I would like to work on this issue if it is still open

henrykironde:
Go for it, I  will PR you a branch that i was working on, you can start from there

henrykironde:
@jainamritanshu, the system with the branch is not with me but in case you need any clarification let me know

jainamritanshu:
Sure I will start working on the issue right away

PR #839, opened 2017-03-09
Adding the feature to lookup credentials file in reference to #421

jainamritanshu:
@henrykironde @ethanwhite I have edited the mysql and postgres engines for looking up and falling back to the default credentials provided in the configuration files of the respective databases. The searching method looks a bit untidy in mysql engine, I would love some suggestions on that part. Also I thought of overriding the `get_input()` function but I cleaned up the inputs in postgres. Kindly review the code and suggest any changes I can start working on :smile:

### DIFF
--- a/retriever/engines/mysql.py
+++ b/retriever/engines/mysql.py
@@ -1,9 +1,9 @@
 from __future__ import print_function
 from builtins import str
 import os
-import io
 from retriever.lib.models import Engine, no_cleanup
-from retriever import ENCODING, MYSQL_CONF_PATH
+from retriever import ENCODING, MYSQL_CONF_PATH, open_fr
+import re
 
 
 class engine(Engine):
@@ -20,25 +20,13 @@ class engine(Engine):
         "bool": "BOOL",
     }
     max_int = 4294967295
-    try:
-        sql_conf = io.open(MYSQL_CONF_PATH, 'r')
-        for line in sql_conf:
-            if re.search("user=", line):
-                username = re.search("\"\"", line).group(1)
-            elif re.search("password=", line):
-                password = re.search("\"\"", line).group(1)
-            else:
-                pass
-    except IOError:
-        username = "root"
-        password = ""
-        
+
     required_opts = [("user",
                       "Enter your MySQL username",
-                      username),
+                      "root"),
                      ("password",
                       "Enter your password",
-                      password),
+                      ""),
                      ("host",
                       "Enter your MySQL host",
                       "localhost"),
@@ -52,6 +40,26 @@ class engine(Engine):
                       "Format of table name",
                       "{db}.{table}"),
                      ]
+
+    def check_credentials(self):
+        """
+        Checks the credentials of the user in the file path ~/.my.cnf for the MySQL
+        credentials by checking the username provided by the user in cli.
+        If found, it updates the credentials to initialize the connection, else uses the
+        default ones.
+        """
+        self.get_input()
+        if os.path.exists(MYSQL_CONF_PATH):
+            sql_conf = open_fr(MYSQL_CONF_PATH)
+            sql_conf_lines = sql_conf.readlines()
+            for line in sql_conf_lines:
+                if re.search("user = ", line) and re.sub("user = ", "", line) == self.opts["user"]:
+                    username = re.sub("user = ", "", line)
+                    pswd_line_index = sql_conf_lines.index(line)+1
+                    while not (re.search("password = ", sql_conf_lines[pswd_line_index])):
+                        pswd_line_index+=1
+                    password = re.sub("password = ", "", sql_conf_lines[pswd_line_index])
+                    self.opts["password"] = password
 
     def create_db_statement(self):
         """Returns a SQL statement to create a database."""
@@ -121,6 +129,7 @@ IGNORE """ + str(self.table.header_rows) + """ LINES
 
     def get_connection(self):
         """Gets the db connection."""
+        self.check_credentials()
         args = {'host': self.opts['host'],
                 'port': int(self.opts['port']),
                 'user': self.opts['user'],
@@ -128,5 +137,4 @@ IGNORE """ + str(self.table.header_rows) + """ LINES
         import pymysql as dbapi
         import pymysql.constants.CLIENT as client
         args['client_flag'] = client.LOCAL_FILES
-        self.get_input()
         return dbapi.connect(**args)
--- a/retriever/engines/mysql.py
+++ b/retriever/engines/mysql.py
@@ -1,8 +1,9 @@
 from __future__ import print_function
 from builtins import str
 import os
+import io
 from retriever.lib.models import Engine, no_cleanup
-from retriever import ENCODING
+from retriever import ENCODING, MYSQL_CONF_PATH
 
 
 class engine(Engine):
@@ -19,12 +20,25 @@ class engine(Engine):
         "bool": "BOOL",
     }
     max_int = 4294967295
+    try:
+        sql_conf = io.open(MYSQL_CONF_PATH, 'r')
+        for line in sql_conf:
+            if re.search("user=", line):
+                username = re.search("\"\"", line).group(1)
+            elif re.search("password=", line):
+                password = re.search("\"\"", line).group(1)
+            else:
+                pass
+    except IOError:
+        username = "root"
+        password = ""
+        
     required_opts = [("user",
                       "Enter your MySQL username",
-                      "root"),
+                      username),
                      ("password",
                       "Enter your password",
-                      ""),
+                      password),
                      ("host",
                       "Enter your MySQL host",
                       "localhost"),
--- a/retriever/engines/postgres.py
+++ b/retriever/engines/postgres.py
@@ -1,8 +1,7 @@
 import os
 import io
 from retriever.lib.models import Engine, no_cleanup
-from retriever import ENCODING, PGPASS_FILE_PATH
-import io
+from retriever import ENCODING, PGPASS_FILE_PATH, open_fr
 
 class engine(Engine):
     """Engine instance for PostgreSQL."""
@@ -18,30 +17,22 @@ class engine(Engine):
         "bool": "boolean",
     }
     max_int = 2147483647
-    try:
-        pgpass_file = io.open(PGPASS_FILE_PATH, 'r')
-        pgpass_credentials = pgpass_file.split(':') # Split the credentials file of the format hostname:port:database:username:password
-        pgpass_lookup = {'username': pgpass_credentials[3], 'password': pgpass_credentials[4], 'host': pgpass_credentials[0], 'port': pgpass_credentials[1], 'database': pgpass_credentials[2]}
-    except IOError:
-        pgpass_lookup = {'username': "postgres", 'password': "", 'host': "localhost", 'port': 5432, 'database': "postgres"}
 
-    for key in pgpass_lookup.keys():
-        if pgpass_lookup[key] == "*" : pgpass_lookup[key] = ""
     required_opts = [("user",
                       "Enter your PostgreSQL username",
-                      pgpass_lookup['username']),
+                      "postgres"),
                      ("password",
                       "Enter your password",
-                      pgpass_lookup['password']),
+                      ""),
                      ("host",
                       "Enter your PostgreSQL host",
-                      pgpass_lookup['host']),
+                      "localhost"),
                      ("port",
                       "Enter your PostgreSQL port",
-                      pgpass_lookup['port']),
+                      5432),
                      ("database",
                       "Enter your PostgreSQL database name",
-                      pgpass_lookup['database']),
+                      "postgres"),
                      ("database_name",
                       "Format of schema name",
                       "{db}"),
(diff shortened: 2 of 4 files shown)


retriever/engines/postgres.py
```diff
@@ -17,21 +18,27 @@ class engine(Engine):
         "bool": "boolean",
     }
     max_int = 2147483647
+    pgpass_file = io.open(PGPASS_FILE_PATH, 'r')
+    pgpass_credentials = pgpass_file.split(':') # Split the credentials file of the format hostname:port:database:username:password
```
henrykironde: line to long,
```

# Split the credentials file of the format
# hostname:port:database:username:password
pgpass_credentials = pgpass_file.split(':')

```
Some thing like this would be fine

retriever/engines/postgres.py
```diff
@@ -17,21 +18,27 @@ class engine(Engine):
         "bool": "boolean",
     }
     max_int = 2147483647
+    pgpass_file = io.open(PGPASS_FILE_PATH, 'r')
```
henrykironde: I prefer using the custom open function for the retriever in ` __init__.py`
before we open the file, we should check if path exists, and if it does exist we could update else we should stick to the default(no updates to the dictionary)

jainamritanshu:
@henrykironde I just added an except condition. I would use the `__init__.py` as you suggested. Also I am just adding the default paths for windows where postgres and mysql configuration files could be found. I would do the suggested changes asap

henrykironde:
How does your mysql config file format look like?

henrykironde:
My suggestion is when we read in the conf files. we create a dictionary. 
If we look at my .pgpass sample for example. 
```
hostname:port:database:root:password1
hostname:port:database:henry:password2
hostname:3333:database:travis:password3
hostname:4322:database:postgres:password4
hostname:port:database:apveyor:password5
```
I could have several user names.
The code should be able to get the name of the user. 
There are many ways of getting this  value, for example, importing the args and you can get the user name.
Use the user name to get other arguments in case the user didn't give those arguments,
Example scenarios
`retriever install postgres  iris` ( By default, the command will use postgres's default values as given in the Postgres engine)

`retriever install postgres -u Travis` ( This should look up the values of travis from the conf file)
`Retriever install postgres -u Peter'( This is not in my pgpass conf file, so the retriever will use the defaults of Postgres engine")

The same will apply to mysql.

retriever/engines/postgres.py
```diff
@@ -17,21 +18,30 @@ class engine(Engine):
         "bool": "boolean",
     }
     max_int = 2147483647
+    try:
```
henrykironde: let us avoid the try block. If we know the user and we check the path we shall not need this. Else we need to give a specific reason or comment for the except IOError. I may be missing something let me know

jainamritanshu:
@henrykironde I made a bad assumption about one set of credentials and considering the issues description and got a little confused about prioritizing the credentials provided in the configuration files while I should have prioritized the information provided by the user to the retriever. I guess I got clear about the issue by your example scenarios. I will fix them asap. 
Correct me if I am wrong, about the `try` block I guess it could work well with the new scenarios like,
if the user provides a username and the file is not present, retriever could directly use the default credentials.
I think I got puzzled for the part of mysql file configuration since I made a custom mysql file in the path `~/.my.cnf` and according to https://dev.mysql.com/doc/refman/5.5/en/option-files.html they had written
`user = "test" `. Later I saw the default configuration file of mysql. I will make the changes for the mysql configuration file accordingly.

henrykironde:
`if the user provides a username and the file is not present, retriever could directly use the default credentials. `Right 

and 
if the user provides a username and the file **is or not** present, but the user name provided is not in the file,
retriever could directly use the default credentials.

henrykironde:
Initially, I wanted to include this logic in the get_ops module.
However, your approach of including these changes in the respective engines, is better.

As you work on this, If we could actually put some of the things in a function so that we keep the engines simple to understand. 

I will put a label on your PR, you can remove it when you are done.  :)

jainamritanshu:
Sure! I will try to include them into a function to make the code more clean.
I was just wondering if it would be good to include a username and password provided by the user in the configuration file which didn't exist previously so that it can speed up things a little the next time?

jainamritanshu:
@henrykironde I have made a couple of changes. Can you please give it a look?

retriever/lib/engine.py
```diff
@@ -35,6 +35,8 @@ class Engine(object):
     _cursor = None
     datatypes = []
     required_opts = []
+    pgpass_lookup = {}
```
henrykironde: we should not have these two changes here in the main engine. Since they have nothing to do with other engines, we should keep these changes to the specific engines.

jainamritanshu: Sure! Will fix it.

jainamritanshu: @henrykironde fixed this bug.

retriever/engines/mysql.py
```diff
@@ -19,6 +21,8 @@ class engine(Engine):
         "bool": "BOOL",
     }
     max_int = 4294967295
+
+    sql_lookup = {}
```
henrykironde: I think this should be with the addition of other comments below
self.sql_lookup = {"user": "root", "password": ""]}
or 
self.sql_lookup = {"user": required_opts[0][3], "password": required_opts[1][3]}
We set them as the defaults in the dictionary

retriever/engines/mysql.py
```diff
@@ -39,6 +43,26 @@ class engine(Engine):
                       "{db}.{table}"),
                      ]
 
+    def check_credentials(self):
+        self.get_input()
+        try:
+            sql_conf = open_fr(MYSQL_CONF_PATH)
+            sql_conf_lines = sql_conf.readlines()
+            if self.opts["user"] != "root":
+                for line in sql_conf_lines:
```
henrykironde: @jainamritanshu, do we really need this try block,  and why are we not considering ``root`` in this if
`if self.opts["user"] != "root"`

Could you give me a scenario in case I am under looking some details.

Regardless of the user value, we should be able to get the password set.
I am imagining a`` user`` as ``root`` with a password different from the default. Whatever the case may be, if we have two sets of values, the defaults values `required_opts` and the actual user supplied values when a command like 
`` retriever install mysql -u root -p password12! iris`` is used.


I think we should do something like this, 

```
# (so here we shall not need an else just process the file if it exist)
if os.path.exists(MYSQL_CONF_PATH): 
    (go ahead and read the values from line)
    for line in sql_conf_lines:
        (get the corresponding value of self.opts["user"])
        if there is a self.opts["user"] in line get the value of password. 
            if that value of password is not set in the conf file  exit the loop(we use defaults then)
            if the value of the password is there then we modify out dictionary.
            self.sql_lookup = {"user": username, "password": password} and exit the loop(no need to loop through other lines)
            by the end of the procedures, we shall have either set the new  sql_lookup tables or we use the default ones
```

jainamritanshu: @henrykironde the file path `~/.my.cnf` is for user specific options. And global options are usually set by the user in `/etc/mysql/my.cnf` usually and the users to store their root credentials in the file path. That is why I excluded user="root" as it was searching for `~/.my.cnf`. Shall I change the file path to search in for root user?

retriever/engines/mysql.py
jainamritanshu: Changes done!

jainamritanshu:
@henrykironde As I mentioned in the comment I guess it will be better if we check both the configuration files of mysql for the root user. I may be missing something. I have made the changes keeping in mind to check both the configuration and have kept that try block for catching the file not found error. Now that we have to files to scan the value I guess the method of checking the existence path might create some problems. Kindly review the changes.

retriever/engines/mysql.py
```diff
@@ -39,6 +42,37 @@ class engine(Engine):
                       "{db}.{table}"),
                      ]
 
+    def check_credentials(self):
+        self.get_input()
+        try:
+            if self.opts["user"] != "root":
```
henrykironde: This check is not needed `if self.opts["user"] != "root":`,
The logic can be summarized and below is a description of a better design for both postgres and mysql


I don't think we need a default dictionary,
``sql_lookup = {"user": "root", "password": ""}``
If we do need one, we should include a default port number and host.

We have 3 variables at this point in the code that we could use optimally.
We know the user, path to the config and required_opts file because self.get_input() does most of the initialization.

All we have to do is check if config file is there, if yes check if our user's credentials are in there.
Get the credentials and overwrite the current opts values. (This is like a greedy approach but I do think it's enough)


Given a command ``retriever install (engine_name either [postgres\mysql]) [options] iris`` with options as the examples below:

```
	retriever install postgres  -u henry 
	retriever install postgres iris 
	retriever install postgres –-user myusername –-password ***** –-host localhost –-port 5432
	retriever install mysql –-user myusername –-password ***** –-host localhost –-port 8888 –-database_name testdbase breed-bird-survey
```

we can say

If we do have a config file, like this(if file does exist)


```
hostname:port:database:root:password1
:::henry:password2
hostname:3333:database:travis:password3
hostname:4322:database:postgres:password4
hostname:port:database:apveyor:password5
hostname:port:database:apveyor:password5
hostname:port:database:apveyor:password5
```


```
	Read the lines, split and get the user
	If the User is same as what we have in the self.opts["user"](if our user is Travis)
	     # update  self.opts. 
	      For example in postgres, 
	       I can say if self.opts["user"] = line.slip(":")[the user user index](line =hostname:3333:database:v:password3)
	      do a greedy updatge and say since user credentials just update self.opts with new port number, and password and              host(from line get the user's port number, password, host)
	```

Why I am calling it greedy.

A user can have many configurations like the user `apveyor` in our example config file above.(I just would take the first read of a line having a user)
Also some users may decide to have just password and username, :::henry:password2

**Comment by jainamritanshu** [comment]
@henrykironde I have made some changes - 
Initializing the lookup dictionary - 
`self.pgpass_lookup = {"username": self.opts["user"], "password": self.opts[ "password"], "host": self.opts["host"], "port": self.opts["port"], "database": self.opts["database"]}`

`pg_map_keys = {"username": 3, "password": 4, "host": 0, "port": 1, "database": 2}` used to map the indices of credentials in the file with the pgpass_lookup keys, this way it looks more clean.
I made a huge mistake by getting confused about the line where I was trying to replace the pgpass_lookup key value pairs according to the credentials. I have corrected it.
```
for line in pgpass_file:
    pgpass_credentials = line.split(":") # split the line
    if self.opts["user"] == pgpass_credentials[pg_map_keys["username"]]:
        for key in self.pgpass_lookup.keys():
            # if there are some empty fields of credentials, they don't get replaced
            if not self.pgpass_credentials[pg_map_keys[key]] == "*" or self.pgpass_credentials[pg_map_keys[key]] == "":
                self.pgpass_lookup[key] = pgpass_credentials[pg_map_keys[key]] # Only the credential values given gets replaced
```
Made the rest of code clean in the mysql and postgres engines. I guess this should work fine now. Kindly review it suggest any changes further required.

File: retriever/engines/mysql.py [hunk 8]
```diff
@@ -39,6 +41,21 @@ class engine(Engine):
                       "{db}.{table}"),
                      ]
 
+    def check_credentials(self):
+        self.get_input()
+        self.sql_lookup = {"user": self.opts["user"], "password": self.opts["password"], "host": self.opts["host"], "port": self.opts["port"]}
```
henrykironde: @jainamritanshu, This is looking good. However, I think we can remove the dictionary `sql_lookup`. What do you think.
From what I see, we are creating an extra dictionary sql_lookup but we can just update the  self.opts dictionary.
Let me know if I am missing something

henrykironde: And if that is the case, we can do the same for postgres

jainamritanshu: @henrykironde I also thought of just modifying the `self.opts` dictionary but I thought making a separate dictionary would make the code more clean and readable. I will change the functioning to change the `sel.opts` dictionary itself if you feel so. Kindly let me know :smile:

henrykironde: Yes, I feel the same way. I always try to work with local variables. For this case I think we can just update the` self.opts`.

Lets include a doc string for def check_credentials()

jainamritanshu: Sure I am making the required changes.

jainamritanshu:
@henrykironde I have added doc string in the engines separately. I have also removed the initialization of extra lookup dictionaries and switched to changing just the `self.opts` dictionary. Kindly review it.

retriever/engines/mysql.py
```diff
@@ -39,6 +41,26 @@ class engine(Engine):
...
+        credentials by checking the username provided by the user in cli.
+        If found, it updates the credentials to initialize the connection, else uses the
+        default ones.
+        """
+        self.get_input()
+        if os.path.exists(MYSQL_CONF_PATH):
+            sql_conf = open_fr(MYSQL_CONF_PATH)
+            sql_conf_lines = sql_conf.readlines()
+            for line in sql_conf_lines:
+                if re.search("user = ", line) and re.sub("user = ", "", line) == self.opts["user"]:
```
henrykironde: This line `re.search("user = ", line)` passes for files that have `user = ` with spaces but will not pass for a line like "user= "
So we have to compile a regex that matches all the possible arrangements like the sample below



```
[client]
user=mysqluser
password=mysqlpass

```


```
[client]
user     =     mysqluser

password  =mysqlpass

```

Something else to consider, some files have settings for different groups.
I think our script should look for the default client group, which is [client] by default.


In order to design this, We can start processing line only after we hit the line containing [client]
and end the processing if any line start with [

To handle this we can set flags to denote when we start processing, i.e line contains [client]
and set a flag that will stop the processing when we hit the line starting with []  

In this way we avoid taking values from other groups

```

[mysqladmin]

user =admin4

[client]

user = retriever
password = testpass
interactive_timeout=400

[mysql]

[mysqladmin]

```

ethanwhite:
Seeing @henrykironde's comment made me think that this should be a solved problem already. It looks like there is an option argument in `pymysql.connect` called `read_default_file` that will let us read in connection information from a specified `my.cnf` file. It looks like we can still override portions of this if we need to (see [Example 5](http://www.programcreek.com/python/example/53260/pymysql.connect)).

For PostgreSQL, have we confirmed that it doesn't just work without modification? A little reading around suggested that `psycopg2` might be handling this automatically and a quick test suggests that it might be working on my machine (though I'm a bit tired so I might be missing something about my postgres permissions).

ethanwhite:
I've confirmed that postgres is already working as intended, so no need for changes there other than some documentation.

ethanwhite:
This was implemented in the way [described above](https://github.com/weecology/retriever/pull/839#issuecomment-301969600)  in d8df565.

Issue #715, opened 2016-11-21
Improve memory usage when creating cross tab


henrykironde:
Currently  `def get_ct_data` returns a list of all the lines in the dataset for large datasets.
This will result in to memory problems.  

Best way to go about this is perform the cross tab processing as we loop through each line in the add to table. Currently we are processing the cross tab lines before we start looping through the lines

ShivamNegi:
I am working on this. Can you tell me the command to use to test the `engine.py`.

ethanwhite:
To run the full test suite you can use `py.test`.

To just run the unit tests `py.test -k "test_retriever"`.

The integration tests run some tests to make sure that the cross tab functionality is working and can be run with `py.test -k "integration"`.

Let me know if you had something else in mind.

ShivamNegi:
I want to locally test the change I have done in `def get_ct_data`, which is under `./lib/engine.py`. How can I go about this?

ethanwhite:
> I want to locally test the change I have done in def get_ct_data, which is under ./lib/engine.py. How can I go about this?

Ah, got it. To do a local install go to the top level directory of the repository and run:

```
pip install . --upgrade
```

You can then run the `retriever` commands as you normally would, but it will include the changes you've made.

As part of checking to see if you're changes worked you'll also want to check to make sure they didn't break anything. You can do this by running:

```
py.test
```

from the same directory. You'll need the pytest module installed to do this which you can install using:

```
pip install pytest
```

Also what is cross tab processing?

It is taking data that is "wide" and making it "long" or properly normalized. See http://www.ats.ucla.edu/stat/stata/modules/reshapel.htm for some examples.

ShivamNegi:
Thanks.

PR #841, opened 2017-03-10
Improve memory usage in cross-tab processing

ShivamNegi:
#715 
Please review. Using a generator in ```get_ct_data```, added ```get_ct_line_length``` function which returns the number of lines in cross-tab database.

### DIFF
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -67,9 +67,9 @@ class Engine(object):
         """This function adds data to a table from one or more lines specified
         in engine.table.source."""
         if self.table.columns[-1][1][0][:3] == "ct-":
-            lines = gen_from_source(data_source)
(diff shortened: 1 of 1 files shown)


henrykironde:
if you use a generator, you will have to traverse it to the end

ShivamNegi:
I think this works now. The problem was that a generator is being used for getting all the lines too. I was using this generator twice before.
Now the `real_lines` in `add_to_table` is getting called using a generator. This reduces the memory overhead. To improve a bit more I think calling the `gen_from_source` generator twice can be reduced to one. By skipping counting the number of lines (`real_line_length`).

henrykironde:
The counting of lines helps to keep track of the progress bar.
Thanks for this @ShivamNegi

PR #843, opened 2017-03-10
Improving --not-cached option functionality

ShivamNegi:
#833 @ethanwhite @henrykironde 
Added ```engine.use_cache``` which allows for simpler and complete implementation of the ```--not-cached``` option.

To my knowledge there is still one unresolved problem in the code. It is for an exceptional usage of the ```--not-cached``` when there is no cache in the system and we are using the ```--not-cached``` option. In this case the files get downloaded twice.
The reason for this is in ```retriever/lib/templates.py``` the ```download``` function of ```BasicTextTemplate``` calls two functions of ```retriever/lib/engine.py``` one ```auto_create_table``` and another ```insert_data_from_url``` both of which are calling the ```download_file``` function.

The files don't get downloaded twice when ```--not-cached``` option is not specified because when ```download_file``` is called for the first time the data file is now present as cache for the next ```download_file``` call.
What should be done here?

### DIFF
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -37,6 +37,7 @@ class Engine(object):
     required_opts = []
     pkformat = "%s PRIMARY KEY %s "
     script = None
+    use_cache = True
     debug = False
     warnings = []
 
@@ -385,9 +386,9 @@ class Engine(object):
             db_name = name
         return db_name.replace('-', '_')
 
-    def download_file(self, url, filename, use_cache=True):
+    def download_file(self, url, filename):
         """Downloads a file to the raw data directory."""
-        if not self.find_file(filename) or not use_cache:
+        if not self.find_file(filename) or not self.use_cache:
             path = self.format_filename(filename)
             self.create_raw_data_dir()
             print("\nDownloading " + filename + "...")
@@ -399,6 +400,9 @@ class Engine(object):
                 # script. If this happens, fall back to the standard Python 2 version.
                 from urllib import urlretrieve as py2urlretrieve
                 py2urlretrieve(url, path, reporthook=reporthook)
+            finally:
+                # Download is complete, set to prevent repeated downloads
+                self.use_cache = True
 
     def download_files_from_archive(self, url, filenames, filetype="zip",
                                     keep_in_dir=False, archivename=None):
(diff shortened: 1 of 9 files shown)


ethanwhite:
Thanks @ShivamNegi. This looks really good.

To handle the double downloading could we set a flag in `download_file` so that once it has run it won't run again. E.g.,

```
    def download_file(self, url, filename):
        if not self.find_file(filename) or (not self.use_cache and not self.downloaded):
            self.downloaded = True
            path = self.format_filename(filename)
```

ShivamNegi:
Yes, I also thought so. On it.

ShivamNegi:
@ethanwhite added the ```downloaded``` attribute.

retriever/lib/engine.py
```diff
@@ -385,9 +387,10 @@ def database_name(self, name=None):
             db_name = name
         return db_name.replace('-', '_')
 
-    def download_file(self, url, filename, use_cache=True):
+    def download_file(self, url, filename):
         """Downloads a file to the raw data directory."""
-        if not self.find_file(filename) or not use_cache:
+        if not self.find_file(filename) or (not self.use_cache and not self.downloaded):
+            self.downloaded = True
```
ethanwhite: I wonder if it's clearer to set this at the end of both the `try` and `except` statements. I don't love setting it in two different places, but in hindsight it's weird to set it up front.

ethanwhite:
Looks like I'm still seeing the double downloading:

```
ethan@croryx:~/RetrieverCode ((a55d85e...))$ retriever reset all

This will remove existing scripts, cached data, and information on database connections. 
Specifically it will remove the scripts and raw_data folders and the connections.config file in /home/ethan/.retriever. 
Do you want to proceed? (y/N)
y
ethan@croryx:~/RetrieverCode ((a55d85e...))$ retriever install --not-cached sqlite mammal-masses 
=> Installing Masses of Mammals (Smith et al. 2003)

Downloading MOMv3.3.txt...
100%  0 seconds Creating table mammal_masses_MammalMasses...

Saving a copy of MOMv3.3.txt...
Progress: 5731 / 5731 rows inserted into mammal_masses_MammalMasses totaling 5731:

Done!
```

Are you seeing the same thing @ShivamNegi?

ShivamNegi:
Yes, I get the same output. But it is downloading only once. Output is showing the extra print because of the ```insert_data_from_url``` function. To correct that along with the ```use_cache``` a check for ```downloaded``` will also have to be added which I am avoiding to make to keep things simple.

retriever/lib/engine.py
ShivamNegi: I think it would be better to add it in both the ```try``` and ```except``` statements. This way it will be more understandable.

ShivamNegi: Or yes, a ```finally``` statement can be added under which ```downloaded``` can be set.

henrykironde:
Still reviewing this PR

ethanwhite:
Thanks @ShivamNegi. What about instead of having the new `downloaded` we just set `self.use_cache = True` after downloading? That line would need a comment to explain why, but then we could take out the extra conditional in `download_file` (which is a little tricky to understand anyway) and this would then cascade through to other places checking `self.use_cache`.

ShivamNegi:
@ethanwhite, removed ```self.downloaded``` and updated ```self.use_cache``` implementation.

henrykironde:
@ethanwhite looks good to me. Tested

ethanwhite:
Thanks @ShivamNegi! Excited to have this in.

PR #844, opened 2017-03-10
Add pre-commit hooks

henrykironde:
Add pre-commit to ensure Pull requests are up-to-date with master
Ensure that developers add upstream as a remote
Developers use up-to-date branches

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -2,22 +2,31 @@
 from __future__ import print_function
 from future import standard_library
 standard_library.install_aliases()
-
 import os
-import sys
-import urllib.request, urllib.parse, urllib.error
+import urllib.request
+import urllib.parse
+import urllib.error
 from distutils.version import LooseVersion
 import re
+import sys
 import subprocess
 
 pattern = re.compile(r'[."]version["\s":=]+(\d+\.\d+.\d+)')
 version_dict = {}
-versionfile = urllib.request.urlopen(
+version_file = urllib.request.urlopen(
     "https://raw.githubusercontent.com/weecology/retriever/master/version.txt")
-versionfile.readline()
+version_file.readline()
 
-for lines in versionfile:
-    key_values = lines.decode("UTF-8").split(",")
+
+def to_string(value):
+    if sys.version_info >= (3, 0, 0):
+        return value.decode("UTF-8")
+    else:
+        return value
+
+
+for lines in version_file:
+    key_values = to_string(lines).split(",")
     version_dict[key_values[0]] = key_values[1]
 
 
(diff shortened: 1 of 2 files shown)


henrykironde:
still needs test on python 2

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+
+
+def to_string(value):
+    if sys.version_info >= (3, 0, 0):
+        return value.decode("UTF-8")
+    else:
+        return value
+
+
+for lines in version_file:
```
ethanwhite: Since this is looping one line at a time let's make this `line`

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+    key_values = to_string(lines).split(",")
+    version_dict[key_values[0]] = key_values[1]
```
ethanwhite: I think this might be a little easier to read:

```
key, value = to_string(lines).split(",")
version_dict[key] = value
```

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+
+
+def get_script_version(file_name):
+    file_name = os.path.split(file_name)[1]
```
ethanwhite: I'm fine either way, but I tend to prefer `os.path.basename(file_name)` instead of `split` with the index. I like it because I know right away which part of the file name the code is using and don't have to read to and think about the index.

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+    subprocess.check_output(
+        "git log --oneline upstream/master..HEAD | wc -l", shell=True))
+logs_behind = to_string(
+    subprocess.check_output(
+        "git log --oneline HEAD..upstream/master | wc -l", shell=True))
+if int(logs_ahead) > 0 and int(logs_behind) > 0:
+    print(
+        "\nCurrent branch is behind upstream/master."
+        "\nPlease update your branch with upstream master before making commits")
+    exit(1)
```
ethanwhite: I don't think we should required folks to be rebased against the current `master` so I think we should take this out.

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+
+# get the changes staged for the next commit relative to HEAD
+staged_scripts = to_string(
+    subprocess.check_output(
+        "git diff --cached --name-only",
+        shell=True))
+
+versions_not_changed = []  # if script changes we expect to change the version
+versions_updated = True
+for files in scripts_changed:
```
ethanwhite: Let's use `file` since each item is a single `file`.

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+    if files in staged_scripts:
+        # add the file if both staged and edited
+        to_string(
+            subprocess.check_output(
+                "git add {}".format(files),
+                shell=True))
```
ethanwhite: Isn't a file already added if it is staged?

hooks/pre-commit
```diff
@@ -0,0 +1,103 @@
...
+        output = to_string(subprocess.check_output('cat {}'.format(files)))
+        match = re.search(pattern, output)
+        script_previous_version = LooseVersion(match.group(1))
+        if not script_previous_version > get_script_version(files):
+            versions_not_changed.append(files)
+            versions_updated = False
+
+if len(versions_not_changed) > 0:
+    if not versions_updated:
```
ethanwhite: Is it possible for one of these conditions to be True without the other one being True? Seems like they should yield the same result based on Lines 86-88.

hooks/pre-commit
zhangcandrew: I think this line was here mainly to circumvent the problem of having version differences in the same script between our upstream/master and a local repo. 

If we do take this line out, how do you think we should look at the problem where a script is changed in one direction on upstream/master and another locally? Should we put upstream's changes on top of the script the rebase the local changes over it, increasing the script version by the total amount of changes made?

hooks/pre-commit
zhangcandrew: That's true, they are updated in pairs, so I can change that along with the other changes.

hooks/pre-commit
ethanwhite: Thanks for the clarification @zhangcandrew. If the script itself has been changed in `upstream/master` then I think we do need to rebase since by definition the version lines would run into conflicts. That said, I don't think this will happen very often and it will show up once the PR is submitted. So, it seems like we have two choices: 1) assume this won't happen that often and deal with it in review; or 2) check to see if there are upstream changes to a changed script.

If we want to try to go for (2) I think we'd want to check the last commit hash from upstream master for the changed script using either:

```
git log --oneline upstream/master scripts/test.json
```

or some version of `git rev-parse` I can't figure out. Then see if that hash is present in the current branch or not. If so, good to go, if not, rebase.

Since @henrykironde is hoping to get this in today I'd suggest we go with (1) and open an issue to improve things by adding (2).

hooks/pre-commit
henrykironde: One case that I did notice is explained below.

A user changes a script without changing version number, The pre-commit will stop it, and ask for a version number update.
User updates the script's version number, but forgets to add the new changes

Since it was already staged, this script will have changes that are unstaged too. If a version number is updated the pre-commit will read in the file version number and will assume that every thing is okey and goes ahead to allow the commit.  

Version,txt will have the new updated version number but the script will have the old initially staged number.

ethanwhite: I never would have thought of that. Let's leave this as is then.

ethanwhite:
Closing since this has been replaced by #849.

Issue #708, opened 2016-11-11
Add ability to print a dataset specific message after installation


ethanwhite:
Sometimes we want to communicate something about a dataset to the user. E.g.,

1. The data disappeared from the web so we can't download it anymore
2. There is an issue in one of the files that prevents us from processing it so we didn't download that one (#238)

We should add a field to take a text based message and print it after the installation of the dataset to cover all of these possibilities. For the first possibility we would then remove all of the tables, thus preventing any installation, and just print out a message saying the dataset is no longer available.

kapilkd13:
Hi, I would like to work on this issue.

ethanwhite:
Great. Thanks @kapilkd13! Go for it and let us know if there's anything we can do to help.

ethanwhite:
Implemented and merged in #846. Thanks again @kapilkd13!

PR #846, opened 2017-03-13
Adding feature- print message on Installation Issue#708

kapilkd13:
adding message field in regard to Issue #708 . setting default message to " Successfully Installed Database!!". Enabled a new field -  message, when user choose `new_json`
This PR affects all json's. close scrutiny required.
Thanks

### DIFF
--- a/retriever/lib/templates.py
+++ b/retriever/lib/templates.py
@@ -15,7 +15,8 @@ class Script(object):
 
     def __init__(self, title="", description="", name="", urls=dict(),
                  tables=dict(), ref="", public=True, addendum=None, citation="Not currently available",
-                 retriever_minimum_version="", version="", encoding="", **kwargs):
+                 retriever_minimum_version="", version="", encoding="",message="", **kwargs):
+
         self.title = title
         self.name = name
         self.filename = __name__
@@ -30,6 +31,7 @@ class Script(object):
         self.retriever_minimum_version = retriever_minimum_version
         self.encoding = encoding
         self.version = version
+        self.message=message
         for key, item in list(kwargs.items()):
             setattr(self, key, item[0] if isinstance(item, tuple) else item)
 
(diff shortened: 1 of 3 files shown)


henrykironde:
Some extra files got included. Clean your branch (git clean-xf)

kapilkd13:
Hi @henrykironde I cleaned my branch with `git clean -xf`. Only extra file  I got was version.txt. Removed it.

ethanwhite:
Thanks for this @kapilkd13. My main comment is that we don't need to have a message for every dataset. The idea is that in special cases we may want to print a message, so we should have the ability to do so, but it should only be used if needed. So, in this PR we should take out all of the additions to the existing scripts and the code in `datapackage.py` that generates this for new scripts.

kapilkd13:
Hi @ethanwhite  I have removed message field from existing JSON's. In new scripts what I am thinking is, we will ask a question like `Want to add a message[y/n]` if user pressed responded Yes, then only we will add message field to it, else not. Is this approach good. Thanks.

ethanwhite:
> In new scripts what I am thinking is, we will ask a question like Want to add a message[y/n] if user pressed responded Yes, then only we will add message field to it, else not. Is this approach good. Thanks.

Yes, that's perfect. Thanks!

kapilkd13:
Hi, I have added option to put special message in `new_json` if user require it, as per your suggestion. Thanks.

retriever/lib/templates.py
```diff
@@ -112,6 +114,10 @@ def reference_url(self):
             if len(self.urls) == 1:
                 return '/'.join(self.urls[list(self.urls.keys())[0]].split('/')[0:-1]) + '/'
 
+    def printMessage(self):
+        if self.message!="":
```
ethanwhite: `""` evaluates to `False` in Python, so the Pythonic form of this conditional is `if self.message:`

ethanwhite:
Oh, and it looks we we need to rebase due to changes in `lib/templates.py`. If that's too much trouble to deal with just let us know and we can always merge this locally.

kapilkd13:
Hi @ethanwhite I made changes you suggested. Also rebased it on latest commits. I think it's ready to merge.

ethanwhite:
Awesome! Thanks @kapilkd13! In it goes.

PR #849, opened 2017-03-14
Add Pre-commit Hooks

zhangcandrew:
Add pre-commit to ensure Pull requests are up-to-date with master
Ensure that developers add upstream as a remote
Developers use up-to-date branches

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+from __future__ import print_function
+from future import standard_library
+
+standard_library.install_aliases()
+from distutils.version import LooseVersion
+import os
+import re
+import sys
+import subprocess
+import urllib.request
+
+pattern = re.compile(r'[."]version["\s":=]+(\d+\.\d+.\d+)')
+version_dict = {}
+version_file = urllib.request.urlopen(
+    "https://raw.githubusercontent.com/weecology/retriever/master/version.txt")
+version_file.readline()
+
+
+def to_string(value):
+    if sys.version_info >= (3, 0, 0):
+        return value.decode("UTF-8")
+    else:
+        return value
+
+
+for line in version_file:
+    key, value = to_string(line).split(",")
+    version_dict[key] = value
+
+
+def get_script_version(file_name):
+    file_name = os.path.basename(file_name)
+    return LooseVersion(version_dict[file_name])
+
+
+# check and get script directory changes
+subprocess.check_output(
+    'git fetch https://github.com/weecology/retriever.git master', shell=True)
+
+output = subprocess.check_output(
+    "git diff --name-only FETCH_HEAD scripts", shell=True)
+
+# get list of changed scripts
+scripts_changed = to_string(output).splitlines()
+
+# get the changes staged for the next commit relative to HEAD
+staged_scripts = to_string(subprocess.check_output(
+    "git diff --cached --name-only scripts", shell=True))
+
+versions_not_changed = []  # if script changes we expect to change the version
+versions_updated = True
+for file in scripts_changed:
+    if file in staged_scripts:
+        # add the file if both staged and edited
+        subprocess.check_output("git add {}".format(file), shell=True)
+        output = to_string(subprocess.check_output('cat {}'.format(file), shell=True))
+        match = re.search(pattern, output)
+        script_current_version = LooseVersion(match.group(1))
+        if not script_current_version > get_script_version(file):
+            versions_not_changed.append(file)
+            versions_updated = False
+
+if scripts_changed and not versions_updated:
+    print("\nThese scripts have changed, update the version numbers before commit:")
+    for items in versions_not_changed:
+        print(items)
+    exit(1)
+
+if staged_scripts and versions_updated:
+    print("Updating version.txt")
+    os.system('python version.py')
+    print("version.txt added to commit")
+    os.system('git add version.txt')
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,17 @@
 """Use the following command to install retriever: python setup.py install"""
 from __future__ import absolute_import
-import os
-
-from setuptools import setup
-from pkg_resources import parse_version
 import platform
 import os
+from setuptools import setup
+from pkg_resources import parse_version
 
 current_platform = platform.system().lower()
 extra_includes = []
+
+if os.path.exists(".git/hooks"):  # check if we are in git repo
+    os.system("cp hooks/pre-commit .git/hooks/pre-commit")
+    os.system("chmod +x .git/hooks/pre-commit")
+
 
 __version__ = 'v2.0.0'
 with open(os.path.join("retriever", "_version.py"), "w") as version_file:


hooks/pre-commit
```diff
@@ -0,0 +1,89 @@
...
+
+
+ # check if upstream is added as remote
+remotes = to_string(subprocess.check_output("git remote -v", shell=True))
+
+if u"upstream\thttps://github.com/weecology/retriever.git (fetch)" not in remotes:
+    print("\nPlease first add upstream as a remote."
+          "\ngit remote add upstream https://github.com/weecology/retriever.git")
+    exit(1)
+
```
henrykironde:
```
# make sure upstream/master is up-to-date
 logs_update = to_string(subprocess.check_output("git fetch upstream", shell=True))
```

henrykironde: we need to update upstream/master using fetch.

henrykironde:
LGTM

hooks/pre-commit
```diff
@@ -0,0 +1,89 @@
...
+def get_script_version(file_name):
+    file_name = os.path.basename(file_name)
+    return LooseVersion(version_dict[file_name])
+
+
+ # check if upstream is added as remote
+remotes = to_string(subprocess.check_output("git remote -v", shell=True))
+
+if u"upstream\thttps://github.com/weecology/retriever.git (fetch)" not in remotes:
```
ethanwhite: This check doesn't consider folks (like me) who use ssh. The relevant line looks like this for me:

`upstream	git@github.com:weecology/retriever.git (fetch)`

ethanwhite: I also worry a bit about cases where someone chooses to call this something other than `upstream`. One option is to deal with both @henrykironde's comment below and these issues by going the following:

1. Remove lines 42-45.
2. Run `git fetch https://github.com/weecology/retriever.git master` to get `upstream` for comparison regardless of local setup
3. Use `git diff --name-only FETCH_HEAD scripts` on Line 49.

This should give us the comparison we want without any requirements for how the local repository is set up.

hooks/pre-commit
zhangcandrew: I think @ethanwhite talked about this issue in pull #844, but I might not be understanding this clearly. So did you want to keep master updated with upstream?

ethanwhite: @zhangcandrew - I think @henrykironde's suggestion is that if we've going to check against `upstream`, like is currently happening on Line 49, we want to make sure that we have the current version. Fetching won't change `master` it will just make sure that `upstream/master` on the local machine matches what's on GitHub.

zhangcandrew: Oh I see I see. Got it, thanks!

hooks/pre-commit
zhangcandrew: That knocks a few birds down with one stone! I implemented the changes so that the newer version checks for differences regardless of how their upstream is setup. It seems to be working well.

henrykironde: @ethanwhite, could you explain this more

 ```
"This check doesn't consider folks (like me) who use ssh. The relevant line looks like this for me:

upstream git@github.com:weecology/retriever.git (fetch)"
```

Comment (ethanwhite) [inline hooks/pre-commit]: @zhangcandrew has fixed this already I think, but the issue is that this line looks different if you're using https vs ssh for communicating with the remote repo, and the check for having the right upstream branch wasn't working for ssh setups.

File: hooks/pre-commit [hunk 3]
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
+from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
+
+import os
+import urllib.request
+import urllib.parse
```
ethanwhite: Appears to be unused.

hooks/pre-commit
```diff
@@ -0,0 +1,85 @@
...
+import urllib.error
```
ethanwhite: Appears to be ununsed.

hooks/pre-commit
```diff
@@ -0,0 +1,85 @@
...
+        print(items)
+    exit(1)
+
+if scripts_changed and not versions_updated:
+    exit(1)
+else:
+    print("Updating version.txt")
+    os.system('python version.py')
+    print("version.txt added to commit")
+    os.system('git add version.txt')
```
ethanwhite: The else here currently runs even if no scripts have been touched at all, which results in confusing output and some extra delay in committing. It seems like this would be better (starting at the current Line 73:

```
if scripts_changed and not versions_updated:
    print("\nThese scripts have changed, update the version numbers before commit:")
    for items in versions_not_changed:
        print(items)
    exit(1)
elif scripts_changed and versions_updated:
    print("Updating version.txt")
    os.system('python version.py')
    print("version.txt added to commit")
    os.system('git add version.txt')

File: hooks/pre-commit (hunk 4 above)
Comment (zhangcandrew) [inline hooks/pre-commit]: I kept the urllib.request in because our initialization of version_file requires it, but otherwise got rid of the other imports!

**Comment by ethanwhite** [comment]
Thanks for the changes @zhangcandrew. Reading through them everything looks good, but commiting is erroring for me on Python 2 at the moment:

```
ethan@croryx:~/RetrieverCode (testinghook +)$ git commit
  File ".git/hooks/pre-commit", line 8, in <module>
    import urllib.request
ImportError: No module named request
```

Probably just an issue of getting the import right on both 2 & 3.

File: hooks/pre-commit [hunk 6]
```diff
@@ -0,0 +1,81 @@
...
+
+
+if scripts_changed and not versions_updated:
+    print("\nThese scripts have changed, update the version numbers before commit:")
+    for items in versions_not_changed:
+        print(items)
+    exit(1)
+elif scripts_changed and versions_updated:
+    print("Updating version.txt")
+    os.system('python version.py')
```
henrykironde: import os

hooks/pre-commit
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
+from __future__ import print_function
+from future import standard_library
```
henrykironde:
```
```


mover standard_library.install_aliases() here so that `urllib.request` can get imported

henrykironde: I think all we need are the imports 
```

from distutils.version import LooseVersion
import os
import re
import sys
import subprocess
import urllib.request
```
I have tested this on both python 3 and 2 and everything works fine for our work flow.(A PR where a script is changed should not have more than one version changes in various commits. We also emphasize that a PR should have one commit unless otherwise.)  Other edge cases can be handled by bypassing the pre-commit hook.

Thanks @zhangcandrew for this work

zhangcandrew: Sure thing!

zhangcandrew:
Just put in the final touches

ethanwhite:
Looks like maybe something went wrong here since the code checking for upstream as a remote is back.

zhangcandrew:
My bad, I went to squash everything into a clean commit, but it must have reverted to an older version. I've updated it now!

hooks/pre-commit
```diff
@@ -0,0 +1,82 @@
...
+            versions_not_changed.append(file)
+            versions_updated = False
+
+
+if scripts_changed and not versions_updated:
+    print("\nThese scripts have changed, update the version numbers before commit:")
+    for items in versions_not_changed:
+        print(items)
+    exit(1)
+elif not staged_scripts and versions_updated:
```
zhangcandrew: I changed this line to not staged_scripts, since we only want to update version.txt when we have staged scripts which have their versions_updated in the script.

henrykironde:
Thanks @zhangcandrew, I guess we shall have less trouble with script changes now :+1:

PR #850, opened 2017-03-15
Bug fix: edit_json and delete_json not working Issue#847

kapilkd13:
fix for Issue#847, `edit_json` and `delete_json` didn't work because of wrong `json_list`. Made json list to render Script names instead of SCRIPT.shortname.
May be Urgent.

### DIFF
--- a/retriever/lib/datapackage.py
+++ b/retriever/lib/datapackage.py
@@ -237,6 +237,7 @@ def create_json():
             contents['resources'].append(table)  
     contents['urls'] = tableUrls
     file_name = contents['name'] + ".json"
+    file_name = file_name.replace('-', '_')
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
         json_str = json.dumps(contents, output_file, sort_keys=True, indent=4,
                               separators=(',', ': '))
@@ -412,6 +413,7 @@ def edit_json(json_file):
     edit_dict(contents, 1)
 
     file_name = contents['name'] + ".json"
+    file_name = file_name.replace('-', '_')
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
         json_str = json.dumps(contents, output_file, sort_keys=True, indent=4,
                               separators=(',', ': '))
--- a/retriever/lib/datapackage.py
+++ b/retriever/lib/datapackage.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 from builtins import input
 import os
 import json
+import glob
 from time import sleep
 from retriever import SCRIPT_LIST, HOME_DIR, ENCODING
 
@@ -406,7 +407,7 @@ def edit_json(json_file):
     try:
         contents = json.load(
             open(os.path.join(HOME_DIR, 'scripts', json_file), 'r'))
-    except FileNotFoundError:
+    except (IOError,OSError):
         print("Script not found.")
         return
 
@@ -421,3 +422,23 @@ def edit_json(json_file):
         print("\nScript written to " +
               os.path.join(HOME_DIR, 'scripts', file_name))
         output_file.close()
+
+def delete_json(json_file):
+    try:
+        # delete scripts from home directory
+        if os.path.exists(os.path.join(HOME_DIR, 'scripts', json_file)):
+            os.remove(os.path.join(HOME_DIR, 'scripts', json_file))
+
+        [os.remove(x) for x in glob.glob(os.path.join(HOME_DIR, 'scripts', json_file[:-4] + 'py*'))]
+
+        # delete scripts from current directory if exists
+        if os.path.exists(os.path.join(os.getcwd(), 'scripts', json_file)):
+            os.remove(os.path.join(os.getcwd(), 'scripts', json_file))
+
+        [os.remove(x) for x in glob.glob(os.path.join(os.getcwd(), 'scripts', json_file[:-4] + 'py*'))]
+    except OSError:
+        print("Couldn't delete Script.")
+
+def get_script_filename(shortname):
+    return shortname.replace('-', '_')+'.json'
+
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -10,7 +10,7 @@ from retriever import MODULE_LIST
 
 module_list = MODULE_LIST()
 script_list = [module.SCRIPT.shortname for module in module_list]
-json_list = [module.SCRIPT.shortname for module in module_list
+json_list = ['.'.join(os.path.basename(module.__file__).split('.')[:-1]) for module in module_list
              if os.path.isfile('.'.join(module.__file__.split('.')[:-1]) + '.json')]
 
 keywords_list = set()
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -10,7 +10,7 @@ from retriever import MODULE_LIST
 
 module_list = MODULE_LIST()
 script_list = [module.SCRIPT.shortname for module in module_list]
-json_list = ['.'.join(os.path.basename(module.__file__).split('.')[:-1]) for module in module_list
+json_list = [module.SCRIPT.shortname for module in module_list
              if os.path.isfile('.'.join(module.__file__.split('.')[:-1]) + '.json')]
 
 keywords_list = set()
@@ -49,7 +49,7 @@ help_parser = subparsers.add_parser('help', help='')
 
 citation_parser.add_argument('dataset', help='dataset name', nargs='?', default=None, choices=script_list + [None])
 new_parser.add_argument('filename', help='new script filename')
-edit_json_parser.add_argument('filename', help='script filename', choices=json_list)
+edit_json_parser.add_argument('dataset', help='dataset name', choices=json_list)
 reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections',
                           choices=['all', 'scripts', 'data', 'connections'])
 install_parser.add_argument('--compile', help='force re-compile of script before downloading', action='store_true')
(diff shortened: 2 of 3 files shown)


henrykironde:
@kapilkd13, Do you want to go a head and update the new_json too

kapilkd13:
Hi @henrykironde  Can you tell me what behavior would you like for `new_json`. I tried and it is working fine on my machine. One thing I would like to point out, When creating `new_json` we do not check if user specified new script name is complying with what we are expecting https://github.com/weecology/retriever/blob/master/retriever/lib/datapackage.py#L151 here. I made few changes on my branch https://github.com/weecology/retriever/compare/master...kapilkd13:check_input here. It now checks that name only contains, lowercase letters and `-`. Also we are saving file with the script name, thus it contains `-` instead of python recommended `_`. If it's ok I would like to change all `-` to `_` before creating file with this name.

henrykironde:
new_json: Based on tests,  if you create a script having `-` , the script's filename will also have `-`.
In the current design, `new_json` takes, a script name `bird-size` This is basically the name key in the json.
When creating the file it , it creates `bird-size.json` instead of `bird_size.json`. This should also be corrected.
Try it and let me know

henrykironde:
`contents['name'] = clean_input("name (a short unique identifier; only lowercase letters and - allowed): ")`
is getting the script name. One fix is to allow the name to have `-` which is okey, but when creating the file 
replace` -`
https://github.com/weecology/retriever/blob/master/retriever/lib/datapackage.py#L239

```
    file_name = file_name.replace('-', '_')
```

kapilkd13:
Hi @henrykironde  That is what I was thinking about. :smiley:  I would do it right now. Also what do you think about checking user's file input name as I mentioned in my previous message. Can I make a PR to merge it. Thanks

retriever/lib/get_opts.py
```diff
@@ -10,7 +10,7 @@
 
 module_list = MODULE_LIST()
 script_list = [module.SCRIPT.shortname for module in module_list]
-json_list = [module.SCRIPT.shortname for module in module_list
```
henrykironde: @kapilkd13, Could you explain the change here and I test other PRs

kapilkd13: hi @henrykironde `json_list` generated here is used as hints for `edit_json` and `delete_json`. Earlier it used to list all  SCRIPT.shortname. But the problem was, here https://github.com/weecology/retriever/blob/master/retriever/__main__.py#L101 we are checking user's input against all json's File name and not SCRIPT.shortname.
So argparse was hinting/forcing to use names like `bird-size` but it didn't match with any file, as file's name is `bird_size`. And when we try to use `bird_size`, argparse declares it as invalid. 
Here I have changed hints from Scripts.shortname to Script's Filename.

henrykironde:
Hi @kapilkd13, could you give me a set of commands that you used to test this PR.

kapilkd13:
Hi @henrykironde  I used:
`retriever edit_json`
got a list of arguments containing file names, that can be edited.
`retriever edit_json any_name_from _above_list`
Similarly, 
`retriever delete_json`
got a list of arguments containing file names, that can be deleted.
`retriever delete_json any_name_from _above_list`

henrykironde:
So in order to get this working well, I will explain a few things.

So users identify the scripts with the `shortname` like what is returned when you run `retriever ls`

```
retriever ls
Available datasets : 60

abalone-age               butterfly-population-network  forest-plots-wghats      mammal-diet            mt-st-helens-veg     plant-taxonomy-us          vertnet-amphibians  
amniote-life-hist         car-eval                      fray-jorge-ecology       mammal-life-hist       NPN                  poker-hands                vertnet-birds       
antarctic-breed-bird      community-abundance-misc      gentry-forest-transects  mammal-masses          nyc-tree-count       portal                     vertnet-fishes      
bioclim                   elton-traits                  home-ranges              mammal-metabolic-rate  pantheria   

```
The command edit_json, delete_json and new json, should take the script short name and then be able to process the file 

If I call `retriever delete_json car-eval`, we should use the script list to find if the script by the shortname is available and if so, get the file name for the script and process it.

In this case the file name is `car_eval.py`

kapilkd13:
Hi @henrykironde I have understood what you are referring. I think this would involve generating filename from user passed script name [here](https://github.com/weecology/retriever/blob/master/retriever/__main__.py#L112) by something like
`(args.dataset).replace('-', '_')`
What do you think?

henrykironde:
@kapilkd13, at that same spot `args.dataset.lower()` is the shortname.
So if we are given the shortname, Could we get the module's file name and use that to delete and edit.


For the case of `new_json`, we get the scripts `shortname `and replace `-` with `_` create the filename. eg `car-eval` > `car_eval.py`
In fact we should move all the logic of finding out the script's filename and deleting it or editing it to `datapackage.py`
For example:

```
        elif args.command == 'delete_json':
            # delete existing JSON script
            for json_file in [filename for filename in
                              os.listdir(os.path.join(HOME_DIR, 'scripts')) if filename[-5:] == '.json']:
                if json_file.lower().find(args.dataset.lower()) != -1:
                    confirm = input("Really remove " + json_file +
                                    " and all its contents? (y/N): ")
                    if confirm.lower().strip() in ['y', 'yes']:
                        # raise Exception(json_file)
                        os.remove(os.path.join(HOME_DIR, 'scripts', json_file))
                        try:
                            os.remove(os.path.join(
                                HOME_DIR, 'scripts', json_file[:-4] + 'py'))
                        except:
                            # Not compiled yet
                            pass
                    return
            raise Exception("File not found")
```



```
elif args.command == 'delete_json':
    # delete existing JSON script from home directory and or script directory if exists in current dir

    confirm = input("Really remove " + json script short name(args.dataset.lower()) +
                            " and all its contents? (y/N): ")
    if confirm.lower().strip() in ['y', 'yes']:
        file_name = get_script_filename(args.dataset.lower())
        if a path exixts in the current directory "scritps"
            delete the file(file_name) in  ....path to dirr...+script/file_name
        if a path os.path.join(HOME_DIR, 'scripts') exists.(~/.retriever/scripts/scriptname)
            delete the file
       

```

kapilkd13:
Hi @henrykironde It's a great idea. :+1:   Even I was thinking of doing it as a part of my GSoC project, as when I was working on my prototype for `native python Interface for Data Retriever` I had to code duplicate entire Install functionality. I would be very happy to do this.
I am envisioning complete `delete_json` , `edit_json` and `install` functionality at one place. This would make it easy for us to reuse these functions in "Native Python Interface".
I would like our `_main_.py` to look like:
``` elif args.command == 'edit_json':
            # edit existing JSON script
               edit_json(args.dataset.lower())         # moving all functionality to edit_json function
     elif args.command == 'delete_json':
               delete_json(args.dataset.lower())      # moving all functionality to delete_json function
     else
              install(args)                                        #moving all functionality to install function```

This would make our main file much cleaner and our functionality more reusable.
If this looks good to you, can you suggest where should I move `install` function.
Thanks :smiley:

**Comment by henrykironde** [comment]
Looks great, but for now we can just work on `delete_json` and `edit_json `. With the information we have, we shall not need a lot of time to come up with the code for  the two.

**Comment by kapilkd13** [comment]
Hi @henrykironde  I have made changes suggested by you. Key changes are:
1.  moved `delete_json` functionality to `datapackage.py`
2. `FileNotFoundError` exception is not valid in python 2, fixed it in `datapackage`
3. `edit_json` took user argument in `filename` variable, changed it to `dataset`, similiar to delete_json.
`delete_json` and `edit_json` are working fine. Tested them locally

File: retriever/lib/datapackage.py [hunk 2]
```diff
@@ -405,17 +406,36 @@ def edit_json(json_file):
...
         json_str = json.dumps(contents, output_file, sort_keys=True, indent=4,
                               separators=(',', ': '))
         output_file.write(json_str + '\n')
         print("\nScript written to " +
               os.path.join(HOME_DIR, 'scripts', file_name))
         output_file.close()
+
+def delete_json(json_file):
+    try:
+        #delete scripts from home directory
```
henrykironde: missing space`#delete` > # delete

retriever/lib/datapackage.py
```diff
@@ -405,17 +406,36 @@ def edit_json(json_file):
...
+        os.remove(os.path.join(HOME_DIR, 'scripts', json_file))
+        os.remove(os.path.join(HOME_DIR, 'scripts', json_file[:-4] + 'py'))
+        os.remove(os.path.join(HOME_DIR, 'scripts', json_file[:-4] + 'pyc'))
+
+        #delete scripts from current directory if exists
+        os.remove(os.path.join(os.getcwd(), 'scripts', json_file))
```
henrykironde: What happens if the path does not exist?

henrykironde:
```
+        os.remove(os.path.join(HOME_DIR, 'scripts', json_file[:-4] + 'py'))
 +        os.remove(os.path.join(HOME_DIR, 'scripts', json_file[:-4] + 'pyc'))
```
we could use one statement  `os.remove(os.path.join(HOME_DIR, 'scripts', json_file[:-4] + 'py*'))`

kapilkd13: Hi, henry I referred [this](http://stackoverflow.com/questions/10840533/most-pythonic-way-to-delete-a-file-which-may-not-exist) for `os.remove()`. what do you suggest, which will be better?
About using `py*`, smart choice :smiley: I would do that.

kapilkd13:
Hi, I have added final changes:
now checking if a json file exists before removing it `os.remove`. In case of python files, `glob` is used for wildcard `py*`. glob also make sure that file exists.

henrykironde:
Thanks for working on this

kapilkd13:
Hi Henry, 
this is off the topic. I am thinking of working on adding `logging` facility to our retriever tool( This way, I could also learn using logging tool.) So I would like to create a `work in progress` branch , so that we can have all the discussion related to `logging` there. Is it advisable or is there a better way. Thanks :smile:

henrykironde:
Hi we always welocome new ideas to the Data Retriever. Please Explain logging. 
YOu can send me an email of create and Issue for the discussion since this PR is closed

PR #852, opened 2017-03-15
Repair doc build on readthedocs

henrykironde:
The build was failing because the module try_install_all.py
was being run causing the server to timeout

Remove the check for updates in the docs/conf.py
since it's run on installation
Remove inclusion of home directory .py/ non core source


ethanwhite:
Hooray: https://readthedocs.org/projects/retriever/builds/5151598/

PR #854, opened 2017-03-15
Adding dataset out of service functionality in reference to #562

jainamritanshu:
@henrykironde @ethanwhite I have added a default message that will be indicated if the dataset is not available or is out of service. I wondered upon the [comment](https://github.com/weecology/retriever/issues/562#issuecomment-232826187) and I guess a good thing would be to add a field in the .json files for the message to show if a dataset goes out of service. As in the comment there were two different appropriate messages to show to user for two different datasets. Should I go on and include an optional field in the datapackage?


henrykironde:
@jainamritanshu, Actually that was the goal. If we put the print statement as you have done, In cases when the network is down, we shall have the wrong message.

So like you have stated, `` I guess a good thing would be to add a field in the .json files``, I recommend that we do as you have said( I guess a good thing would be to add a field in the .json files)

So we need to add a key in the `datapackage.py`,  and it's presence in a scripts(both json and .py) lets the user know that the data is not available.

This message varies based on the status of the dataset. 
For example: If a dataset provider i.e Iris says the files will be available after June-30-2017, we shall update out iris script buy including this key and a message.

"appropriate_key_name": "Sorry, We are having difficulties and the data will be available not later than  June-30-2017"

henrykironde:
Additionally, It would be good to have another key that lets user know that the dataset is very huge. 
A good example would be vertnet. It takes a lot of time to download and install this dataset.
If we could also provide a key that just warns people that it will take a long time for that particular dataset. This would be another PR but similar changes that you make for both. There is an issue for this too.

jainamritanshu:
@henrykironde I have made some required changes and have added a field of `dataset_availability=True` in all the scripts by default. Kindly review the code and tell me if I am missing something.

henrykironde:
Sorry if I was not clear.
By default, I meant that if a script is working fine or doesn't  need that kind of information then we don't have to set it specifically for that script.
This means we could initialize it like citation.
https://github.com/weecology/retriever/blob/master/retriever/lib/templates.py#L17

jainamritanshu:
@henrykironde I apologize for not understanding clearly. I have made some changes, kindly give it a read.

henrykironde:
Could you provide some script examples you have used to test this addition.

jainamritanshu:
To test this I made a custom small json script using `new_json` (replication of abalone-age.json with the new key), since the former datapackages don't have the key `"dataset_availability"`. `retriever install` to install the dataset. If the key `"dataset_availability"` is "True", the file downloads, else gives the appropriate message to user.

henrykironde:
Could you please make one and send it to me to test

ethanwhite:
Ideally we should add a test to the test suite as well so that this gets tested automatically going forward.

jainamritanshu:
@ethanwhite You are right. My bad. I will add a test suite for the same asap. 
@henrykironde now that I am writing test, should I send you the script to test or proceed with the test right away?

ethanwhite:
Thanks @jainamritanshu. Both would be good.

henrykironde:
Everything looks good, I was going to add the test but if you want go ahead and let me know where I could help.

jainamritanshu:
Sure @ethanwhite @henrykironde. Will start working on it.

jainamritanshu:
@henrykironde @ethanwhite I am sorry for the delays in PR's, I have been sick for the past few days. Anyways I have started working again. This was the script I used to test the `dataset_availability` functionality - 
```
{
    "citation": "Lichman, M. (2013). UCI Machine Learning Repository [http://archive.ics.uci.edu/ml]. Irvine, CA: University of California, School of Information and Computer Science.",
    "dataset_availability": "this dataset is not available for now",
    "description": "Database to aid in the prediction of the age of an Abalone given physical measurements",
    "encoding": "utf-8",
    "homepage": "http://archive.ics.uci.edu/ml/datasets/Abalone",
    "keywords": [],
    "name": "abalone-age-test2",
    "resources": [],
    "retriever": "True",
    "title": "Abalone Age and Size Data",
    "urls": {},
    "version": "1.0.0"
}
``` 

Also for the test suite to test this functionality, shall I go on and a method for the same in `test_retriever.py`?

retriever/lib/datapackage.py
```diff
@@ -164,6 +164,8 @@ def create_json():
     contents['retriever_minimum_version'] = "2.0.dev"
     contents['encoding'] = clean_input("encoding: ", ignore_empty = True)
     if is_empty(clean_input("encoding: ", ignore_empty = True)) : contents['encoding'] = ENCODING
+    contents['dataset_availability'] = clean_input("Message for non-availability of dataset: ", ignore_empty = True)
+    if is_empty(clean_input("Message for dataset availability: ", ignore_empty = True)) : contents['dataset_availability'] = 'True'
```
henrykironde: 1. How about checking  contents dictionary for key 'dataset_availability' and if we don't have a value in that dictionary, we can set it to  contents['dataset_availability'] = 'True'

2. I think this alone` contents['dataset_availability'] = clean_input("Message for non-availability of dataset: ", ignore_empty = True)` would be enough. In this case we are saying if for some case we create a and want to say that the data is having a problem right now. we could go ahead and include it else just ignore it


The same issue happens with 
```
contents['encoding'] = clean_input("encoding: ", ignore_empty = True)
if is_empty(clean_input("encoding: ", ignore_empty = True)) : contents['encoding'] = ENCODING
```
When you run this program, try to make the first variable empty when supplying encoding. 
It will ask for` encoding` twice. 
So the first time you are asked for the value and you don't supply one.  `contents['encoding'] = clean_input("encoding: ", ignore_empty = True)` the next if statement will display the same message again.
I think we need to fix that.

jainamritanshu: @henrykironde I didn't realize that. Will make the changes asap.

jainamritanshu:
@henrykironde I have made some changes. Kindly have a look

ethanwhite:
Closing since this is stale and not a top priority at the moment.

PR #861, opened 2017-04-03
Move testing datasets from Ecological Archives to Figshare

henrykironde:
The datasets from ecologial Archieves have been moved to a much more stable environment, Figshare.


ethanwhite:
Thanks @henrykironde

PR #869, opened 2017-04-07
Move la-selva-trees links from Ecological Archives to figshare

ethanwhite:
The old Ecological Archives is now unstable.


henrykironde:
@ethanwhite, Tested and all looks fine. waiting for the response on the issue page about citation

PR #872, opened 2017-04-07
Move elton-traits links from Ecological Archives to Figshare

henrykironde:
The old Ecological Archives is now unstable.


henrykironde:
Travis is just tired today.

PR #886, opened 2017-04-10
Replace Non-ASCII -

henrykironde:


henrykironde:
this was failing on python2

PR #888, opened 2017-04-10
Move amniote links to Figshare

henrykironde:


henrykironde:
@ethanwhite do we need all these files in amniote?

ethanwhite:
Yes, it would be best to have all of those files. It would also be good to have `_Synonyms` tables and the `Dataset_Field_Information` table from the `Supplemental Materials` directory.

This PR is currently erroring for me.

henrykironde:
@ethanwhite, The Authors need to change the files on figshare.
Initially, I think this downloaded all the files well.
homepage:https://figshare.com/collections/An_amniote_life-history_database_to_perform_comparative_analyses_with_birds_mammals_and_reptiles/3308127

Looking at the files now, they have changed the names due to wrong slashes 
"\" instead of "/"
https://figshare.com/articles/Full_Archive/3563457

ethanwhite:
I've emailed figshare about this and will hopefully hear back soon.

henrykironde:
@ethanwhite based on the previous script [amniote_life_hist.json](https://github.com/weecology/retriever/blob/master/scripts/amniote_life_hist.json), we used cross-tab to produce the final file for this dataset in table `main` .

With the new script and additional file/tables included, could we verify the data that we want to actually  install and the data/tables we want to apply the cross tab processing to.

ethanwhite:
I think for consistency (and because I think it's generally a more useful data structure) we should keep the main table named main and with the previously used data structure. So that means renaming `Data_Aug` to `main` and doing the same cross-tab restructuring as before (everything after common_name gets converted to the long format). If we're doing that we should probably do the same think for all of the other tables as well. We can drop the "sparse" table since it's just the same data in a different format.

In reality this database could use a more substantial restructuring that combines the main and the Range tables with "summary statistic" as another column, but I think we should probably stick with just cross tabbing the three relevant tables for now.

henrykironde:
@ethanwhite have a look at the script and data produced

ethanwhite:
Looks like this last round of changes also ended up in the `main` table. We just want them in `reference`.

henrykironde:
@ethanwhite sorry about the previous mistake, I checked the changed files and left the main assuming that it was fine. 

I have made the changes and tested all the files this time. Hope I did not go wrong again :)

ethanwhite:
No worries. Looks great. Thanks!

Issue #876, opened 2017-04-07
update fray-jorge-ecolog data


henrykironde:
The data on the page has some seed data that is not currently included.

current figshare hopege: https://figshare.com/collections/Long-term_monitoring_of_mammals_in_the_face_of_biotic_and_abiotic_influences_at_a_semiarid_site_in_north-central_Chile/3305493

https://ndownloader.figshare.com/files/5625387 (Fray_Jorge_Seeds_20032005I.txt )
https://ndownloader.figshare.com/files/5625390 (Fray_Jorge_Seeds_20032005II.txt  )

zhangcandrew:
I'll add these in when I get the chance!

zhangcandrew:
Closing this issue after the merging of Fray Jorge. Feel free to open if a new pertinent issue arises

PR #907, opened 2017-05-08
Updated the Fray Jorge dataset to include extra data

zhangcandrew:
There was data in the figshare Fray-Jorge dataset that we had no
previously accounted for, so I added in the other two datasets.

addresses issue #876

### DIFF
--- a/scripts/fray_jorge_ecology.json
+++ b/scripts/fray_jorge_ecology.json
@@ -240,6 +240,100 @@
             "url": "https://ndownloader.figshare.com/files/5625384"
         },
         {
+            "dialect": {},
+            "name": "seed_03_051",
+            "schema": {
+                "ct_column": "sampling_date",
+                "ct_names": [
+                    "Dec-03",
+                    "Apr-04",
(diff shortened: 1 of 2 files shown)


henrykironde:
Looks fine in terms of structure, The data needs to be cleaned up.
A sample of  the data result from one of the files

```
record_id,species,treatment,grid,year,cover
1,ADFI,+D-P,1,1989,0.0
2,ADFI,+D-P,1,1990,0.0
.
.
337,ADTE,-D+P,4,2002,0.33
338,ADTE,-D+P,4,2003,0.0
339,ADTE,-D+P,4,2004,0.0
340,ADTE,-D+P,4,2005,0.5
341,ADTE,-D-P/-SM,5,1989,0.0
342,ADTE,-D-P/-SM,5,1990,0.0
```
The problem is with the treatment value(+D-P). IF you open them in excel, The application thinks that treatment value is a  formula and gets some rendering problems as shown below. 
 
image

let me try to figure out a solution to this

PR #912, opened 2017-05-17
Re-add high speed bulk importing for SQLite

ethanwhite:
This was originally added in bc182e4 and then removed in defdea4 because
we thought all of the functionality had been moved into the main `add_to_table`
code. However the speed-ups dropped off substantially with that move. We thought
this was due to the need to restrict the number of inserted rows due to other engines
but that turns out not to be the case. There are differences in how inserts are handled
by this code since it can assume data that doesn't need to be cleaned, specifically
chunked line reading (and possibly the parameter binding). This restores the code to
it's previous form so that the bulk inserts work quickly again. This is on the order of a
20x speed up for large clean datasets.

### DIFF
--- a/retriever/engines/sqlite.py
+++ b/retriever/engines/sqlite.py
@@ -52,6 +52,45 @@ class engine(Engine):
         insert_stmt = insert_stmt.rstrip(", ") + ")"
         return insert_stmt
 
+    def insert_data_from_file(self, filename):
+        """Perform a high speed bulk insert
+
+        Checks to see if a given file can be bulk inserted, and if so loads
+        it in chunks and inserts those chunks into the database using
+        executemany.
+        """
+        CHUNK_SIZE = 1000000
+        self.get_cursor()
+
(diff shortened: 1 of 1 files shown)


retriever/engines/sqlite.py
```diff
@@ -52,6 +52,42 @@ def get_bulk_insert_statement(self):
...
+    def insert_data_from_file(self, filename):
+        """Perform a high speed bulk insert
+
+        Checks to see if a given file can be bulk inserted, and if so loads
+        it in chunks and inserts those chunks into the database using
+        executemany.
+        """
+        CHUNK_SIZE = 1000000
+        self.get_cursor()
+        ct = len([True for c in self.table.columns if c[1][0][:3] == "ct-"]) != 0
```
goelakash: A comment to describe this operation would be useful. Also, a descriptive variable name would be better.

ethanwhite:
Thanks @goelakash, both good calls. Changes made.

PR #917, opened 2017-05-30
Added function for ls, as retriever.lib.datasets

ShivamNegi:
@henrykironde @ethanwhite

### DIFF
--- a/retriever/lib/datasets.py
+++ b/retriever/lib/datasets.py
@@ -1,0 +1,30 @@
+from retriever import SCRIPT_LIST
+
+
+def datasets(arg_script=None):
+    """Return list of all available datasets."""
+    script_list = SCRIPT_LIST()
+
+    all_scripts = []
+
+    for script in script_list:
+        if script.name:
+            if arg_script is not None:
+                script_name = script.title + "\nName: " + script.name + "\n"
+                if script.keywords:
+                    script_name += "Keywords: " + \
+                        str([tag for tag in script.keywords]) + "\n"
+                not_found = 0
+                for term in arg_script:
+                    if script_name.lower().find(term.lower()) == -1:
+                        not_found = 1
+                        break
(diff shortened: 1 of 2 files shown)


henrykironde:
why are you creating a new module? 
datasets.py

ShivamNegi:
The CLI will call the functions from the python interface. I am not sure about its placement so made a module. Where should I put this function?

henrykironde:
So in this case, how would the call from python look.
That's to say; 
If I were to use retriever in one of my projects as a library/package, 
How would the import statements of the dataset() function look like after the final python interface implementation?

ShivamNegi:
As you can see it currently looks like 
```python
>>>import retriever.lib.datasets as retls
>>>retls.datasets()
```
I was thinking of adding all such function imports in the ```__init__.py```. So that later the import will be simpler.
```python
>>>from retriever.lib import datasets
>>>datasets()
```

henrykironde:
Yes imports should be simpler.
I think the final usage may be something like, 

```python
import retriever as rp
rp.dataset()
 
```

ShivamNegi:
I am thinking of adding all functions under the `datasets` module first and then working on making the import simpler. Cause I believe changing the way imports are being done will be another separate issue to work on.

henrykironde:
Ok

retriever/lib/datasets.py
```diff
@@ -0,0 +1,30 @@
+from retriever import SCRIPT_LIST
+
+
+def datasets(arg_script=None):
+    """Function retruns list of all available datasets."""
```
henrykironde: returns

henrykironde: why do we need this parameter `arg_script=None`

ShivamNegi: Changes made.

PR #918, opened 2017-05-30
Cleaning up unecessary imports

ShivamNegi:
There are some unused imports in respective python file. I have tried to clean this up. @henrykironde


henrykironde:
Did you run test locally?

PR #825, opened 2017-03-06
Refactoring to avoid cyclic dependency

kunalpal:
Hi. Earlier, the well-used functions `MODULE_LIST` and `SCRIPT_LIST` were in `__init__.py` which caused cyclic dependencies throughout other imports (e.g. `lib/compile.py -> lib/templates.py -> __init__.py -> lib/compile.py`). Since, these functions are used everywhere it seemed reasonable to refactor them (into `lib/scripts.py`) to avoid these problems. As suggested in PR #814, I am creating this separate PR for better review and seamless merging.

PR #920, opened 2017-06-08
Removing Cyclic Redundancy

ShivamNegi:
Taking inputs from #825 . This commit works on removing cyclic redundancies. @henrykironde please review.

### DIFF
--- a/retriever/engines/msaccess.py
+++ b/retriever/engines/msaccess.py
@@ -1,8 +1,9 @@
 from __future__ import print_function
-from builtins import str
 import os
+from builtins import str
 from retriever.lib.models import Engine, no_cleanup
-from retriever import DATA_DIR, current_platform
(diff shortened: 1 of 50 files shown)


retriever/engines/msaccess.py
```diff
@@ -1,8 +1,8 @@
-from __future__ import print_function
```
henrykironde: from __future__ import print_function
This is needed for p3

ShivamNegi: what is p3?

ethanwhite: Python 3, thought I think there's a typo and @henrykironde meant Python 2 in this case. Without the import lines that use print like a function `print(...)` will error in Python 2.

ShivamNegi: updated.

PR #921, opened 2017-06-09
Altered Update function for package

ShivamNegi:
`_` has been added to functions that are used only internally. There is no requirement for them in the package. This will required when the #920 gets resolved and retriever package is given a flat look.

### DIFF
--- a/retriever/lib/repository.py
+++ b/retriever/lib/repository.py
@@ -12,19 +12,13 @@ from pkg_resources import parse_version
 from retriever.lib.defaults import REPOSITORY, SCRIPT_WRITE_PATH, HOME_DIR
 from retriever.lib.models import file_exists
 
-global abort, executable_name
-abort = False
-executable_name = "retriever"
 
-
-def download_from_repository(filepath, newpath, repo=REPOSITORY):
+def _download_from_repository(filepath, newpath, repo=REPOSITORY):
     """Downloads the latest version of a file from the repository."""
     try:
-        filename = filepath.split('/')[-1]
         urllib.request.urlretrieve(repo + filepath, newpath)
     except:
         raise
-        pass
 
 
 def check_for_updates():
@@ -49,7 +43,7 @@ def check_for_updates():
         if not os.path.isdir(SCRIPT_WRITE_PATH):
             os.makedirs(SCRIPT_WRITE_PATH)
 
-        update_progressbar(0.0 / float(total_script_count))
+        _update_progressbar(0.0 / float(total_script_count))
         for index, script in enumerate(scripts):
             script_name = script[0]
             if len(script) > 1:
@@ -59,7 +53,7 @@ def check_for_updates():
 
             path_script_name = os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name))
             if not file_exists(path_script_name):
-                download_from_repository("scripts/" + script_name,
+                _download_from_repository("scripts/" + script_name,
                                          os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
 
             need_to_download = False
@@ -73,19 +67,18 @@ def check_for_updates():
             if need_to_download:
                 try:
                     os.remove(os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name)))
-                    download_from_repository("scripts/" + script_name,
+                    _download_from_repository("scripts/" + script_name,
                                              os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
                 except Exception as e:
                     print(e)
                     pass
-            update_progressbar(float(index + 1) / float(total_script_count))
+            _update_progressbar(float(index + 1) / float(total_script_count))
     except:
         raise
-        return
     print("\nThe retriever is up-to-date")
 
 
-def update_progressbar(progress):
+def _update_progressbar(progress):
     """Show progressbar
     Takes a number between 0 and 1 to indicate progress from 0 to 100%.
     And set the bar_length according to the console size


retriever/__main__.py
```diff
@@ -63,7 +63,10 @@ def main():
             return
 
         if args.command == 'update':
-            check_for_updates()
+            if update_scripts():
+                print("\nThe retriever is up-to-date")
+            else:
+                raise
```
ethanwhite: Having this defensive logic here is a little confusing to read since `update_scripts()` is doing something not just checking something. My initial thought would be to keep this logic inside `update_scripts()` and have it `raise` or `print` from there. Is there a reason to do it this way instead?

I'm also not sure this will actually execute as expected since I think the `raise` in `update_scripts` will prevent the `return False` from executing (see my comment below).

retriever/lib/repository.py
```diff
@@ -12,80 +12,96 @@
...
-        scripts = []
-        print("Downloading scripts...")
-        for line in version_file:
-            scripts.append(line.decode().strip('\n').split(','))
+            scripts = []
+            print("Downloading scripts...")
+            for line in version_file:
+                scripts.append(line.decode().strip('\n').split(','))
+        else:
+            scripts = script_names
```
ethanwhite: This approach works well, but I'll mention one alternative in case you like it better. If we name the optional argument `scripts` instead of `script_names` and we set it's default value to `[]` instead of `None` then we could use something like this:

```
if not scripts:
    version_file = urllib.request.urlopen(REPOSITORY + "version.txt")
    version_file.readline()
    print("Downloading scripts...")
    for line in version_file:
        scripts.append(line.decode().strip('\n').split(','))
````

This lets us drop a couple of lines and also helps it read a little more directly as "If there are no scripts provided, use the list in the version file". I don't think it's a big improvement, but just something for you to consider in case you thing it reads better.

File: retriever/lib/repository.py [hunk 3]
```diff
@@ -12,80 +12,96 @@
...
                     pass
-            update_progressbar(float(index + 1) / float(total_script_count))
+            _update_progressbar(float(index + 1) / float(total_script_count))
+
+        return True
     except:
         raise
-        return
-    print("\nThe retriever is up-to-date")
+        return False
```
ethanwhite: Is this `return` ever going to execute? I think the `raise` will prevent that return from ever executing.

ethanwhite:
A couple of additional thoughts:

1. Can you update the commit message to be a little more explicit. I really like [Chris Beam's post on writing good commit messages](https://chris.beams.io/posts/git-commit/) if you're looking for something good to read in this area.
2. I don't think the we have any tests that currently cover the script updating code. It's not super obvious how to test that code, but since you're working on it this might be a good time to think about whether there's a good way to add a couple of tests.

retriever/lib/repository.py
ShivamNegi: Yes, this surely reads better.

henrykironde:
@ShivamNegi, This PR needs to be separated. 
One PR could be for making protected functions, and the other could be refactoring and change of name in the update function `def update_scripts(script_names=None):`.

 **Pep8**
Sometimes we need to compromise pep8 rules to have better readability. In the case of `repository.py` changes, it looks like some changes would be read better the way they were initially. For Example 

```python
path_script_name = os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name))
```
**vs** 

```python
path_script_name = os.path.normpath(os.path.join(HOME_DIR,
                                                             "scripts",
                                                             script_name))
```


Also the docstring could be rewritten, for example 

    """Check and update scripts in the script home directory.

    If script_name is None, update the script home directory, else 
    update the only the requested scripts in script_name.
    """

The function uses `script_names=None` and I want to assume you are considering parsing a list as argument to this function.
Also, if we assume a script name(string) as the argument(Not list),  we may fail when it comes to this part 


```python
        else:
            scripts = script_names

        total_script_count = len(scripts)
```

I think we developed a function to get the version file and return a list with the version numbers. 
Check the repo or lets refactor this to create that function

```
            version_file = urllib.request.urlopen(REPOSITORY + "version.txt")
            version_file.readline()

            scripts = []
            print("Downloading scripts...")
            for line in version_file:
                scripts...........

```



```python
def update_scripts(script_names=None):

    try:
        if script_names is None: # testing for None depends on the type of object(collections, class,.., ) it can be tricky. Find out what applies to your case.
            
        else:
            scripts = script_names
        total_script_count = len(scripts)
        .
        .
        .
        
        # return True  # we do not return if we are raising 
    except:
        raise
    
    return True
```

henrykironde:
Lets  us try to use a few lines for each statement.
If we have several lines of code to go through, the more compact the code, the better it becomes to follow. 
Python focuses on minimizing the lines of code but keeping disjointed statements on separate lines.  A case like this 
```python
                if parse_version(str(script_version)) > parse_version(m):
                    need_to_download = True
                else:
                    need_to_download = False
```
becomes

```python
need_to_download = parse_version(str(script_version)) > parse_version(m)

```
```python
if x == 1: print 'one'
```
becomes

```python
if x == 1:
    print 'one'
```

Good reads would be 

1. http://python-guide-pt-br.readthedocs.io/en/latest/writing/style/
2. Writing Idiomatic Python Book


**Here are some of the examples from the PR:**
- In the first example, I used Pycharm to auto refactor code.
- In the second example, I pasted a code reformatted by excessive application of pep8 using `autopep8`. The code looks really neat but requires more emphasis when reading.

`$autopep8 --in-place --aggressive --aggressive retriever/lib/repository.py `
- In the third example, I used simple pep8
`$autopep8 --in-place retriever/lib/repository.py `

_Results_
1.
```python
        _update_progressbar(0.0 / float(total_script_count))
        for index, script in enumerate(scripts):
            script_name = script[0]
            if len(script) > 1:
                script_version = script[1]
            else:
                script_version = None

            path_script_name = os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name))
            if not file_exists(path_script_name):
                _download_from_repository("scripts/" + script_name,
                                          os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))

            need_to_download = False
            try:
                file_object, pathname, desc = imp.find_module(''.join(script_name.split('.')[:-1]), [SCRIPT_WRITE_PATH])
                new_module = imp.load_module(script_name, file_object, pathname, desc)
                m = str(new_module.SCRIPT.version)

                if parse_version(str(script_version)) > parse_version(m):
                    need_to_download = True
                else:
                    need_to_download = False
            except:
                pass
            if need_to_download:
                try:
                    os.remove(os.path.normpath(os.path.join(HOME_DIR, "scripts", script_name)))
                    _download_from_repository("scripts/" + script_name,
                                              os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
                except Exception as e:
                    print(e)
                    pass
            _update_progressbar(float(index + 1) / float(total_script_count))
```

2.
```python
_update_progressbar(0.0 / float(total_script_count))
        for index, script in enumerate(scripts):
            script_name = script[0]
            if len(script) > 1:
                script_version = script[1]
            else:
                script_version = None

            path_script_name = os.path.normpath(
                os.path.join(HOME_DIR, "scripts", script_name))
            if not file_exists(path_script_name):
                _download_from_repository(
                    "scripts/" + script_name,
                    os.path.normpath(
                        os.path.join(
                            SCRIPT_WRITE_PATH,
                            script_name)))

            need_to_download = False
            try:
                file_object, pathname, desc = imp.find_module(
                    ''.join(script_name.split('.')[:-1]), [SCRIPT_WRITE_PATH])
                new_module = imp.load_module(
                    script_name, file_object, pathname, desc)
                m = str(new_module.SCRIPT.version)

                if parse_version(str(script_version)) > parse_version(m):
                    need_to_download = True
                else:
                    need_to_download = False
            except:
                pass
            if need_to_download:
                try:
                    os.remove(
                        os.path.normpath(
                            os.path.join(
                                HOME_DIR,
                                "scripts",
                                script_name)))
                    _download_from_repository(
                        "scripts/" + script_name,
                        os.path.normpath(
                            os.path.join(
                                SCRIPT_WRITE_PATH,
                                script_name)))
                except Exception as e:
                    print(e)
                    pass
            _update_progressbar(float(index + 1) / float(total_script_count))
```
3.
```python
        _update_progressbar(0.0 / float(total_script_count))
        for index, script in enumerate(scripts):
            script_name = script[0]
            if len(script) > 1:
                script_version = script[1]
            else:
                script_version = None

            path_script_name = os.path.normpath(
                os.path.join(HOME_DIR, "scripts", script_name))
            if not file_exists(path_script_name):
                _download_from_repository(
                    "scripts/" + script_name, os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))

            need_to_download = False
            try:
                file_object, pathname, desc = imp.find_module(
                    ''.join(script_name.split('.')[:-1]), [SCRIPT_WRITE_PATH])
                new_module = imp.load_module(
                    script_name, file_object, pathname, desc)
                m = str(new_module.SCRIPT.version)

                if parse_version(str(script_version)) > parse_version(m):
                    need_to_download = True
                else:
                    need_to_download = False
            except:
                pass
            if need_to_download:
                try:
                    os.remove(os.path.normpath(os.path.join(
                        HOME_DIR, "scripts",       script_name)))
                    _download_from_repository(
                        "scripts/" + script_name, os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
                except Exception as e:
                    print(e)
                    pass
            _update_progressbar(float(index + 1) / float(total_script_count))
```

ShivamNegi:
@ethanwhite the function must returns a boolean value because in the interface I think it will be a necessary feature when trying to debugging and updating specific scripts. The `return` below the `raise` function has been removed as spotted by you.

Currently there are three states of the function
 - `return True` when scripts get updated
 - `return False` when `script_names` parameter doesn't match the available scripts (a spelling mistake in script name)
 - Error raise, like in case of `urlopen` error.

@henrykironde, I have tweaked the doc-string for better understanding. The first result of indentation that you have shown in your comment above looks best to me. But as I can see there are no hard and fast rules to follow. The indentation changes that you want to make in this PR are most welcome.

About the test for `update_scripts`, we could make test which calls the function for updating all the scripts and then tallies the script version number against the version numbers of the repository.

ethanwhite:
> the function must returns a boolean value because in the interface I think it will be a necessary feature when trying to debugging and updating specific scripts

The desirability of this approach varies stylistically a bit among languages. It is commonly used in some languages but is generally seen as not very Pythonic. It's OK to use it if we really have to but raising informative errors is typically preferred in these cases. E.g.,

* https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python
* http://python.net/~goodger/projects/pycon/2007/idiomatic/handout.html#eafp-vs-lbyl
* https://blogs.msdn.microsoft.com/pythonengineering/2016/06/29/idiomatic-python-eafp-versus-lbyl/
* https://stackoverflow.com/questions/1630706/best-practice-in-python-for-return-value-on-error-vs-success

That's not to say there aren't cases where returning True/False isn't useful, but I'm not (yet) convinced that this is one of those cases. This feels like a Try/Except to me. At a first glance I would use the code supplied in my original comment at the top of this function (to set the values in `scripts` and then use an `except` where we actually try to load the script (possible the `except` currently on [Line 85 in this PR](https://github.com/ShivamNegi/retriever/blob/6f828a845d92fea11a37568640b52688a15a52b6/retriever/lib/repository.py#L85)) to throw an informative error message including the script name.

I also don't think we want to do the comparison against the downloaded script list because it ignores custom scripts and non-updated states of the scripts directory. Trying and catching errors automatically tests against the scripts that are currently present including any custom scripts.

ShivamNegi:
Thanks for the references, some important lessons in there. After reading those I think the implementation with `try/except` works better. The functionality that I was hoping to add was selectively updating scripts, which doesn't seem to go along well. The code requires input to be in a 2d array format along with the version number which won't be convenient at all.
I will just push for protected functions, and code refactoring in this PR.

ethanwhite:
>  The functionality that I was hoping to add was selectively updating scripts, which doesn't seem to go along well. The code requires input to be in a 2d array format along with the version number which won't be convenient at all. I will just push for protected functions, and code refactoring in this PR.

Sounds good and the changes look good to me.

I like the idea of being able to update individual scripts, but I think that's something we could put on the To Do list to look at once the basic interface is up and running.

PR #923, opened 2017-06-10
Moving dummy connection and cursor code out to a shared module

goelakash:


retriever/lib/dummy.py
```diff
@@ -0,0 +1,23 @@
+"""Re-usable dummy classes for connectio-less engine instances
```
ethanwhite: There's a typo `connection` not `connectio`, we can probably skip the "Re-usable", and no hyphen in "connectionless" gives:

`Dummy connection classes for connectionless engine instances`

PR #925, opened 2017-06-11
Moving dummy connection and cursor code out to a shared module

goelakash:
Did a `git push :test-ci` to delete old branch and re-push the amended commit, thinking that the PR would pick up on that. But I guess it [only happens for forced pushes](https://stackoverflow.com/a/32818657/3285138) (`git push -f`). 
Same as #923


ethanwhite:
Perfect. Thanks!

Issue #927, opened 2017-06-11
Pre-commit hook preventing new script additions?


ethanwhite:
I've been in the process of cleaning up #821 and am running into an issue with the pre-commit hook:

```
ethan@gandalf:~/RetrieverCode (pr/821 *+)$ git commit --amend 
remote: Counting objects: 11, done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 11 (delta 8), reused 11 (delta 8), pack-reused 0
Unpacking objects: 100% (11/11), done.
From https://github.com/weecology/retriever
 * branch            master     -> FETCH_HEAD
Traceback (most recent call last):
  File ".git/hooks/pre-commit", line 62, in <module>
    if not script_current_version > get_script_version(file):
  File ".git/hooks/pre-commit", line 35, in get_script_version
    return LooseVersion(version_dict[file_name])
KeyError: 'predicts.py'
```

I haven't looked into this, but a quick guess from the error message is that we may be looking up a version number for the script, but if it's a new script it isn't in the associated dictionary and so it errors.

zhangcandrew:
So there's a part of our code where we check changes with a diff filter that looks for added and modified files but not deleted ones. I don't remember why I had it check to make sure we included added files in the git diff, but I made some comments to @henrykironde 's branch that should fix it up.

zhangcandrew:
Looks like it's fixed for now. Closing the issue. Re-open if necessary

PR #928, opened 2017-06-12
Add a check for new scripts

henrykironde:
New scripts do not have to compare versions
The new scripts are considered updated by default
Fixes: #927

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -10,12 +10,8 @@ import sys
 import subprocess
 import urllib.request
 
-
 pattern = re.compile(r'[."]version[\'"\s":=]+(\d+\.\d+.\d+)')
 version_dict = {}
-version_file = urllib.request.urlopen(
-    "https://raw.githubusercontent.com/weecology/retriever/master/version.txt")
-version_file.readline()
 
 
 def to_string(value):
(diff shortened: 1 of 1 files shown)


henrykironde:
Most of the changes are refactoring. The main logic change is,
`if os.path.basename(file) in version_dict.keys():  # New scripts are not checked for version changes`.


We use the file paths changed in `scripts_changed`,
get the base name and check to see if it is available in the dictionary version_dict.keys

henrykironde:
This is really great `git diff --cached --name-only --diff-filter=aMd scripts`, I don't seem to get the part of the permissions. Could be explain more, thanks.

henrykironde:
@zhangcandrew,  It turns out that we already have that set.

```
staged_scripts = to_string(subprocess.check_output(
    "git diff --cached --name-only --diff-filter=AMd scripts", shell=True))
```
And I think it works fine. I may be overlooking something. let me know

zhangcandrew:
Hi Henry, so currently we have 
```
git diff --cached --name-only --diff-filter=AMd scripts
```
The "A" in the diff filter is capitalized meaning the diff filter will also include those files that are changed. I think what we want is 
```
git diff --cached --name-only --diff-filter=aMd scripts
```
Where the "a" in the diff filter is lowercase. This tells the diff-filter to exclude any files that are added, and I think that's what we want.

zhangcandrew:
And in regards to the permission change, so everytime we run ```python setup.py install```, for some reason it says that  our hooks/pre-commit get modified. I think the setup might be modifying the permissions of the pre-commit to something else but I'm not sure because looking into the file itself, nothing seems to be changed. See pictures:

image

So in the above picture, we see that the branch is clear and there are no changes. Then after finishing ```python setup.py install```:

image

henrykironde:
What does the diff have?

zhangcandrew:
Hey Henry, I've put up a PR with some updated logic. Let me know if you have any questions on that one.

henrykironde:
clossing ref: #929

PR #935, opened 2017-06-16
Use parameter binding for insert statements

### LINKED ISSUE #168: SQL INSERT statements use custom quoting rather than parameter binding

ethanwhite:
This is the correct way to handle passing values to databases. It
automatically handles escaping, improves security, and is often faster.

The automatic escaping allows us to remove several chunks of
fragile and confusing escaping code.

Closes #168

### DIFF
--- a/retriever/engines/csvengine.py
+++ b/retriever/engines/csvengine.py
@@ -53,9 +53,13 @@ class engine(Engine):
         """Write a line to the output file"""
         self.output_file.writerows(statement)
 
+    def executemany(self, statement, values, commit=True):
+        """Write a line to the output file"""
+        self.output_file.writerows(statement)
+
     def format_insert_value(self, value, datatype):
         """Formats a value for an insert statement"""
-        v = Engine.format_insert_value(self, value, datatype, escape=False, processed=True)
+        v = Engine.format_insert_value(self, value, datatype)
         if v == 'null':
             return ""
         try:
--- a/retriever/engines/jsonengine.py
+++ b/retriever/engines/jsonengine.py
@@ -67,9 +67,13 @@ class engine(Engine):
         """Write a line to the output file"""
         self.output_file.writelines(statement)
 
+    def executemany(self, statement, values, commit=True):
+        """Write a line to the output file"""
+        self.output_file.writelines(statement)
+
     def format_insert_value(self, value, datatype):
         """Formats a value for an insert statement"""
-        v = Engine.format_insert_value(self, value, datatype, escape=False, processed=True)
+        v = Engine.format_insert_value(self, value, datatype)
         if v == 'null':
             return ""
         try:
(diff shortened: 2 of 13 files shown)


ethanwhite:
Tests are all passing (locally at least). It would be worth testing out a few other datasets using both flat file and database formats to confirm that this didn't break any edge cases.

henrykironde:
I think mysql parameter should be ?

henrykironde:
Not sure but will do some tests later

ethanwhite:
> I think mysql parameter should be ?

Yeah, I thought so originally too, but couldn't get to it work and then looked at the docs again.

> Not sure but will do some tests later

Looks like the tests are failing on CI so no need to worry about this one until I get that sorted out.

ethanwhite:
Looks like I broke something on Python 2 (hooray for CI!). I'll take a look when I get the chance and update later.

ethanwhite:
Looks like it's just MySQL on Python 2, so some sort of lovely edge case. It fails when trying to insert the first line of `mammal-masses` (to provide an example). First line is the insert statement, second line is the first of the sets of values to be inserted by `executemany`, and then the error message:

```
Creating table mammal_masses.MammalMasses...
INSERT INTO mammal_masses.MammalMasses (continent, status, sporder, family, genus, species, log_mass_g, comb_mass_g, reference) VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)
[u'AF', u'extant', u'Artiodactyla', u'Bovidae', u'Addax', u'nasomaculatus', '4.85', '70000.3', u'60']
INSERT INTO mammal_masses.MammalMasses (continent, status, sporder, family, genus, species, log_mass_g, comb_mass_g, reference) VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)
(1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '4.85'', b''70000.3'', '60'),('AF', 'extant', 'Artiodactyla', 'Bovidae', 'Aepycer' at line 1")
```

The error message has `b''70000.3''` instead of `b"70000.3"`. Not sure if this is the source of the issue or just a display quick.

ethanwhite:
Apparently in some edge case in Python 2 passing a string version of a double through pymysql doesn't work properly, but it works fine in Python 3. I can't actually reproduce this in a simplified case even using the exact data it's failing on, but by switching to passing the actual double instead of the `str()` of the double everything appears to be working.

When someone gets the chance this is now ready for review.

PR #942, opened 2017-06-20
Initialize directory function added

ShivamNegi:


henrykironde:
Change this line https://github.com/ShivamNegi/retriever/blob/dfce8c372595f5d8da16ce77244af39b3418579e/scripts/npn.py#L19  to `retriever` spelling mistake.

retriever/lib/tools.py
```diff
@@ -10,26 +10,49 @@
...
 from builtins import str
 from builtins import input
+import csv
 import difflib
+import io
+import json
 import os
+import platform
+import shutil
 import sys
```
henrykironde: Could you explain why we are removing these imports. Are they coming in from previous imports?

ShivamNegi: The Imports haven't been removed they have just been reordered.

PR #944, opened 2017-06-21
Code Refactoring, debug parameter

ShivamNegi:


ShivamNegi:
Reducing one extra `debug` parameter in all `download` function calls by moving it into `engine`.  Please review.

ethanwhite:
We should test this against `2.0.0` to make sure the modified scripts don't cause issues in 2.0.0.

ethanwhite:
This was merged into `master` instead of the development branch. It did break 2.0.0. Reverting.

ethanwhite:
Just to explain what happened here (see also #947 and #948). In `2.0.0` Line 177 in `__main__.py` will continue to pass `debug=debug` to the `download` methods in the `.py` scripts, but the new scripts created in this PR don't expect that argument, so they error.

@ShivamNegi - this was a good refactor in concept, it just happens to cause issues with backward compatibility because of how we handle providing scripts. If we want to keep something like this we could do so by replacing `debug=False` in the `download` methods with `**kwargs`. This would allow the `debug` variable (and another other keyword arguments) to be passed through.

ShivamNegi:
True, I didn't think about backward compatibility. Replacing with `**kwargs` will work.
Now that you mentioned the interface update breaks backward compatibility with a lot of scripts. This being due to the relative import changes.

ethanwhite:
@ShivamNegi - don't worry about the interface addition causing issues just yet. Build it the best way possible and then we can address the backward compatibility issues in one of two ways: 1) We can decide that we need to break backward compatibility and move to 3.0 for a release at the end of the summer; or 2) we can work around it in the scripts in question. Feel free to submit a new PR using the `**kwargs` approach and we'll get that one tested and merged.

henrykironde:
There will be some changes that will affect the scripts files and even the compiled json files. such changes will have to be taken care of.
I will talk to you about that but for now, those changes can be put in the python interface branch.

I have gone through the code and the changes will not affect 2.0 if tailored well. 
Lets continue adding these changes to the interface, I will walk you through them tonight

PR #946, opened 2017-06-22
Restructure commits

ShivamNegi:
Merged commits of python-interface branch and master.

### DIFF
--- a/retriever/__init__.py
+++ b/retriever/__init__.py
@@ -4,33 +4,9 @@ This package contains a framework for creating and running scripts designed to
 download published ecological data, and store the data in a database.
 
 """
-from __future__ import print_function
-from __future__ import absolute_import
-
-import os
-import platform
-
 from .lib import *
-from retriever.lib.defaults import HOME_DIR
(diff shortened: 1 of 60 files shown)


ShivamNegi:
There are some import errors due to newly added scripts in `master`. Also import error in `test_modified_scripts.py`. Correcting those.

retriever/__init__.py
```diff
@@ -4,224 +4,9 @@
...
-    proxies = ["https_proxy", "http_proxy", "ftp_proxy",
-               "HTTP_PROXY", "HTTPS_PROXY", "FTP_PROXY"]
-    for proxy in proxies:
-        if os.getenv(proxy):
-            if len(os.environ[proxy]) != 0:
-                for i in proxies:
-                    os.environ[i] = os.environ[proxy]
-                break
 
+create_dir()
```
henrykironde: We should rename this to something like `create_retriever_home` or `create_home_dir`

ShivamNegi: Yes, `create_home_dir` sounds good to me.

PR #950, opened 2017-06-26
Read connection information from DBMS connection files

ethanwhite:
Implements reading connnection information from MySQL .my.cnf and
documents the use of these connection files for both MySQL and postgreSQL.

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -241,6 +241,39 @@ The ``citation`` command show the citation for the retriever and for the scripts
 
 **To create new, edit, delete scripts please read the documentation on scripts**
 
+
+Storing database connection details
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


docs/introduction.rst
```diff
@@ -241,6 +241,35 @@ The ``citation`` command show the citation for the retriever and for the scripts
...
+
+For MySQL, create or modify `~/.my.cnf`. This is a file named `.my.cnf` located
+in the users home directory. The relevant portion of this file for the retriever
+is the `client` section which should take the general form:
+
+::
+
+   [client]
+   host=localhost
+   port=3306
```
henrykironde: I think we should change it to something like

```
host=localhost(yourhost if not localhost)
port=3306(yourport for MySQL)
```

ethanwhite: Good call. I just changed them to match the same wording as for postgres and added an equivalent paragraph describing what to do with them.

PR #951, opened 2017-06-26
Restructure commits #2

ShivamNegi:


henrykironde:
@ShivamNegi, Could you resolve the conflict

Issue #899, opened 2017-04-17
Store and provide license information


ethanwhite:
We should have a license field for each script which stores the license information and should make it possible for users to obtain that information using commands like:

```
retriever license datasetname
```

and

```
retriever ls -l MIT GPL
```

jainamritanshu:
@ethanwhite I want to work on this. Just a few clarifications. The basic idea to tackle this I can get is something like - 
- Adding a field `license` in the .json scripts
- adding a function `license` in datapackage.py 
```
def license(json_file):
    try:
        contents = json.load(
            open(os.path.join(HOME_DIR, 'scripts', json_file), 'r'))
    except (IOError,OSError):
        print("Script not found.")
        return

    if is_empty(contents['license'])
    	print ("\nNo license specified")
    	
    print ("\nMIT GPL")
```

Adding an argument clause of `license` in `__main__.py` and we will be good to go I guess. Kindly tell me if I am missing something.

ethanwhite:
That's the basic idea.

The `datapackage.py` function's purpose is to get a license from the user creating a new JSON script, so it should read a license values using `clean_input` and assign it to the appropriate field to be written in the JSON.

The description of the license(s) should follow the datapackage spec. See the description under Properties -> Optional Properties -> Licenses at https://specs.frictionlessdata.io/data-package/#properties. It requires a link and we should also include the option license name and read that name for displaying when `retriever license` is called.

This might also require modifying the behavior of `retirever ls -l` so that it can match on licenses. I'm not sure if it will do this by default of will need to be modified.

ethanwhite:
The core of this was implemented in #953. Closing and opening an issue focused on searching.

PR #953, opened 2017-06-28
Add license lookup for datasets

ethanwhite:
Implements a function and CLI for looking up dataset licenses from a
field containing a single license following the `licenses` property
in the frictionless data datapackage spec:
https://specs.frictionlessdata.io/data-package/#licenses

Also adds licenses for some datasets.


ethanwhite:
Implements #899.

I decided to submit this against `python-interface2` so that I couple implement it in both interface and CLI instead of putting that off.

I've tried to design it in such a way that we can easily move `citation` over to the same basic structure and location.

retriever/lib/datasets.py
```diff
@@ -28,3 +35,12 @@ def datasets(arg_script=None):
     all_scripts = sorted(all_scripts, key=lambda s: s.lower())
 
     return all_scripts
+
+def license(dataset):
```
henrykironde: scripts = script_dict()
This could be a function that test if a script is available.
It will be used by both `ls` and `citation` and `license`.
The function could take a string and convert it to lowercase.
An in the function we could handle the errors.


```
    if dataset in scripts:
        script = scripts[dataset]
    else:
        raise UnknownDataset(dataset)
```

ethanwhite: Agreed. It could also be used as a check in `install`, `download`, etc. Are you envisioning a structure like:

```
if dataset_exists(dataset):
    do stuff
```

or we could go with:

```
check_dataset_exists()
do stuff
```

And have the function throw the error?

ethanwhite:
Based on feedback from @henrykironde both here and on Slack I've refactored the PR as follows:

1. Dropped the special error class and now simply thrown and informative `KeyError`
2. Combined error handling and script lookup into a single `get_script` function that takes the name of a dataset as an argument, returns the associate script if found, and otherwise throws an error
3. Instead of returning `'Not currently available'` in the Python interface if a license doesn't exist, return `None`. Handle the information on available in the print statement for the CLI.

@ShivamNegi - I'd welcome a review from you on this as well both to get your suggestions for improvement and so that you can see how I'm thinking about handling printing (via the CLI) vs. returning programmatically usable information (in the Python interface).

henrykironde:
@ethanwhite, Sorry for taking long. I asked @ShivamNegi to make some changes that we could add to this PR and he is working on them. I Will give you feedback when we come up with the changes.

PR #954, opened 2017-06-28
Added download and install function

ShivamNegi:


ShivamNegi:
This is the first look at the download function. There is a need for some more iterations to make this final. One important question in mind is, Do we require print statements?

ShivamNegi:
Print statements like:
```python
>>>import retriever as ret
>>>ret.download("wine-quality")
=> Downloading wine-quality

Downloading winequality-white.csv...
100%  2 seconds 
Downloading winequality-red.csv...
100%  0 seconds Copying winequality-white.csv from /home/username/.retriever/raw_data/wine-quality
Copying winequality-red.csv from /home/username/.retriever/raw_data/wine-quality
```

ethanwhite:
This is an interesting situation with regard to print statements. In most cases I'd say don't print information like this in the Python interface, reserve it for the CLI. However, it is common to provide progress information during long running tasks, primarily using progress bars, so I guess my initial reaction here is to keep the print statements so that folks don't enter a command and then think the program has hung since there is no output and it isn't visibly doing anything.

I'd be interested in hearing what other folks think about this as well.

henrykironde:
Thank you for all the contributions and ideas, I want to say that for the time being we keep the print statements.
Lets work till we bring in the install command and we can then design the print statements well.

However, I think we shall end up with a function like `download(..., cli=FALSE)`.
The interface by default will set `cli` flag to `false`. 
This flag will be used when print is used to either print or not. `def download(dataset, path='./', quite=False, subdir=False, use_cache=False, debug=False, cli=FALSE):`

There are many print statements that we shall have to definitely remove for the case of the interface, then later we can decide about those datasets that take long.

henrykironde:
@ShivamNegi this looks good.
I request you to go ahead and work on the install or what we may call fetch and we add it to this PR.

henrykironde:
@ShivamNegi, for these functions, I guess we should use def install(*argv)
then use get_opts.py
```
from retriever.lib.get_opts import parser
parser.parse_args(argv)
```
This will enable us do what you doing below
```
def install_csv(dataset, table_name=os.path.join(DATA_DIR, '{db}_{table}.csv'), compile=False, debug=False, quite=False, use_cache=False):
    """Install scripts in csv for retriever."""
    args = {
        'command': 'install',
        'compile': compile,
        'dataset': dataset,
        'engine': 'csv',
        'quite': quite,
        'table_name': table_name
    }

```

henrykironde:
Let me know if you want more [reference](https://pymotw.com/2/argparse/) or explanation

ShivamNegi:
The `download_install` function has been added. It is up for further improvement and discussion.
It is using `argparse` as you said. I think using `argparse` here is not every explicit. It makes sure that the user will read the documentation. As the parameters need to be inputted in a specific order.
```python
import retriever
retriever.download_install(False, "install", "csv", "iris")
```

henrykironde:
This part is repeated so we should extract it to a function    

 ```
engine = choose_engine(args)
    engine.use_cache = use_cache

    scripts = name_matches(script_list, args['dataset'])
    if scripts:
        for script in scripts:
            print("=> Downloading", script.name)
            try:
                script.download(engine, debug=debug)
                script.engine.final_cleanup()
            except Exception as e:
                print(e)
                if debug:
                    raise
    else:
        print("The dataset {} isn't currently available in the Retriever".format(args['dataset']))
        print("Run 'retriever ls to see a list of currently available datasets")
```

**Comment by henrykironde** [comment]
I will look at the arguments later to see how we can implement that part

File: retriever/lib/install.py [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+            try:
+                script.download(engine, debug=debug)
+                script.engine.final_cleanup()
+            except Exception as e:
+                print(e)
+                if debug:
+                    raise
+    else:
+        print("The dataset {} isn't currently available in the Retriever".format(args['dataset']))
+        print("Run 'retriever ls to see a list of currently available datasets")
```
henrykironde: Since this is the interface, we should raise some error instead of printing 
```
print("The dataset {} isn't currently available in the Retriever".format(args['dataset']))
 +        print("Run 'retriever ls to see a list of currently available datasets")
```

henrykironde:
In the extracted function, it could be a private function that is like def install(*argv).
The argv in this case will be the args created from 

```
    args = {
        'command': 'install',
        'compile': compile,
        'dataset': dataset,
        'engine': 'csv',
        'quite': quite,
        'table_name': table_name
    }
```

ShivamNegi:
The `compile` option compiles all the available scripts in CLI. Currently I am using the same implementation. 
I am working on a function which would compile specified dataset scripts.

ShivamNegi:
A new `_compile` function has been made which compiles only the specified script.

henrykironde:
Thanks @ShivamNegi, before I review the PR,
`A new _compile function has been made which compiles only the specified script.
`
could you explain more and give the benefits of adding the  new _compile function.

ShivamNegi:
```python
import retriever
retriever.install_csv("iris", compile=True)
```
Earlier this command used to compile all the scripts in retriever. This is the exact way `retriever install --compile csv iris` has been implemented in CLI.
The new `_compile` function compiles only the specified scripts, in this case `iris`.

henrykironde:
I am reviewing this, will update you

retriever/lib/download.py
```diff
@@ -0,0 +1,36 @@
...
+def download(dataset, path='./', quite=False, subdir=False, use_cache=False, debug=False):
+    """Download scripts for retriever."""
+    args = {
+        'dataset': dataset,
+        'command': 'download',
+        'path': path,
+        'subdir': subdir,
+        'quite': quite,
+        'use_cache': use_cache
+    }
```
henrykironde: @ShivamNegi  this code is still replicated in two files. `retriever/lib/download.py` and `retriever/lib/install.py`
Refactor it 
```
 scripts = name_matches(script_list, args['dataset'])
   if scripts:
       for script in scripts:
           print("=> Downloading", script.name)
           try:
               script.download(engine, debug=debug)
               script.engine.final_cleanup()
           except Exception as e:
               print(e)
               if debug:
                   raise
   else:
      .
      .




```

henrykironde:
@ShivamNegi, did you check my latest comment about refactoring?

henrykironde:
@ShivamNegi, looks good so far. We just need some changes.
If you run this code 

```
import retriever as rt
print(rt.datasets("-l birds"))
```

we get 

```
["A database on the life history traits of the Northwest European flora\nName: plant-life-hist-eu\nKeywords: ['plants', 'observational']\n", "A database on visible diurnal spring migration of birds\nName: bird-migration-data\nKeywords: ['Taxon > Plants']\n", 'A stream gage database for evaluating natural and altered flow conditions in the conterminous United States.\nName: streamflow-conditions\n', .....

```

This does not look good, revisit it and refactor as necessary and we need to remove `\n`

In the CLI we have the results below and they look fine 

```
retriever ls -l birds
Available datasets : 8

1. A database on visible diurnal spring migration of birds
Name: bird-migration-data
Keywords: ['Taxon > Plants']

2. Antarctic Site Inventory breeding bird survey data, 1994-2013
Name: antarctic-breed-bird
Keywords: ['birds']

```

ShivamNegi:
I think the `-l` parameter hasn't been implemented properly in the `datasets` function. I have made a new issue regarding this #969 .

PR #956, opened 2017-07-02
Convert sys args to lower case

henrykironde:


henrykironde:
Thanks @ethanwhite, changes are in.

PR #957, opened 2017-07-02
[WIP]: Add compatibility test.

henrykironde:
This makes sure that we have the right scripts and tests
for backward compatibility of the scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,13 @@ before_script:
 after_success:
   - codecov
 
+
+after_script:
+  - pip install retriever -I
+  - retriever -v
+  - retriever ls
+  - py.test test/test_backward_compatibility.py -v
+
 notifications:
   slack:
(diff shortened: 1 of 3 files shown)


henrykironde:
This PR is for a discussion on how we could go ahead with backward compatibility. 
I have tried several approaches and I have included some here

Trying to use `retriever ls` to get the list of datasets, using split, will require further preprocessing since we have some print statements in 2.0.0. (The [byte](https://ci.appveyor.com/project/ethanwhite/retriever/build/1.0.1294/job/ucvhbo9n3cqyc94c#L541)  /  [string](https://ci.appveyor.com/project/ethanwhite/retriever/build/1.0.1294/job/3g6omuvylbh9hk2v#L710) part is okey we have that sorted out)

The best way is to install retriever 2.0.0 using `pip install retriever` and look at the out put after the tests have run. (As shown in the `.yml` files
```
- pip install retriever -I
- retriever -v
- retriever ls

```

However, this approach installs the right version on appveyor but not on Travis.

ethanwhite:
Thanks for getting this started @henrykironde. A couple of initial thoughts:

1. Instead of running `pip install retriever` we could probably do `git checkout v2.0.0 && pip install .`. Not sure if this would work better on Travis.
2. Instead of using `retriever ls` and counting we could just rerun some or all of the tests. This would add a little time to the build, but it might be simpler and give us better information. Alternatively we could identify a subset of tests that are important for this check and just run those using `-k`.
3. Is there a reason to run these in the `after` section or can we have multiple lines in the `script` and `test_script` sections of the .yml files and just include these extra lines there?

PR #960, opened 2017-07-03
Install function for csv, mysql

ShivamNegi:


ShivamNegi:
Two different functions for installing csv and installing mysql have been made. This is due to the sheer number of parameters that are required by each function. 
The CLI takes care of all these parameters using argparse, adding default parameters using the `required_opts` variable present for each engine. 
@henrykironde what are your thoughts on implementing this as a single function?

henrykironde:
Could you add this pr to the download pr

ShivamNegi:
Done. Moved to #954 .

PR #964, opened 2017-07-08
Remove double function definition

ethanwhite:
There are currently two versions of the same function in templates.py
with the second overwriting the first. This removes the first since it
is unused and the functionality is duplicated by the second.

### DIFF
--- a/retriever/lib/templates.py
+++ b/retriever/lib/templates.py
@@ -47,15 +47,6 @@ class Script(object):
         self.engine.db_name = self.name
         self.engine.create_db()
 
-    def reference_url(self):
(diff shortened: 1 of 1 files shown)


ShivamNegi:
I think `reference_url` of `Script` is being inherited in the `HtmlTableTemplate`, `DownloadOnlyTemplate`. In `BasicTextTemplate` it is being overridden. So on first look removing `reference_url` of `Script` doesn't seem right.

ethanwhite:
Ah, you're totally right. Great catch! Thanks for the review @ShivamNegi! Closing.

PR #970, opened 2017-07-15
Added keyword param in datasets

ShivamNegi:

PR #977, opened 2017-07-20
Add missing value

henrykironde:


ethanwhite:
For some reason (and I can't see why this would be) this appears to be causing all of the MySQL based tests, both regression and integration to fail, which is really weird.

ShivamNegi:
In PR #970 I am also facing the exact same tests are failing.

ethanwhite:
I just tested this locally and found some odd behavior. If I run `py.test -v` everything passes as expected. If I run `py.test` without the `-v` I get failures on about half of the regression tests. I can't see why having or not having `-v` would change the results. Can anyone else confirm this behavior?

ethanwhite:
@henrykironde - do you see the weird `-v` difference?

henrykironde:
Yes, its  happens on mac not on windows
After a few dives into this, I think the problem comes with pytest's, parser when parsing the arguments

henrykironde:
instead of running py.test , include a positional arg like path. eg py.test test

ethanwhite:
Yeah - that fixes it as well, but it's still strange since not having an argument works fine on `master` for me.

henrykironde:
Okey, that is strange. I was failing on master with out any other argument  `pytest`

This PR doesn't introduce any core changes in the code base. 
And if there is a unique chance of having a bad Json created, we can test that on two commits. This 1fa529f768d6da93cb71b30458c72e1530d38b62 and one before.  by removing the changes.

ethanwhite:
Just confirmed that e5432e5 passes but 1fa529f fails, so at least on Linux it looks like it's the commit in this PR, but I agree that this is super strange.

ethanwhite:
OK, well that is/was confusing. The errors from running just `py.test` that I'm seeing are being triggered by `test/test_modified_scripts.py`. The trigger, in this case, is the version number change in `breast_cancer_wi.json`. You can just change the version number and nothing else and get the same error and you can also change the version number in any json file and get the same result. If you delete `test/test_modified_scripts.py` everything works normally.
 
With that file removed everything is passing locally and the dataset installs well, so I'm going to go ahead and merge this and open a new issue about the `test/test_modified_scripts.py` problem.

PR #980, opened 2017-07-23
[WIP] Add ability to use existing data packages

ethanwhite:
Downloads Frictionless Data data packages from the web and adds
them to the retriever. Accomplished by:

* Retrieving the JSON metadata file
* Modifying it to work with the retriever
* Writing it to the scripts directory

New packages are added by adding a name-url pair to
scripts/datapackages.yml.

### DIFF
--- a/retriever/lib/get_datapackages.py
+++ b/retriever/lib/get_datapackages.py
@@ -48,6 +48,26 @@ def add_retriever_metadata(dp):
     dp["retriever_minimum_version"] = "2.0.0"
     return dp
 
+def match_types(dp):
+    """Match the datapackage types to the retriever types
+
+    The retriever still describes types using its old system instead of the
+    official frictionless data spec. This converts frictionless data types to
+    retriever types.
+
+    """
+
+    types = {'string': 'char',
+             'number': 'double',
+             'integer': 'int',
+             'date': 'char'
+    }
+
+    for resource in dp['resources']:
+        for field in resource['schema']['fields']:
+            field['type'] = types.get(field['type'], field['type'])
+    return dp
+
 def download_dps_json(dps):
     """Download json files for each external data package"""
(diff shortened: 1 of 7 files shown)


ethanwhite:
The basics of this are working. Running:

```
from retriever.lib.get_datapackages import get_dps, download_dps_json
dps = get_dps()
download_dps_json(dps)
```

Will grab the json data package file, modify it to run with the retriever, and add it to the `~/.retriever/scripts` directory.

If other groups data package files looked exactly like ours we'd just need to hook this code up, but they don't, and I think this is going to mean a fair bit of additional work. E.g., there are lots of fields that are allowed in the json according to the spec that we don't process explicitly in `compile.py`. We currently pass those through unquoted, but many of them are strings and so this leads to errors loading the resulting `.py` scripts. This is the first thing to solve, but I suspect there will be a number of others.

retriever/lib/repository.py
```diff
@@ -26,6 +27,10 @@ def download_from_repository(filepath, newpath, repo=REPOSITORY):
         raise
         pass
 
+def download_external_dps():
```
henrykironde: Some clean up. 
- [ ] more line space before the function.

retriever/lib/get_datapackages.py
```diff
@@ -0,0 +1,96 @@
...
+    official frictionless data spec. This converts frictionless data types to
+    retriever types.
+
+    """
+
+    types = {'string': 'char',
+             'number': 'double',
+             'integer': 'int',
+             'date': 'char'
+    }
```
henrykironde: - [ ]  Closing bracket not indented well
- [ ] Double spacing between functions
- [ ] Do we need the extra libs `urlparse, urlunparse, ParseResult`

zhangcandrew:
Closing this pr now that changes have been incorporated into PR #1010. Feel free to open this PR up again if anything new comes to light!

PR #982, opened 2017-07-23
[WIP]:Revert "Added download and install function"

henrykironde:
Reverts weecology/retriever#954


retriever/lib/install.py
```diff
@@ -1,209 +0,0 @@
...
-                    sys.stderr.write("Failed to load script: %s (%s)\nException: %s \n" % (
-                        dataset, search_path, str(e)))
-
-
-def _install(args, use_cache, debug, compile):
-    """Calling the download-install function."""
-    engine = choose_engine(args)
-    engine.use_cache = use_cache
-
-    if compile:
```
henrykironde: Remove this part of def _compile(dataset):

retriever/lib/install.py
```diff
@@ -1,209 +0,0 @@
...
-import os
-import sys
-from pkg_resources import parse_version
-from retriever.engines import choose_engine
-from retriever.lib.compile import compile_json
-from retriever.lib.defaults import DATA_DIR, SCRIPT_SEARCH_PATHS, VERSION
-from retriever.lib.tools import name_matches
-from retriever.lib.scripts import SCRIPT_LIST
-
-script_list = SCRIPT_LIST()
```
henrykironde: If you print out this script_list in this module, it is empty.
print (script_list)

**_Follow the streps._**

- Install retriever using `pip install . -U` 
- Clean out the scripts using `retriever reset all`. 
- Clean up compiled scripts in working directory of retriever using git add -u (to add your changes) followed by git clean -fx (to clean all unwanted files)

- Then run a simple install 

```
from retriever import install_json
install_json('iris')
```

retriever/lib/install.py
henrykironde: I think we should just use 

scripts = name_matches(SCRIPT_LIST(), args['dataset'])

henrykironde:
@ShivamNegi, lets rewrite this PR. we need to create a PR that will over write from commit  dada58e78ad3e0377e37c87951831df64259b1c4.

These are some of the reasons.

1. After trying to create test using the interface, I realized that we should remove the function def _compile(dataset):

2. problems from import line 13 in `retriever/lib/install.py`
-script_list = SCRIPT_LIST()

retriever/lib/install.py
ShivamNegi: The `_compile` function is being used for compiling the specific scripts. I don't see how the `name_matches` function can replace it.

retriever/lib/install.py
ShivamNegi: Yes, that is right. When there are no available scripts the `install_json` function or any other install function won't work. The user will have to first update the scripts and then run the install functions:
```python
import retriever as rt
rt.check_for_updates()
rt.install_json("iris")
```
I am not sure if this is what you were trying to say.

retriever/lib/install.py
henrykironde: We do not need two function that look the same.  lets not worry about compiling specific scripts because we are still compiling the entire scripts at all times.

retriever/lib/install.py
henrykironde: I think check_for_updates() is supposed to be run by the code. This is true if you use `python setup.py install`. however, it is not true for `pip install . -U`. So if we can change and make sure that if `install` is to run, we need to update or check or check if there are scripts.

PR #984, opened 2017-07-25
Fix testing of modified scripts

### LINKED ISSUE #983: Tests failing (sometimes) due to `test_modified_scripts.py`
When running `py.test` without any arguments (and only without any arguments) some of the regression tests fail, at least on some systems. The failures are being triggered by `test/test_modified_scripts.py`. If there are any scripts with version number changes these failures occur. If you delete `test/test_modified_scripts.py` and keep the version number changes everything works normally.

This also appears to happen, more idiosyncratically, without any changes to the version numbers. I saw these failures on e5432e5eee09bc14261d16871692631eca3ec44f, and they seemed to potentially reflect the issues with `breast_cancer_wi.json` that were fixed in 1fa529f768d6da93cb71b30458c72e1530d38b62. This set of issues is still confusing me. But I'm guessing that once I merged 1fa529f7, `test_modified_scripts.py` detected that `breast_cancer_wi.json` was different than the local version in e5432e, and so tested that script, found the errors due to the `?` noting being declared as a missing value, but for some reason didn't report those issues in it's own test but leaked them into the regression tests someone... maybe?

I'm not sure exactly where the failures are coming from, and it's definitely super weird that they happen with `py.test` but not `py.test -v` or `py.test test`, but `test_modified_scripts.py` is definitely the source of the issues.

Note: I'm pretty sure this is a different set of issues than #978.

@ShivamNegi - any thoughts on what might be going on here. Can you replicate this locally?

@henrykironde - now that I've laid this all out is it consistent with what you're seeing? On current `master`, after merging your PR, the tests all pass locally using `py.test`, but if I checkout e5432e5eee09bc14261d16871692631eca3ec44f then I get test failures in `master`.

ethanwhite:
* Runs `final_cleanup` after each test of the modified script to avoid interfering with other tests
* Moves command line argument code handling so that it operates as expected
* Some code clean up en route to these fixes

Fixes #983.

### DIFF
--- a/test/test_modified_scripts.py
+++ b/test/test_modified_scripts.py
@@ -129,6 +129,7 @@ def install_modified(engine_list=ENGINE_LIST()):
                     print("==>", module.__name__, value.name, "..........", module.SCRIPT.name)
                     try:
                         module.SCRIPT.download(value)
+                        module.SCRIPT.engine.final_cleanup()
                     except KeyboardInterrupt:
                         pass
                     except Exception as e:
--- a/test/test_modified_scripts.py
+++ b/test/test_modified_scripts.py
@@ -18,25 +18,8 @@ from retriever import MODULE_LIST, ENGINE_LIST, ENCODING
 from retriever.lib.tools import get_module_version
 
 reload(sys)
-os_password = ""
 if hasattr(sys, 'setdefaultencoding'):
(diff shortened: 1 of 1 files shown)


ShivamNegi:
Yes, it looks good to me too.

PR #985, opened 2017-07-26
Add download, install function

ShivamNegi:


ShivamNegi:
#982 . Updated with changes.

PR #986, opened 2017-07-26
Scripts as a dictionary and written using json

henrykironde:
Changes the way scripts are represented.
Store the script values in a dictionary and dump them as json

### DIFF
--- a/retriever/lib/tools.py
+++ b/retriever/lib/tools.py
@@ -292,24 +292,28 @@ def sort_csv(filename):
 
 
 def create_file(data, output='output_file'):
-    """Writes a string to a file for use by tests"""
+    """writes lines to a file from a list"""
     output_file = os.path.normpath(output)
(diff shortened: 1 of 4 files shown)


retriever/lib/tools.py
```diff
@@ -292,24 +292,29 @@ def sort_csv(filename):
...
 def create_file(data, output='output_file'):
-    """Writes a string to a file for use by tests"""
+    """writes lines to a file from a list"""
     output_file = os.path.normpath(output)
     with open(output_file, 'w') as testfile:
-        testfile.write(data)
+        for line in data:
+            testfile.write(line)
+            testfile.write("\n")
         testfile.close()
```
ShivamNegi: `testfile.close()` is not require as `with` clause has been used.

henrykironde: 👍thanks @shivam will make changes

PR #989, opened 2017-08-06
Adding the field names to the data package spec

henrykironde:
When testing, test scripts should have the fields given. This enables the engines to export the data to CSV.


ShivamNegi:
Looks good.

PR #990, opened 2017-08-07
Use the interface for regression tests

henrykironde:


ShivamNegi:
I think the PR works well. It is ready to go in. @ethanwhite what are your views?

test/test_regression.py
```diff
@@ -27,6 +36,7 @@
 retriever_root_dir = os.path.abspath(os.path.join(file_location, os.pardir))
 working_script_dir = os.path.abspath(os.path.join(retriever_root_dir, "scripts"))
 HOMEDIR = os.path.expanduser('~')
+script_home = "~/.retriever/scripts"
```
ethanwhite: Is there a reason to use `~` here instead of `HOMEDIR`?

test/test_regression.py
```diff
@@ -47,21 +57,24 @@ def get_script_module(script_name):
     return imp.load_module(script_name, file, pathname, desc)
 
 
-def get_csv_md5(dataset, engines, tmpdir):
+def get_csv_md5(dataset, engines, tmpdir, install_function, cofig):
```
ethanwhite: Is `cofig` is supposed to be `config` as in "configuration"?

test/test_regression.py
```diff
@@ -47,21 +57,24 @@ def get_script_module(script_name):
...
     workdir = tmpdir.mkdtemp()
     workdir.chdir()
+    install_function(dataset.replace("_", "-"), **cofig)
```
ethanwhite: If changing `cofig` -> `config` do it here as well.

test/test_regression.py
ethanwhite: It looks like we're only passing in one engine at a time in which case change `engines` to `engine`.

test/test_regression.py
henrykironde: I didn't see that. will make changes

henrykironde:
@ethanwhite updated

PR #992, opened 2017-08-17
Backward compatibility of scripts

henrykironde:
During development we changed the structure of imports. We did not take care of backward compatibility for retriever 2.0.
This PR finalizes the script changes making the scripts compatible with both 2.0 and current development version.

The change in version numbers is relative to master and not to the changes we have made through the development.
That is to say, We updated the version numbers a few times during development, however I have considered that the development changes should only change once hence reflect once version number change.


scripts/breed_bird_survey.py
```diff
@@ -12,11 +12,16 @@
 import urllib.request, urllib.parse, urllib.error
 import zipfile
 from decimal import Decimal
+try:
+    from retriever.lib.templates import Script
```
ethanwhite: The first line of the try and except blocks looks the same and so I think it can be moved to the main import block.

scripts/amniote_life_hist.py
```diff
@@ -5,7 +5,10 @@
 
 from retriever.lib.models import Table, Cleanup, correct_invalid_value
 from retriever.lib.templates import Script
-from retriever.lib.defaults import VERSION
+try:
+    from retriever import VERSION
+except ImportError:
+    from retriever.lib.defaults import VERSION
```
ethanwhite: Does this script need a version bump? It's the only one without one.

ShivamNegi: `biomass_allometry_db` is also missing a version bump.

henrykironde:
Thanks @ethanwhite and @ShivamNegi, I have made the changes requested.
I think the PR is ready for the next review.

PR #995, opened 2017-08-22
Quite parameter in check_for_updates

ShivamNegi:


ShivamNegi:
Adding `quite` parameters in `check_for_updates`. Just so when writing scripts in python there is no unnecessary console output.

PR #1004, opened 2017-08-30
Update FIA links to new data locations

### LINKED ISSUE #1002: Update FIA links
It's that time of year again where FIA puts out a new release and moves things around. See https://github.com/weecology/retriever/issues/865#issuecomment-325588307

We need to track down the most recent links and update the script.

Thanks to @nestorperez for reporting this.

ethanwhite:
Fixes #1002.

### DIFF
--- a/scripts/forest_inventory_analysis.py
+++ b/scripts/forest_inventory_analysis.py
@@ -20,10 +20,10 @@ class main(Script):
         self.title = "Forest Inventory and Analysis"
         self.name = "forest-inventory-analysis"
         self.retriever_minimum_version = '2.0.dev'
-        self.version = '1.4.0'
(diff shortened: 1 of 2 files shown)


henrykironde:
The links are fine, one problem is that the new server settings seems to be using ciphers that are not handled by any machine running on OpenSSL version 0.9.8
Based on the [SSL Report](https://www.ssllabs.com/ssltest/analyze.html?d=apps.fs.usda.gov) they utilize ECDHE. 

If you are using windows / Linux that is easy to update. If you are using MAC test using 
terminal
$openssl version 
 or 
import ssl
print(ssl.OPENSSL_VERSION)

If you have OpenSSL version 0.9.8 , the download will fail

PR #814, opened 2017-03-02
Avoid creation of intermediate python scripts

kunalpal:
Hi @henrykironde, @ethanwhite. This PR avoids creation of separate python scripts from pre-existing .json files and directly creates `Script` template instances. I have also edited the tests a bit to reflect the changes. The functionality can be verified on any .json file without a compiled .py script. 

Please review the code. Looking forward to your suggestions. If accepted, we can also get rid of the compiled python template files in `scripts/` directory. We can merge the functions `MODULE_LIST` and `SCRIPT_LIST` as well since their functionalities are identical.

PR #1005, opened 2017-09-12
Remove creation of intermediary python scripts

zhangcandrew:
This PR is built off a previously closed PR (#814) by @kunalpal. At the request of @henrykironde, I've added an cmdline argument to debug the json scripts and print out what is being passed on to templates (in compile.py line 177-181) 

You can pull down the PR and test it out by using ls and the --debug flag followed by a json script i.e.:

```
retriever ls --debug iris
```

*Due to the age of the PR, some conflict with the new retriever version so I'm still working on trying to mangle the appropriate changes over to the newer version of retriever. Please feel free to suggest, comment etc on the changes.


henrykironde:
Let's PR to this branch https://github.com/henrykironde/retriever/tree/python-interface2.

zhangcandrew:
Pretty excited about this new PR! There are a lot of changes that I had to make to bring it up to date, but the intermediary python scripts are now gone. This is probably going to take a lot of review, but the items added are: 

-No intermediary python scripts
-debug mode for json templates
-cyclic dependencies removed. (There were a few methods which needed to be moved around for this to happen. The names for these files may no longer make any sense, so we can look into renaming them or something else.)


Since this is such a big overhaul, if you decide to pull it down and test it out, you'll have to uninstall retriever completely (pip uninstall retriever - maybe a few times), and then run `python setup.py install` to bring in the new changes. Looking forward to this one!

henrykironde:
Nice @zhangcandrew, some imports old imports are still used, from a quick look at the Travis fails

zhangcandrew:
Yeah, I saw that haha. My next goal is to clean all that up! In the meantime, feel free to test the functionality and see how you like it!

test/test_regression_cli.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import absolute_import
```
henrykironde: This file should not be there

test/test_integration.py
```diff
@@ -19,8 +19,8 @@
 import pytest
 from retriever.lib.compile import compile_json
 from retriever.lib.defaults import HOME_DIR
-from retriever.engines import engine_list
-from retriever.lib.tools import file_2list
+from retriever import ENGINE_LIST
+from retriever.lib.tools import file_2string
```
henrykironde: current imports should be 
```
from retriever.engines import engine_list
from retriever.lib.tools import file_2list
```

test/test_integration.py
```diff
@@ -361,7 +361,7 @@
 xml_test_parameters = [(test, test['expect_out'])
                        for test in tests if test != csv_latin1_encoding]
 
-file_location = os.path.dirname(os.path.realpath(__file__))
+file_location = os.path.dirname(os.path.realpath(_file))
```
henrykironde: Shouldn't this be `os.path.dirname(os.path.realpath(__file__))`?

test/test_retriever.py
```diff
@@ -414,8 +415,8 @@ def test_add_dialect():
     table['dialect']['dummy_key'] = 'dummy_value'
 
     result = {}
-    result['cleanup'] = 'Cleanup(correct_invalid_value, missing_values=\x00)'
-    result['delimiter'] = "'\t'"
+    result['cleanup'] = Cleanup(correct_invalid_value, nulls='\x00')
```
henrykironde: I think we should keep missing_values? Unless I am missing something

zhangcandrew: Yeah this should be missing_values. That was my bad!

test/test_integration.py
zhangcandrew: In line 62 of compile.py I think the keyword for file was changed from __file__ to _file. I wasn't sure if that was a new change or an outdated one, but if we go with __file__, I can quickly change that over!

zhangcandrew:
Closing now that all the changes have been incorporated into PR #1010

PR #1013, opened 2017-10-23
Repair builds

henrykironde:
The change in the directory structure
and addition of new modules led to failures in builds
This PR adds the paths to the modules required


henrykironde:
@ethanwhite this is ready,

PR #1014, opened 2017-10-26
Update version before release

henrykironde:
Update CHANGES.md with major updates since last release
Update the version number in retriever_installer.iss
Update setup.py
Run version.py

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,6 @@
 
 ## Major changes
 - Add Python interface
-- Adopting the Frictionless Data specification standards
 - Add Retriever to conda
 - Auto complete of Retriever commands on Unix systems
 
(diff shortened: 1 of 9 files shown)


CHANGES.md
```diff
@@ -1,3 +1,55 @@
+# v2.1.0
+
+## Major changes
+- Add Python interface
+- Adopting the Frictionless Data specification standards
```
ethanwhite: Can you clarify the distinction between this and the change in 2.0. Does this mean "Use Frictionless Data specification terminology for internals" or something like that. If so that might be a minor change.

CHANGES.md
```diff
@@ -1,3 +1,55 @@
...
+- Add Retriever to conda
+- Auto complete of Retriever commands on Unix systems
+
+## Minor changes
+
+- Add license to datasets
+- Change the structure of raw data from string to list
+- USe precise on travis from trusty
```
ethanwhite: I don't think we need this line, but if you want to keep it  "USe" -> "Use"

CHANGES.md
```diff
@@ -1,3 +1,55 @@
...
+- Add testing on any modified dataset
+- Improve memory usage in cross-tab processing
+- Add capabilitiy for datasets to use custom Encoding
+- Test regression using python interface
```
ethanwhite: I'd change this to "Use new Python interface for regression testing"

henrykironde:
I have added a few lines for both the interface and CLI to take care of the edge cases that may happen.

Issue #994, opened 2017-08-21
Changing to Read the Docs: sphinx_rtd_theme


ShivamNegi:
I really like this [sphinx theme](http://deeplearning.net/software/theano/index.html). I propose to update our docs to this theme. I am ready to take this up if everyone approves. The repository of the theme is here: [sphinx_rtd_theme](https://github.com/rtfd/sphinx_rtd_theme).

ethanwhite:
:+1:

henrykironde:
I prefer the old classic one  :smile: though it looks outdated.
Having said that, I look forward to having a new theme. :+1:

ethanwhite:
Complete in #1027. Thanks again @ShivamNegi!

PR #1027, opened 2017-12-04
Change theme

ShivamNegi:


ShivamNegi:
#994 , the sphinx-rtd-theme has been added. Checks are failing as the config file has been changed.
This commit puts the new theme in place. Content changes and reordering still needs to be done.

henrykironde:
Thanks shivam, If we rebase the test will run again. You could do it or I can go ahead and remove the conflict.

henrykironde:
remove this 

```
retriever.lib.parse_script_to_json module
-----------------------------------------

.. automodule:: retriever.lib.parse_script_to_json
    :members:
    :undoc-members:
    :show-inheritance:
```
from file retriever.lib.rst

I can not build the scripts there is something happening, I am not sure. Where you able to build.

Also include this package  in requirements 
`sphinx_rtd_theme`

henrykironde:
So the scripts fail when `make html` tries to run `conf.py`.
Just found out that we should run 

```
from retriever.lib.repository import check_for_updates
check_for_updates(False)
```
So that the scripts are updated.
Also I think this line should be 
`reference_link = script.ref` to `reference_link = script.ref.strip()`
```
# get info from the scripts
for script_num, script in enumerate(script_list, start=1):
    if script.ref.strip():
        reference_link = script.ref.strip()
    elif bool(script.urls.values()):
```
Try that @ShivamNegi, good catch Shivam, I would not have realized that

ShivamNegi:
Even after the making the changes and re-installing retriever the `WARNING: Unknown target name` still persists in my system.

```
WARNING: Unknown target name: "ten_years_of_abundance_data_within_a_spatial_population_network_of_the_alpine_butterfly_i_parnassius_smintheus_i".
```

henrykironde:
Okey two things I have noted. 

1. Lets us changes the reference line to "\n| **reference:**    `{}`" . So we use `` for references.

2. The other thing is that the names kind of conflict. I have given you a sample PR #5 to test with. But this is a warning. I think we can put it in later, after PR 1010, since a change of name would cause many changes. We could put in changes handling  "\n| **reference:**    `{}`" then deal with the name change later.

ShivamNegi:
Yes, I get it now. As you suggest I put the references in ` so that there are no conflicts. I have changed the formatting of the page also. Hope you like it.

ethanwhite:
Wow, the new site looks great. Thanks @ShivamNegi!

PR #1030, opened 2017-12-21
Improve error message when a file to be downloaded is missing

### LINKED ISSUE #1003: Improve error message when a file to be downloaded is missing
If I file isn't present at the url when we try to download it the error that is reported isn't very useful:

```
cannot import name 'urlretrieve'
```

We get this in part because of joint support for Python 2/3 (we try the Python 3 call, fall back to Python 2, and so if issue isn't Python 3 vs 2 we get an unhelpful error message), but we need to catch this better and report something more useful to the user.

See, e.g., https://github.com/weecology/retriever/issues/865

apoorvaeternity:
Fixes #1003

### DIFF
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -136,7 +136,7 @@ class Engine(object):
         real_line_length = 0
         for values in lines:
             initial_cols = len(self.table.columns) - \
-                (3 if hasattr(self.table, "ct_names") else 2)
+                           (3 if hasattr(self.table, "ct_names") else 2)
             # add one if auto increment is not set to get the right initial columns
             if not self.table.columns[0][1][0] == "pk-auto":
                 initial_cols += 1
@@ -412,7 +412,7 @@ class Engine(object):
             print("\nDownloading " + filename + "...")
             try:
                 urlretrieve(url, path, reporthook=reporthook)
-            except:
+            except ImportError:
                 # For some urls lacking filenames urlretrieve from the future
                 # package seems to fail. This issue occurred in the PlantTaxonomy
                 # script. If this happens, fall back to the standard Python 2 version.


apoorvaeternity:
@ethanwhite The AppVeyor build is failing.

retriever/__main__.py
```diff
@@ -174,11 +174,12 @@ def main():
                     dataset.engine.final_cleanup()
                 except KeyboardInterrupt:
                     pass
-                except Exception as e:
-                    print(e)
+                except Exception:
```
henrykironde: I think its better if we also (print e)

henrykironde:
@apoorvaeternity thank you for the contribution. Sorry for the late reply.
I have fixed the issue that you reported `The AppVeyor build is failing.`in #1031. Next time you could create an issue. Feel free to review the PR. 

I  have reviewed your work and the main download function is https://github.com/apoorvaeternity/retriever/blob/8c6840654ba4319877c7399ce9de02c379ef8bb6/retriever/lib/engine.py#L407

apoorvaeternity:
@henrykironde Okay so this means I have to make the changes in `engine.py` and not in `__main__.py`. Right?

henrykironde:
Yes

henrykironde:
If you debug using `PyCharm` or `pdb` you will notice that you hit that download function in engine first. In order to test or reproduce the error, I recommend you go to your script directory and change the urls to some wrong url for just one script. For example the Iris dataset,
https://github.com/weecology/retriever/blob/master/scripts/iris.json, is easy to work with.
Change the urls to wrong values and then try to install it `retriever install sqlite iris`. remember before you install the dataset, run `retriever reset all` to clean the environment

apoorvaeternity:
@henrykironde How does it look now?

retriever/lib/engine.py
```diff
@@ -406,19 +406,28 @@ def database_name(self, name=None):
...
             except:
                 # For some urls lacking filenames urlretrieve from the future
                 # package seems to fail. This issue occurred in the PlantTaxonomy
                 # script. If this happens, fall back to the standard Python 2 version.
                 from urllib import urlretrieve as py2urlretrieve
-                py2urlretrieve(url, path, reporthook=reporthook)
+                try:
+                    py2urlretrieve(url, path, reporthook=reporthook)
+                except:
+                    pass
```
henrykironde: Would replacing `print("Unable to download dataset {}".format(filename))`  for pass in line 425 , not be enough to catch the errors?

henrykironde: if you make changes would you also update to current master.
1. update master
`git fetch upstream`
`git merge upstream master`


2. In your current branch
`git rebase master`

apoorvaeternity: If I replace line 425 with ```print("Unable to download dataset {}".format(filename))``` it just prints ```cannot import name 'urlretrieve'``` and not the message.

henrykironde:
Thanks @apoorvaeternity , am testing this in the next 5 hours

apoorvaeternity:
@henrykironde Any updates on this?

zhangcandrew:
Hey @apoorvaeternity!

So I went through and looked at the code, there might be a few things that need to be fixed. Also you might want to update your code base to the retriever repository master branch, since we just put in a huge change on how we deal with scripts (although, I think your PR is not affected much by this, it'd still be good so that you can test changes locally). 

Also, as @henrykironde mentioned, I think some of the boolean values like downloaded are unneeded since there is a flow already present in the try-except blocks. I know you mentioned that some things weren't printing, it might be because your excepts are too generic. Try excepting like this: 

```
except ImportError as e
```
Or something along those lines, since we know the problem with the urlretrieve is an importError. Just changing that locally for me let retriever print out slightly more verbose errors which is what I think #1003 was asking for.

apoorvaeternity:
@zhangcandrew Okay. I will make the changes by today.

apoorvaeternity:
@zhangcandrew You are right. I have made the changes. Just using `except ImportError` prints a better message `HTTP Error 404: Not Found`. I have also rebased on top of master branch. Kindly review.

zhangcandrew:
Thanks!

That seems to do the trick, although I'm curious as to why it is caught by ImportError and not URLError as this documentation suggests it should be: https://docs.python.org/3.0/library/urllib.request.html#urllib.request.urlopen. 

Merging this in for now. @apoorvaeternity, if you find that problems arise because we are excepting importError instead of URLError or some of the other errors mentioned with urllib (HTTPError, etc), feel free to open this again, we can discuss and put in another PR!

PR #1031, opened 2017-12-25
Use BEGIN ... COMMIT

henrykironde:
Mark the start and end of the execution
Rollback if there is an error


henrykironde:
Due to some changes in the image used on  `appveyor`, we needed to change some settings. Mysql 5.7 has some updates.

Issue #1032, opened 2017-12-29
Raise NotImplementedError for methods in base classes that should be overridden in child classes.


goelakash:
For eg., the `engine` class has several 'abstract' methods that are supposed to be overridden in the child classes, however, it is not enforced, as a call to the unimplemented parent class function will `pass`.
More context here: https://stackoverflow.com/questions/44576167/force-child-class-to-override-parents-methods

PR #1033, opened 2018-01-01
new_update

shreyneil:
The pull request is with reference to the issue raised for overriding abstract method of parent class, please verify and get back.
https://github.com/weecology/retriever/issues/1032

### DIFF
--- a/retriever/engines/csvengine.py
+++ b/retriever/engines/csvengine.py
@@ -29,7 +29,8 @@ class engine(Engine):
 
     def create_db(self):
         """Override create_db since there is no database just a CSV file"""
-        return None
+        raise NotImplementedError
+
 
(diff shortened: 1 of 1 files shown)


henrykironde:
Thank you @shreyneil for showing interest in contributing to the data retriever

shreyneil:
sir, what are the base classes and where are they present.
Thank you.

henrykironde:
Bases classes are 
1. engine,py in https://github.com/weecology/retriever/blob/master/retriever/lib/engine.py which is inherited by all engines in the `./retriever/lib`

2. scripts.py which defines the properties of the all the datasets
3. table.py which defines the properties of the tables in the datesets

Others are mainly helper modules.

PR #1034, opened 2018-01-01
Update engine.py

shreyneil:
The above pull request is with reference to the following issue

https://github.com/weecology/retriever/issues/1032

Please verify and get back.

### DIFF
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -10,7 +10,7 @@ from builtins import input
 from builtins import zip
 from builtins import next
 from builtins import str
-from abc import abstractmethod
+from abc import ABCMeta, abstractmethod
 import sys
 import os
 import getpass
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -62,7 +62,7 @@ class Engine(object):
             self._cursor = None
 
     def get_connection(self):
-        """This method should be overidden by specific implementations
+        """This method should be overridden by specific implementations
         of Engine."""
         raise NotImplementedError
 
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -44,7 +44,7 @@ class Engine(object):
     use_cache = True
     debug = False
     warnings = []
-    
+    __metaclass__ = ABCMeta
 
     def connect(self, force_reconnect=False):
         if force_reconnect:
@@ -62,7 +62,8 @@ class Engine(object):
             self.connection.close()
             self._connection = None
             self._cursor = None
-
+    
+    @abstractmethod
     def get_connection(self):
         """This method should be overloaded by specific implementations
         of Engine."""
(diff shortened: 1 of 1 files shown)


shreyneil:
@henrykironde  sir please verify.

retriever/lib/engine.py
```diff
@@ -60,11 +61,12 @@ def disconnect(self):
...
             self._connection = None
             self._cursor = None
-
+    
+    
     def get_connection(self):
         """This method should be overloaded by specific implementations
         of Engine."""
-        pass
+        raise NotImplementedError
```
henrykironde: In Python,  it's not required to implement the abstract methods of the parent class. Unless we really want that particular method to be implemented by any inheriting class, otherwise the previous version was fine.

henrykironde: In other languages you have to implement the abstract methods

shreyneil: So is this pull request not required?

shreyneil: I think this small change can be included for engine.py which might be needed at a later stage.

henrykironde: Based on the comment this should be overridden not overloaded, 
"""This method should be overriddend by specific implementations
        of Engine."""

retriever/lib/engine.py
```diff
@@ -43,6 +43,7 @@ class Engine(object):
     use_cache = True
     debug = False
     warnings = []
+
```
henrykironde: Remove extra space

retriever/lib/engine.py
```diff
@@ -60,11 +61,12 @@ def disconnect(self):
             self.connection.close()
             self._connection = None
             self._cursor = None
-
+
```
henrykironde: remove extra space

henrykironde:
Lets clean this up.  Remove the unwanted spaces.

retriever/lib/engine.py
```diff
@@ -657,7 +657,7 @@ def insert_statement(self, values):
         if self.debug:
             print(insert_stmt)
         return insert_stmt
-
+
```
henrykironde: space causing diff

henrykironde:
The lines that have no changes at all should not be touched. 
In this pr you introduced space causing a diff

henrykironde:
We forgot to change the overloaded to overridden.
We can include that change .

henrykironde:
overridden not overriden not

henrykironde:
And the spaces are still there on lines 46,63 and 660

henrykironde:
looks good 👍 ,I will have more tests on this later and get it going

henrykironde:
you pushed a wrong commit,  0fc7779 this was having the right changes

shreyneil:
I actually copied the original code and made the changes because of the error in spaces. Please verify.

PR #1010, opened 2017-10-18
Removes compiling from json to python scripts.

henrykironde:
1. Removes compiling from json to python scripts.

2. Add new tables for spatial data.
For the retriever to handle Geospatial data

3. Define table for each script before pre-processing.

4. Added the functionaly to handle
compressed/archied datasets by using key word archieved
and the files that are in the zip for each table.

5. Handling data packages registered in the `datapackage.yml` file
I have remove the scripts from this file that seem
not to have followed the standards.
To make this possible, we are matching the major `frictionless`
types, integer,string etc,  with the types in the retriever and falling
back to char for some other types
For example, datetime YYYY-MM-DDThh:mm:ssZ.

Ref: This is a combination of the work from PR #814 #980 #1005.

PR #1042, opened 2018-01-11
Remove Compiling JSON dataset to Python

zhangcandrew:
Removes compiling JSON datasets to Python


zhangcandrew:
I think this is good for review @ethanwhite. We're breaking #1010 down into a few PRs now, and we'll merge them in one by one. All the relevant changes requested in #1010 though should be reflected in this PR. 

This PR deals specifically only with getting rid of the intermediary python scripts made from compiling JSONs

henrykironde:
Forgot to update the docs. Will send in an update

ethanwhite:
Some of the .py scripts use `openfr` and `openfw`. These have now been moved again so we'll have to figure out how to get them to the right option of three choices instead of two.

henrykironde:
@ethanwhite, I don't get the comment above clearly.

ethanwhite:
For example `breed_bird_survey.py` imports `openfr` & `openfw`. It currently tries two different locations for this. By moving these functions from `scripts` to `tools` we need to move where there are imported from, while maintaining compatibility with 2.0 and 2.1.

henrykironde:
Sorry about the imports, It turns out that we missed several things. I have gone ahead to send the updates to Andrew. He may need to do some clean up on the commit since I didn't want to update the script's version number for Mammal-supper-tree twice.(Or we could do a squash merge)

I have intentionally blocked tests for modified scripts since they were taking long.

retriever/lib/scripts.py
```diff
@@ -1,36 +1,71 @@
...
 
     for search_path in [search_path for search_path in SCRIPT_SEARCH_PATHS if exists(search_path)]:
-        to_compile = [
+        data_packages = [
             file for file in os.listdir(search_path) if file[-5:] == ".json" and
             file[0] != "_" and (
                 (not isfile(join(search_path, file[:-5] + '.py'))) or (
                     isfile(join(search_path, file[:-5] + '.py')) and (
                         getmtime(join(search_path, file[:-5] + '.py')) < getmtime(
                             join(search_path, file)))) or force_compile)]
```
ethanwhite: Seems like we can drop most of this pile of conditionals now. I don't think we need either of the `isfile` checks or the `getmtime` check or the `force_compile`. I don't think we're using a leading underscore for anything either, so this whole thing could probably become:

```
data_packages = [file for file in os.listdir(seach_path) if file.endswith(".json")]
```

retriever/lib/scripts.py
```diff
@@ -1,36 +1,71 @@
...
+                    setattr(read_script, "_name", script_name)
+                    modules.append(read_script)
+                    loaded_scripts.append(script_name)
 
         files = [file for file in os.listdir(search_path)
                  if file[-3:] == ".py" and file[0] != "_" and
-                 '#retriever' in ' '.join(open(join(search_path, file), 'r').readlines()[:2]).lower()]
+                 ('#retriever' in
+                  ' '.join(open(join(search_path, file), 'r').readlines()[:2]).lower()
+                  or '# retriever' in
```
ethanwhite: Is there a reason to add this special case instead of just making all of the comments consistent?

retriever/lib/scripts.py
```diff
@@ -39,27 +74,26 @@ def MODULE_LIST(force_compile=False):
...
-                            print("Current version is {}".format(VERSION))
+                        if not check_retriever_minimum_version(new_module):
                             continue
                     # if the script wasn't found in an early search path
                     # make sure it works and then add it
                     new_module.SCRIPT.download
-                    modules.append(new_module)
+                    setattr(new_module.SCRIPT, "_file", os.path.join(search_path, script))
+                    setattr(new_module.SCRIPT, "_name", script_name)
+                    modules.append(new_module.SCRIPT)
```
ethanwhite: We've been dropping the `.SCRIPT` elsewhere, do we still want it here.

retriever/lib/table.py
```diff
@@ -10,24 +10,109 @@
...
+
+        for key in self.schema:
+            if key == "fields":
+                column_list = []
+                for obj in self.schema["fields"]:
+                    type = None
+                    if str(obj["type"]).startswith("pk-") or str(obj["type"]).startswith("ct-"):
+                        type = obj["type"]
+                    else:
+                        type = spec_data_types.get(obj["type"],"char")
```
ethanwhite: Space after `,`

ethanwhite:
This new round of changes looks good and I think the nested try-except is a good solution to the moved imports.

retriever/lib/load_json.py
```diff
@@ -0,0 +1,68 @@
...
+def read_json(json_file, debug=False):
+    """Read Json dataset package files"""
+    json_object = OrderedDict()
+    json_file = str(json_file) + ".json"
+
+    try:
+        json_object = json.load(open(json_file, "r"))
+    except ValueError:
+        pass
+    if type(json_object) is dict and "resources" in json_object.keys():
```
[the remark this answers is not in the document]
henrykironde: If load fails, `json_object` is an empty dictionary.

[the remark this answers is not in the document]
ethanwhite: Ah, OK, thanks. :+1:

ethanwhite:
This looks great. A big thanks to @kunalpal, @zhangcandrew, and @henrykironde for this major improvement to the internals that should make a lot of things simpler in the future.

PR #1045, opened 2018-01-20
Issue#905 Add updating null values in txt file

MakarovYaroslav:
Two functions have been added for replacing in txt file bad values('-999.00', '-999') by 'None'.

### DIFF
--- a/version.txt
+++ b/version.txt
@@ -54,7 +54,7 @@ nematode_traits.json,1.0.0
 ngreatplains-flowering-dates.json,1.0.0
 npn.py,2.1.2
 nyc_tree_count.json,1.0.0
-pantheria.py,1.3.1
+pantheria.py,1.3.2
 phytoplankton_size.json,1.2.0
 plant_comp_ok.json,1.2.0
 plant_life_hist_eu.py,1.4.2
--- a/scripts/pantheria.py
+++ b/scripts/pantheria.py
@@ -16,7 +16,7 @@ class main(Script):
         self.title = "Pantheria (Jones et al. 2009)"
         self.name = "pantheria"
         self.retriever_minimum_version = '2.0.dev'
-        self.version = '1.3.1'
+        self.version = '1.3.2'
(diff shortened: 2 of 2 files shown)


henrykironde:
@MakarovYaroslav thanks for the contribution. I do like what you have done with the code 👍.
On the other hand, the retriever has the capabilities of handling that internally.

Sorry the issue was not clear, but this can be done with identifying the null values used in this dataset and adding them to the list in the part shown below.

```
            self.cleanup_func_table = Cleanup(correct_invalid_value, nulls=['NA', 'more-values'])
        else:
            self.cleanup_func_table = Cleanup(correct_invalid_value, missing_values=['NA', ' [more-values]'])
```
Note: 
**Why the issue was created**: The dataset seemed to use more than one value to represent null values, `NA` being one of them. This is based on the last time I went through the dataset.

henrykironde:
Looks good just a few changes.  set the version number to `1.3.2`
run `Python version.py ` to update the `version.txt` and commit all changes

MakarovYaroslav:
Done. Thanks for explanation @henrysenyondo. Can you reply to my message in gitter, please?

henrykironde:
@MakarovYaroslavm sorry for the late reply. I read it and forgot about it. Thanks for the reminder.

version.txt
```diff
@@ -53,7 +54,7 @@ nematode_traits.json,1.0.0
 ngreatplains-flowering-dates.json,1.0.0
 npn.py,2.1.2
 nyc_tree_count.json,1.0.0
-pantheria.py,1.3.1
+pantheria.py,1.3.2
```
henrykironde: Some info should not be in this PR. Looks like you have some scripts that have been added in the version.txt.
Clean up the repo and run python version.py again.

You should also squash the commits to one. 

```
git clean -fx -d
retriever reset scripts
python version.txt
```

henrykironde:
@MakarovYaroslav, all looks good. Thanks for checking those values

PR #1049, opened 2018-01-25
Add download_files_from_archive parameters

henrykironde:


henrykironde:
This PR makes sure that we use all the parameters of the download_files_from_archive() as part of automatic install of archived files

ethanwhite:
:+1:

PR #1050, opened 2018-01-25
[WIP]:Another example of archived dataset

henrykironde:


henrykironde:
@shreyneil this is an example of the json file nd_gain. 
We should add all the files as resources. The resources is a list that defines each table. 
The format is as shown below

```
  "resources": [
    {
      "path": "resources/trends/gain.csv",
      "dialect": {
       "delimiter": ",",
      "header_rows": 1
      },
      "name": "trends_gain",
      "schema": {
        "fields": [
          {
            "name": "IS03",
            "type": "char"
          },.......
        ]
      },
      "url": "https://gain-new.crc.nd.edu/assets/gain/files/resources-2017-03-10-14h04.zip"
    },
    {
      "path": "resources/gain/gain.csv",
      "dialect": {
        "delimiter": ",",
         "header_rows": 1
      },
      "name": "gain_gain",
      "schema": {
        "fields": [
            {
              "name": "ISO3",
              "type": "char"
            },
            {
              "name": "year_Name",
              "type": "char"
            },
            
            {
              "name": "year_2008",
              "type": "char"
            },....
        ]
```

henrykironde:
Note: There are some edge cases.
Some files have headers as integers(years)
For example in the File 
`resources/gain/gain.csv` the headers are 
`"ISO3","Name","1995","1996","1997","1998","1999","2000","2001","2002","2003","2004","2005","2006","2007","2008","2009","2010","2011","2012","2013","2014","2015","2016"`
So we need to rename these header values.

In the example above I have renamed the values as 
```
           "name": "year_2008",
              "type": "char"

```

The type in this example is wrong so you could provide the right types for all the fields. 
Most of the fields type are double. 

image

shreyneil:
Done sir, Please review!

PR #1054, opened 2018-01-31
Update version and improve downloads on archived datasets

henrykironde:
Solve backward compatibility caused by updated  feature 
download from archives

### DIFF
--- a/scripts/gentry_forest_transects.py
+++ b/scripts/gentry_forest_transects.py
@@ -29,7 +29,7 @@ class main(Script):
         self.title = "Alwyn H. Gentry Forest Transect Dataset"
         self.name = "gentry-forest-transects"
         self.retriever_minimum_version = '2.0.dev'
-        self.version = '1.4.1'
+        self.version = '1.4.2'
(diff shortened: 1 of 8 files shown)


scripts/gentry_forest_transects.py
```diff
@@ -62,7 +62,8 @@ def download(self, engine=None, debug=False):
         local_zip.close()
         self.engine.download_files_from_archive(self.urls["stems"], filelist)
 
-        filelist = [os.path.basename(filename) for filename in filelist]
+        if parse_version(VERSION).__str__() < parse_version("2.1.dev").__str__():
+            filelist = [os.path.basename(filename) for filename in filelist]
```
henrykironde: For older versions, we need to use the file name,  not its relative path.

henrykironde:
@ethanwhite, I think this will solve the nd-gain dataset installation errors

henrykironde:
This is weird that it is now failing.  Let me check it again

ethanwhite:
Looks great @henrykironde! Sorry for the delay in getting to the review.

PR #1057, opened 2018-01-31
Adding and editing files at : scripts/

ss-is-master-chief:
Added a respective JSON to the `amniote_life_hist.py` and `aquatic_animal_excretion.json` under scripts
@ethanwhite @henrykironde Please support with your views on this.


henrykironde:
Create a new PR for
`Delete datasets.rst `

ss-is-master-chief:
@henrykironde  I have issued a PR Sir

henrykironde:
This is great. I wanted to have these converted to Json, however the old retriver versions are using the old scripts. If we convert to the new json version, then there is no backward compatibility for these datasets(scripts) on the old retriever versions like 1.8 and 2.0. 

In fact the new scripts we make now will not support the old version hence we have to include the version as `retriever_minimum_version = '2.1.dev'`

henrykironde:
I Will close this for now. we can open it sometime if need  be

ss-is-master-chief:
That is one of the doubts I had during making the JSON files. Thank you for the clarification.

PR #1068, opened 2018-02-04
[WIP] Create status_dashboard.py

apoorvaeternity:
Initial script to check whether datasets are installing.


henrykironde:
Good start. For us  to make sure that it runs well. Lets us create the list of dataset to run as examples.
`for dataset in  ['bird-size', 'mammal-masses', 'airports']:`
checkout the regression tests. We use engine_obj.to_csv()
to export installed dataset to csv so that we calculate the hash.

So instead of looping through datase_names, we can get the script objects.
```
for script_obj in dataset():
  If script_obj.name in ['bird-size', 'mammal-masses', 'airports']:
  Do install the datset and export to csv
  Then get getmd5()
```

apoorvaeternity:
Okay. So I have to first install the dataset using `install_sqlite` and then export it to csv and then check md5 of the csv file.Right?
Do I have to compare the md5 with something?

henrykironde:
`Okay. So I have to first install the dataset using install_sqlite and then export it to csv and then check md5 of the csv file.Right? `correct
Do I have to compare the md5 with something? I will re calculate them after

apoorvaeternity:
@henrykironde When I am using `engine_obj.to_csv()` and the dataset contains multiple tables it converts only one table to csv leaving the rest.

henrykironde:
The current design of the retriever gives us the script's list. If you look at the to_csv function. I guess there are some small changes that could be implemented to make sure that all files are exported. 
```
    def to_csv(self):
        # Due to Cyclic imports we can not move this import to the top
        from retriever.lib.engine_tools import sort_csv
        for _ in list(self.script.urls.keys()):
            table_name = self.table_name()
```
From what I do see, we may need to change the for loop to consider tables names instead of urls keys. Each Script object has at least an instance of a table. Try playing around with that and see if it will work.

apoorvaeternity:
The url keys are considered as table names. The loop is executed for every table name but self.table_name() returns the same table name in each iteration so the same csv is written in each iteration resulting in just one csv in the end. I have fixed it in #1070 .

henrykironde:
I was just looking at  you PR. But I have a meeting will get back to it later

apoorvaeternity:
@henrykironde I am able to print the md5 of the generated csvs. Please review.

henrykironde:
So if you look at the regression tests, we have something like 
```
download_md5 = [
    ('mt-st-helens-veg', 'd5782e07241cb3fe9f5b2e1bb804a794'),
    ('bird-size', '45c7507ae945868c71b5179f7682ea9c'),
    ('mammal-masses', 'b54b80d0d1959bdea0bb8a59b70fa871')
]

db_md5 = [
    # ('mt_st_helens_veg', '0'),
    ('bird_size', '98dcfdca19d729c90ee1c6db5221b775'),
    ('mammal_masses', '6fec0fc63007a4040d9bbc5cfcd9953e')
```
We should have a list of constant pre-calculated md5 hashes like in regression tests.
Then we should have a part that check if the exported files have the same hash as that registered.

Also when exporting, you should export to a directory or temp directory, so that `getmd5` can be given the directory of the file to calculate the md5 of all files in that directory. 

And we can test for change in the hash, assert oldhash == new harsh



Is the design clear ?

apoorvaeternity:
Okay so initially I have to select some datasets and one by one convert them to csv and find the md5sum of those files and then store them in a list for matching.

Then when the script is run I have to create a directory and convert a particular dataset to csv and find md5 using `def getmd5(data, data_type='dir'):` and match it with the list of hash created earlier.
Right?

henrykironde:
Yes

apoorvaeternity:
@henrykironde Please review.

henrykironde:
Looks good. It would be nice if you install from outside the directory that you want to export csv to
```python

for script in datasets():
    if script.name in example_datasets:
        install_sqlite(script.name, use_cache=False)
        workdir = mkdtemp(dir=file_location)
        os.chdir(workdir)
        engine_obj = script.checkengine(sqlite_engine)
        engine_obj.to_csv()
         os.chdir change back to where you started
        os.remove("sqlite.db")
        current_md5 = getmd5(workdir , data_type='dir')
        rmtree(workdir)
        assert current_md5 == precalculated_md5[script.name]
```

henrykironde:
Some thing like that. we should be able to make it as a test case to run on travis and appveyor. So that we know that it works.

apoorvaeternity:
We have calculated the md5 but what about those datasets which get updated regularly like `airports` .
It's md5 changed while I was working.

henrykironde:
That is the reason for this project, to track down what changed what is failing and what is installing correctly. And display the results on the dashboard.

apoorvaeternity:
@henrykironde Okay. So how should I go further with this?

henrykironde:
Let us add elton-trail dataset to the test and edit the script by adding `"dialect": {"do_not_bulk_insert": "True"},` to each of the resources. And calculate the hash value.

status_dashboard.py
```diff
@@ -0,0 +1,33 @@
...
+        engine_obj = script.checkengine(sqlite_engine)
+        engine_obj.to_csv()
+        current_md5 = getmd5(os.getcwd(), data_type='dir')
+        os.chdir(file_location)
+        os.remove("new_sqlite.db")
+        rmtree(workdir)
+        if current_md5 == precalculated_md5[script.name]:
+            print('Passed')
+        else:
+            print("Current md5 of {} is {} expected {}".format(script.name, current_md5, precalculated_md5[script.name]))
```
henrykironde: If a script's md5 sum is not the same could we add some function that tries to identify why we are having different md5 hashes. 
This could be because the files did not get downloaded or they are missing(url 404 error ) Or the files changed by adding more data or renaming.

henrykironde: Refer to the `test_modified script` to see how we install and make sure that there are no errors. If there are error we should be able to identify the errors.

apoorvaeternity:
@henrykironde So I was doing some research to see how to find when a dataset is modified. I found that we can easily get the `Last-Modified` date for different resources of a dataset in the response headers.
response

`Last-Modified` is available in response headers in the majority of datasets.
I think in this way we can check if the dataset has changed even before downloading it.
Is this approach reliable?

henrykironde:
@apoorvaeternity I think from my recollection we tried that and found out that not all responses provided that info. Feel free to add that logic if you have done some more research on that.

apoorvaeternity:
@henrykironde I checked all datasets and found that only 10-11(including those that are no longer present at their urls or giving request errors) are not returning that info. So I think it will decrease the processing to a greater extent. If we don't find the `Last-Modified` attribute we can go for further checks for a dataset.  I will try this logic in my further enhancements to this script 👍

apoorvaeternity:
@henrykironde Please take a look.

henrykironde:
Looks like you getting there. :+1:

apoorvaeternity:
@henrykironde  Yeah. I am pretty close to creating a basic version of the script but it still needs a lot of improvements and I also have to optimize the code.

apoorvaeternity:
@henrykironde  The failing tests of #1071 show that the expected md5 value of `elton-traits` dataset is `d355a18ac7d26e213ff2971f98cda648` but my script is calculating it to be `9da42cd158074d68eb464e8c7b37d3ee`. For other datasets it is matching. Please have a look.

henrykironde:
@apoorvaeternity could you drop elton-traits and use either portal, community-abundance-misc, plant-comp-ok, forest-plots-michigan, croche-vegetation-data.

apoorvaeternity:
@henrykironde Okay 👍

henrykironde:
for example 
```
        {
            "name": "0",
            "type": "char"
          },
          {
            "name": "M",
            "type": "char"
          },
          {
            "name": "",
            "type": "char"
          },
          {
            "name": "",
            "type": "char"
          },
          {
            "name": "",
            "type": "char"
          },
          {
            "name": "4",
            "type": "char"
          },
          {
            "name": "5700",
            "type": "char"
          },
```
what where did you get these fields?

henrykironde:
Sorry for the previous/deleted comment, it does not apply to this PR

scripts/elton_traits.json
```diff
@@ -10,25 +10,25 @@
     "name": "elton-traits",
```
henrykironde: Remove changes to this file since you are not using it.

status_dashboard.py
```diff
@@ -0,0 +1,97 @@
...
+import urllib.request
+import json
+from datetime import datetime
+
+mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, csv_engine, download_engine, json_engine, xml_engine = engine_list
+
+file_location = os.path.dirname(os.path.realpath(__file__))
+
+example_datasets = ['bird-size', 'mammal-masses', 'airports', 'portal']
+precalculated_md5 = {'airports': 'f16a0bb21d347e2ba180cde26f35c87a',
```
henrykironde: Do we need these hash numbers if you already created a separate file for that kind of data?

status_dashboard.py
```diff
@@ -0,0 +1,97 @@
...
+                     'bird-size': '98dcfdca19d729c90ee1c6db5221b775',
+                     'mammal-masses': '6fec0fc63007a4040d9bbc5cfcd9953e',
+                     'portal': 'aaa2b66a3a77efe416e3555c444f4e2e'}
+
+
+def check_urls_exist(dataset):
+    """Check to see if all resources of a dataset
```
henrykironde: `def check_urls_exist(dataset):` should be refactored. 
when I read the name I feel like a url is given and return a value based on the check.

status_dashboard.py
```diff
@@ -0,0 +1,97 @@
...
+        request.get_method = lambda: 'HEAD'
+        try:
+            urllib.request.urlopen(request)
+        except urllib.request.HTTPError as e:
+            failed[url_key] = (url, e.code)
+    return failed
+
+
+def get_last_modified(dataset):
+    """Returns a dictionary of tables with the last modified date as
```
henrykironde:
```
Returns a dictionary of tables with the last modified date as
 values for particular dataset and None if it cannot find it.
```

I don't understand what you are trying to say

apoorvaeternity: It returns a dictionary of particular resources of a dataset with their last modified dates.
For eg. consider airports dataset then it will return something like
```python 
{'regions':python datetime,'navaids':python datetime ...... ,....... ,.....}
```

status_dashboard.py
apoorvaeternity: @henrykironde What should I rename it to? `check_resources`?

status_dashboard.py
apoorvaeternity: @henrykironde No we don't need it now.

status_dashboard.py
apoorvaeternity: @henrykironde I will update the comments so it becomes clearer what each of the functions are doing.

apoorvaeternity:
@henrykironde I have improved the comments. If anything is still unclear I will explain it. Right now the script contains a few functions to perform different operations like finding the md5 of dataset, finding last modified dates of resources, checking whether the urls for a resource exist or not and the reason for it if it doesn't.
These function will help in different ways in checking whether the dataset is installing properly or not.
Right now I have worked on the basic structure of the script. The script still needs lots of improvements so that the errors encountered can be caught in a better way so that we may get closer to the reasons when a dataset fails to install.

henrykironde:
That is the error I get, @apoorvaeternity

apoorvaeternity:
@henrykironde The script is running fine for me. I am unable to reproduce the issue. I have just updated the md5 of airports dataset in the dataset_details.json  using the same script. Can you please try again.
Try to install airports using `retriever install sqite airports` and then by running the script and tell me what happens.

zhangcandrew:
@apoorvaeternity, works fine for me on python3.5. 

Henry is testing with python2.7, I reproduced his error by running with python2.7, so that's where the problem is probably occurring! Code for retriever should be compatible with both versions of python. 

You should be able to reproduce this if you have python2.7 installed and run:
 `python2.7 ./status_dashboard.py`

apoorvaeternity:
@henrykironde @zhangcandrew This will get fixed with #1097 . I think we will have to run this script in both `python 2` and `python 3` environments.

apoorvaeternity:
There is a `difflib` module in python which we can use for finding diffs.

Here is an example-----
```
file1.csv
Name,Sno,Marks
Apoorva,1,70
Henry,2,80
Ethan,3,90
Ironman,4,95
```
```
file2.csv
Name,Sno,Marks
Apoorva Pandey,1,70
Henry Senyondo,2,80
Ethan White,3,90
```
Here's the code:
```python
from difflib import HtmlDiff
a=HtmlDiff()
with open('file1.csv','r') as file1, open('file2.csv','r') as file2,open('diff.html','w') as file3:
    file3.writelines(a.make_file(file1,file2))
```

This is the HTML file that is generated.
screenshot from 2018-03-01 14-10-13

henrykironde:
I like that :+1:

apoorvaeternity:
@henrykironde Okay!!! I will incorporate this into the script.

apoorvaeternity:
@henrykironde Why is the build failing? :confused:

henrykironde:
squash the commits to one and rebase with master

apoorvaeternity:
@henrykironde I think I will have to copy the code of `getmd5` function from `engine_tools.py` to the script since `getmd5` is not available outside the retriever directory. This is required to make the script standalone.

henrykironde:
@apoorvaeternity, I thought of something that may help. Since we want to know and show that the data is changed,  we could use two scripts, i.e airports and a copyofairports. The only difference in the scripts would be adding `"dialect": {"do_not_bulk_insert": "True"},` and `"dialect": {"do_not_bulk_insert": "False"}.`

This would cause the change in data installed and we could be able to show diff. You can also stick to the design we discussed of creating a script that would cause a change in the data.

apoorvaeternity:
@henrykironde I was just wondering how are we going to show the diff? After first installing the dataset then converting it to csv to find the diff or before installing it? Because there may be cases where a dataset fails to install and it cannot be converted to csv but still we want to show the diff.

henrykironde:
@apoorvaeternity, great question. That is why we need a dashboard. The dashboard will tell us is a dataset actually install. If it does not install we do not have to test the diff. Its like a nested if statement if you think about it as code. 

In a normal run, 

```python
 for each dataset:

 	# Test if it installs fine
 	try to install
 	  	install dataset
 	if no error on installing
 		update the dataset icon that show install status
 	else:
 	   update other items with something like 
 	   grey status to show no active results
 	   *CONTINUE to the other dataset

 	# Test if md5 is diffent 
 	Since we continue on datasets that fail to install
 	this step is only run by those dataset that install fine

 	# getting diff
 	This will only be used by those datasets that show a different md5
 	from the pre calculated one 
```

henrykironde:
` I was just wondering how are we going to show the diff? After first installing the dataset then converting it to csv to find the diff or before installing it?` we could decide to note the hash of the raw data, and process it too. I think the work flow will be the same. May be more processing needed. Since we always download the file raw, we could also add those checks to trace changes in the raw file. (note: For compressed/ zipped sources, we only need to test the file that is extracted to be installed )

apoorvaeternity:
In the above commit I have created a sample-dataset json file and added it to .retriever/scripts folder programatically then I have modified the url of the table and installed again to find the diff if the md5 of the dataset does not match with previous value.

zhangcandrew:
Hey @apoorvaeternity, just tested this out again, status_dashboard.py is still failing for me with python2.7. 
image

apoorvaeternity:
@zhangcandrew  I fixed the issue in #1106  but forgot to rebase the current branch onto master. I have rebased it now. I just checked it on python2.7 and it's working now. If the error still persists for you please let me know.

zhangcandrew:
Awesome! Looks like it's working well now! 

Looks like something else just popped up too: 
image


Although I'm guessing by the error and the nature of this branch, that would be your dashboard functioning correctly and displaying that the files aren't being downloaded correctly (?)

apoorvaeternity:
@zhangcandrew Working fine for me. HTTP Error 503 is for service unavailable.
photo tester

dashboard_script_tester.py has been created to check the status_dashboard.py script. For this I have stored two versions of a dataset on github one the original version and the modified version to check the script. see http://github.com/apoorvaeternity/sample-dataset .
Though right now I can't seem to figure out the reason for this error. :thinking:

henrykironde:
Works for me, yet to find what to expect or what it had actually done. If you could be explain a little on what to run.

apoorvaeternity:
@henrykironde @zhangcandrew The `dashboard_script_tester.py` script has been created to create the scenario that we will deal with when actually checking the datasets. I have stored two versions of a dataset at http://github.com/apoorvaeternity/sample-dataset an original version and a modified version. This is to create the scenario of changes in a dataset. Running `dashboard_script_tester.py` performs the following operations.
1. Creates a sample-dataset.json script in .retriever/scripts so that it can be used to install a sample dataset using functions such as `install_sqlite()`
2. Firstly the script creates a directory named `original`. Here we store the csvs of original dataset.
3. Then the script updates the resource url to the modified version in sample-dataset.json.
4. Now the script computes the md5 value of the dataset which will now get changed because now we are installing using the modified version.
5. So now the script will create a `modified` directory to store csvs of the dataset which has been modified and a `diffs` directory that will contain the html diff file which will get created by comparing the csvs in `original` and `modified` directory.

If you run the script you will see that these directories will get created.

zhangcandrew:
I see! Ok, looks like it's looking good to me, I was able to open up the html file and everything. I was wondering if it was possible for you to format the generated html files to only show the lines which have been modified. 

Since we'll be working with large datasets, if everything gets shown, we end up with html with millions of lines from the dataset whereas we might only be interested in the 3 or 4 lines that were changed.

apoorvaeternity:
@zhangcandrew Yeah I had discussions about the diff with @henrykironde . I am trying to find other ways of finding diffs so that it renders properly.
A possible way is by using `git` and `ansi2html`.
`git diff --color-words --no-index ../retriever/original/sample_dataset_FL_insurance_sample.csv ../retriever/modified/sample_dataset_FL_insurance_sample.csv | ansi2html  > new_diff.html`

It renders like this---
screenshot from 2018-03-28 21-28-24
 Though it doesn't look very good. The file that is generated is a lot smaller compared to previous way.

henrykironde:
Do not use git. From what I learned, git wont be able to handle huge files. 
try diff with -u  or comm. read through the two manuals to see what options best represent the data

Here are some of the examples I have run
```
diff -u original modified
diff -u original modified > output
diff -a --suppress-common-lines -y original modified > output
cat original | sd 'cat modified' > output
```
also try this, I have not run this.
https://gist.github.com/stopyoukid/5888146

Test and see what would work best

apoorvaeternity:
@henrykironde  `Do not use git. From what I learned, git wont be able to handle huge files.` What does huge mean here like the size of the file or the no. of lines? What size are we actually targetting?

henrykironde:
A good read https://githubengineering.com/how-we-made-diff-pages-3x-faster/

henrykironde:
@apoorvaeternity a good dataset that we shall have to use and test would be vertnet-birds or any vertnet-*.
I talked to Ethan and he does not think that the git way will be the best. I guess we can just utilize these commands and wrap the output into html tags
```
diff -u original modified
diff -u original modified > output
diff -a --suppress-common-lines -y original modified > output
cat original | sd 'cat modified' > output
```
else try this  https://github.com/google/diff-match-patch

apoorvaeternity:
@henrykironde Okay for using vertnet-birds I will have to create two versions of the dataset and store somewhere. Yeah, I will not use git. I was just trying out different ways.
Finding diff is a very expensive task. The more information we want in a diff the more time it will take. So we have to do tradeoffs between information we want and speed.
After going through several StackOverflow answers and blogs I think `diff` is the fastest.
diff also has a command `diff --speed-large-files file1 file2` for large files.

One thing, the output that diff produces is not colored so it becomes difficult to see the changes so I used a wrapper named [colordiff](https://www.colordiff.org/) to color the diff.
After trying out different git commands and using colordiff this is the best output that I was able to generate using the command
`diff -C 1 retriever/original/sample_dataset_FL_insurance_sample.csv retriever/modified/sample_dataset_FL_insurance_sample.csv | colordiff`
screenshot from 2018-03-29 21-36-37

apoorvaeternity:
@henrykironde Yeah, I have either changed a single value or deleted entire lines.

apoorvaeternity:
I think initially I should work with a small file and note down the changes that I have made and see if it is working correctly i.e showing the differences correctly and then move on to larger files.

apoorvaeternity:
@henrykironde Okay. I will use portal main as the sample-dataset.

henrykironde:
Let me know when you have made the changes. It will be also good if we go ahead and clean this PR and rebase it to the current master

apoorvaeternity:
@henrykironde  I have made the changes. One thing, the order of rows generated when using to_csv function is different from the original csvs which were used for installing the dataset. So when finding diff we are comparing two version of csvs generated using to_csv and not with the actual csv which was used for installing dataset.

dashboard_script_tester.py
```diff
@@ -0,0 +1,94 @@
+from retriever import install_sqlite
+from retriever import datasets
```
henrykironde: You should avoid writing code more than 10 lines of code.Break down this source in to smaller functions.

Let us talk about the design concept. I think you have all the parts you need but the design is missing

henrykironde: for the to_csv, We should introduce a flag and we should miss your 
```python
def to_csv(self, sort=True):
        # Due to Cyclic imports we can not move this import to the top
        from retriever.lib.engine_tools import sort_csv
        for table_n in list(self.script.tables.keys()):....

 
         if sort is set to True
             sort_csv(csv_file_output)
        self.disconnect()
```

apoorvaeternity: Okay I will improve the design. For the to_csv function shall I make the changes in this PR only or send a new PR for it?

henrykironde:
Send in a PR for the to_csv

henrykironde:
Whenever you get time, ping me on google hangout to discus your PR

henrykironde:
@apoorvaeternity is this ready for testing?

apoorvaeternity:
@henrykironde I will make some updates today and tell you by tomorrow when it's ready for testing.

apoorvaeternity:
@henrykironde You may test the dashboard_script_tester.py.
What to expect?
1. First the dataset will be installed and converted to csv in the `original` directory.
2. Then the url will be programatically changed to the modified version of the dataset.
3. Then the md5 of the dataset will be found out which has now changed so the script will install the modified version of the dataset and convert it to csv in `modified` directory.
4. The diff of the files in `original` and `modified` directory will be found out and saved to `diffs` directory and files from the `modified` directory will be moved to the `original` directory so that they are there for comparison the next time.

henrykironde:
I will run the tests and later you know later today.

apoorvaeternity:
@henrykironde Any updates regarding the PR?

henrykironde:
Done, good so far

henrykironde:
Let us change the names, of the scripts. For example one that creates the environment can be post_ ......, or create_enviro... or setup_envir.... something like that.

henrykironde:
Thanks @apoorvaeternity, what is the sequence of the commands I have to run?

apoorvaeternity:
Just run `python dashboard_script_parallel.py` it will display the time it takes to install the five datasets mentioned in the script on a pool of 3 processes.

henrykironde:
```
henry@henrys-IMac retriever (new1068) $ python dashboard_script_parallel.py
airports
biodiversity-response
bird-size
Downloading 5603108: 0.00B [00:00, ?B/s]
Saving a copy of runways.csv...
Downloading runways.csv: 0.00B [00:00, ?B/s]
Saving a copy of 5599229...
Downloading 5603108: 192kB [00:05, 35.8kB/s]                                                              
Downloading runways.csv: 2.84MB [00:05, 499kB/s]                                                          
Installing biodiversity_response_population_response: 100%|██████████| 761/761 [00:00<00:00, 817.40rows/s]
('Successfully installed==>', u'biodiversity-response')
gdp
Downloading 5599229: 832kB [00:07, 116kB/s]                                                               
Downloading gdp.csv: 448kB [00:01, 345kB/s]                                                               
Installing airports_runways:  38%|███████████▍                  | 15491/40651 [00:05<00:08, 3087.33rows/s]Couldn't create table (database is locked). Trying to continue anyway.
Installing gdp_gdp:   0%|                                                     | 0/11507 [00:00<?, ?rows/s]INSERT INTO bird_size_species (family, species_number, species_name, english_name, subspecies, m_mass, m_mass_n, f_mass, f_mass_n, unsexed_mass, unsexed_mass_n, m_wing, m_wing_n, f_wing, f_wing_n, unsexed_wing, unsexed_wing_n, m_tarsus, m_tarsus_n, f_tarsus, f_tarsus_n, unsexed_tarsus, unsexed_tarsus_n, m_bill, m_bill_n, f_bill, f_bill_n, unsexed_bill, unsexed_bill_n, m_tail, m_tail_n, f_tail, f_tail_n, unsexed_tail, unsexed_tail_n, clutch_size, egg_mass, mating_system, display, resource, refs) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
database is locked
Exception KeyError: KeyError(<weakref at 0x10a19e310; to 'tqdm' at 0x1091dde10>,) in <object repr() failed> ignored
('Successfully installed==>', u'bird-size')
portal
Installing airports_runways:  40%|███████████▉                  | 16200/40651 [00:06<00:08, 2778.85rows/s]INSERT INTO gdp_gdp (country_name, country_code, year, value) VALUES (?, ?, ?, ?)
no such table: gdp_gdp
Exception KeyError: KeyError(<weakref at 0x10a1afaf8; to 'tqdm' at 0x10a188590>,) in <object repr() failed> ignored
('Successfully installed==>', u'gdp')
Installing airports_runways:  66%|███████████████████▋          | 26701/40651 [00:10<00:04, 2953.13rows/s]Couldn't create table (database is locked). Trying to continue anyway.

Saving a copy of 3299474...
Downloading 3299474: 8.00kB [00:01, 7.18kB/s]██████████▎        | 28945/40651 [00:12<00:04, 2617.78rows/s]
Installing portal_plots:   0%|                                                                           | 0/24 [00:00<?, ?rows/s]INSERT INTO portal_plots (plot_id, plot_type) VALUES (?, ?)
no such table: portal_plots
Exception KeyError: KeyError(<weakref at 0x10a19eaa0; to 'tqdm' at 0x10a170850>,) in <object repr() failed> ignored
('Successfully installed==>', u'portal')
Installing airports_runways: 100%|██████████████████████████████| 40651/40651 [00:16<00:00, 2520.08rows/s]
Installing airports_airports: 100%|█████████████████████████████████████████████████████| 53275/53275 [00:18<00:00, 2883.23rows/s]
Installing airports_countries: 100%|████████████████████████████████████████████████████████| 247/247 [00:00<00:00, 1910.40rows/s]
Installing airports_regions: 100%|████████████████████████████████████████████████████████| 3999/3999 [00:00<00:00, 4082.64rows/s]
Installing airports_airport_frequencies: 100%|██████████████████████████████████████████| 28644/28644 [00:03<00:00, 7925.22rows/s]
Installing airports_navaids: 100%|██████████████████████████████████████████████████████| 11117/11117 [00:04<00:00, 2650.43rows/s]
('Successfully installed==>', u'airports')
Time taken=53.126912117 seconds
```

apoorvaeternity:
@henrykironde  This is the output I get.
```
python3 dashboard_script_parallel.py 
biodiversity-response
bird-size
airports

Saving a copy of countries.csv...
Downloading countries.csv: 0.00B [00:00, ?B/s]
Saving a copy of 5603108...
Downloading 5603108: 0.00B [00:00, ?B/s]
Saving a copy of 5599229...
Downloading countries.csv: 24.0kB [00:03, 6.64kB/s]                             
Installing airports_countries: 100%|██████| 247/247 [00:00<00:00, 9324.70rows/s]
Installing airports_airport_frequencies:  16%|▏| 4553/28644 [00:00<00:00, 45527.Installing airports_airport_frequencies:  31%|▎| 9000/28644 [00:00<00:00, 45056.Installing airports_airport_frequencies:  48%|▍| 13694/28644 [00:00<00:00, 45603Installing airports_airport_frequencies:  62%|▌| 17798/28644 [00:00<00:00, 44128Installing airports_airport_frequencies:  79%|▊| 22489/28644 [00:00<00:00, 44925Installing airports_airport_frequencies:  95%|▉| 27156/28644 [00:00<00:00, 45434Installing airports_airport_frequencies: 100%|█| 28644/28644 [00:00<00:00, 45083.83rows/s]
Installing airports_runways: 100%|███| 30313/30313 [00:02<00:00, 14430.82rows/s]
Installing airports_regions: 100%|█████| 3999/3999 [00:00<00:00, 37614.76rows/s]
Installing airports_airports: 100%|██| 53338/53338 [00:03<00:00, 16865.86rows/s]
Downloading 5603108: 192kB [00:13, 14.4kB/s]                                    
Installing biodiversity_response_population_response:   0%| | 0/761 [00:00<?, ?rInstalling biodiversity_response_population_response:  78%|▊| 590/761 [00:00<00:Installing biodiversity_response_population_response: 100%|█| 761/761 [00:00<00:00, 5311.70rows/s]
Downloading 5599229:  11%|█▉                | 88.0k/828k [00:13<02:16, 5.57kB/s]Successfully installed==> biodiversity-response
gdp
Downloading 5599229:  13%|██▍                | 104k/828k [00:14<01:38, 7.54kB/s]
Saving a copy of gdp.csv...
Downloading 5599229:  19%|███▋               | 160k/828k [00:15<00:43, 15.7kB/s]Successfully installed==> airports
portal
Downloading 5599229:  21%|████               | 176k/828k [00:16<00:33, 20.0kB/s]
Saving a copy of 3299483...
Downloading 3299483: 8.00kB [00:03, 2.59kB/s] | 272k/444k [00:05<00:01, 103kB/s]
Downloading gdp.csv: 448kB [00:05, 77.7kB/s]                                    
Installing gdp_gdp: 100%|████████████| 11507/11507 [00:00<00:00, 26330.57rows/s]
Installing portal_main:   2%|▏          | 728/35549 [00:00<00:09, 3742.10rows/s]Successfully installed==> gdp
Installing portal_main: 100%|█████████| 35549/35549 [00:06<00:00, 5092.48rows/s]
Downloading 5599229:  78%|██████████████▊    | 648k/828k [00:28<00:02, 75.7kB/s]Successfully installed==> portal
Downloading 5599229: 832kB [00:30, 27.8kB/s]                                    
Installing bird_size_species: 100%|█████| 3769/3769 [00:00<00:00, 4055.45rows/s]
Successfully installed==> bird-size
Time taken=32.838395833969116 seconds
```

ethanwhite:
I get the same error as @henrykironde on Python 2.7.14 and a clean run on Python 3.6.4, so probably a Python 2 vs 3 issue?

apoorvaeternity:
@ethanwhite  I get a clean install on python 2.7 too
```
apoorva@eternity:~/Desktop/retriever$ python --version
Python 2.7.12+
apoorva@eternity:~/Desktop/retriever$ python dashboard_script_parallel.py 
airports
biodiversity-response
bird-size


Saving a copy of 5599229...
Saving a copy of runways.csv...
Downloading runways.csv: 0.00B [00:00, ?B/s]
Saving a copy of 5603108...
Downloading 5603108: 192kB [00:05, 39.2kB/s]                                    
Installing biodiversity_response_population_response:   0%| | 0/761 [00:00<?, ?rInstalling biodiversity_response_population_response:  54%|▌| 411/761 [00:00<00:Installing biodiversity_response_population_response: 100%|█| 761/761 [00:00<00:00, 3339.89rows/s]
('Successfully installed==>', u'biodiversity-response')
gdp
Downloading 5599229:  25%|████▊              | 208k/828k [00:05<00:11, 56.8kB/s]
Saving a copy of gdp.csv...
Downloading gdp.csv: 448kB [00:00, 528kB/s]                                     
Installing gdp_gdp: 100%|████████████| 11507/11507 [00:00<00:00, 21140.25rows/s]
Downloading 5599229:  58%|███████████▌        | 480k/828k [00:07<00:02, 156kB/s]('Successfully installed==>', u'gdp')
portal

Saving a copy of 3299474...
Downloading 5599229: 832kB [00:08, 105kB/s]                                     
Downloading 3299474: 8.00kB [00:02, 3.90kB/s]3/3769 [00:01<00:00, 2711.56rows/s]
Installing bird_size_species: 100%|█████| 3769/3769 [00:01<00:00, 2750.64rows/s]
('Successfully installed==>', u'bird-size')
Downloading runways.csv: 2.84MB [00:12, 243kB/s]                                
Installing airports_runways: 100%|████| 40651/40651 [00:07<00:00, 5709.27rows/s]
Installing portal_main: 100%|█████████| 35549/35549 [00:11<00:00, 3119.26rows/s]
Installing airports_airports:  11%|▍   | 5619/53338 [00:00<00:07, 6332.00rows/s]('Successfully installed==>', u'portal')
Installing airports_airports: 100%|███| 53338/53338 [00:08<00:00, 6438.47rows/s]
Installing airports_countries: 100%|██████| 247/247 [00:00<00:00, 4446.28rows/s]
Installing airports_regions: 100%|█████| 3999/3999 [00:00<00:00, 11545.82rows/s]
Installing airports_airport_frequencies:   6%| | 1722/28644 [00:00<00:01, 17211.Installing airports_airport_frequencies:  12%| | 3396/28644 [00:00<00:01, 17064.Installing airports_airport_frequencies:  18%|▏| 5080/28644 [00:00<00:01, 16995.Installing airports_airport_frequencies:  24%|▏| 6837/28644 [00:00<00:01, 17161.Installing airports_airport_frequencies:  30%|▎| 8514/28644 [00:00<00:01, 17041.Installing airports_airport_frequencies:  36%|▎| 10187/28644 [00:00<00:01, 16946Installing airports_airport_frequencies:  42%|▍| 11954/28644 [00:00<00:00, 17154Installing airports_airport_frequencies:  48%|▍| 13634/28644 [00:00<00:00, 17044Installing airports_airport_frequencies:  53%|▌| 15313/28644 [00:00<00:00, 16967Installing airports_airport_frequencies:  59%|▌| 17000/28644 [00:01<00:00, 16909Installing airports_airport_frequencies:  65%|▋| 18739/28644 [00:01<00:00, 17049Installing airports_airport_frequencies:  71%|▋| 20411/28644 [00:01<00:00, 16921Installing airports_airport_frequencies:  77%|▊| 22101/28644 [00:01<00:00, 16913Installing airports_airport_frequencies:  83%|▊| 23859/28644 [00:01<00:00, 17106Installing airports_airport_frequencies:  89%|▉| 25559/28644 [00:01<00:00, 17025Installing airports_airport_frequencies:  95%|▉| 27261/28644 [00:01<00:00, 17022Installing airports_airport_frequencies: 100%|█| 28644/28644 [00:01<00:00, 16533.11rows/s]
Installing airports_navaids: 100%|████| 11117/11117 [00:01<00:00, 6086.60rows/s]
('Successfully installed==>', u'airports')
Time taken=34.4780440331 seconds
```

apoorvaeternity:
@henrykironde  I think the issue is with the progress bar because I generated the csvs and they were fine. @ethanwhite  This should not result into problem with the updated code that I have rebased and pushed.

ethanwhite:
@apoorvaeternity - ah, yep, I missed that update. Everything is working great for me now. :clap:

Issue #1019, opened 2017-11-04
Add `fetch` to Python interface


ethanwhite:
In the R package we have a `fetch` function which downloads a dataset as csv and then loads it into a list of data frames (one per table). We should add the same function to the Python interface. The default format for fetching should be a pandas data frame, but it might also be worth having some other optional formats (e.g., a numpy structured array).

henrykironde:
Just wondering which one would serve better in terms of memory.

ethanwhite:
Which input format (pandas vs. numpy) or something else? I think most folks will want the Pandas dataframe as the default, but once the basics are implemented I could definitely see adding some options for larger datasets that store the data in an SQL database or HDF5 so that they can be worked with on disk. We could also start to implement fine tuning based on types (e.g., https://www.dataquest.io/blog/pandas-big-data/).

zhangcandrew:
Pandas seems to have the ability to[ download a csv file directly from a url into a dataframe ](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html), and numpy supports [similar functionality when used with urllib](https://machinelearningmastery.com/load-machine-learning-data-python/). Which I think makes this a lot easier for us.

 When fetching the data, do we want our fetch to download a local copy of the csv file?

ethanwhite:
Because we want to take advantage of all of the cleaning code I think we probably want to use either `install_csv` to create a set of csv's and then load them or `install_sqlite` to create an SQLite file and then load all of the tables from it. Using SQLite would be a little more complicated up front because we'd have to write the code to look up the table names and then load them, but it might also be simpler in the long run since it would be an easier change to keep the data in the database instead of pulling it into memory.

Here's what we do in R (which uses csv not SQLite):
https://github.com/ropensci/rdataretriever/blob/master/R/rdataretriever.R#L72

adhaamehab:
Hi @ethanwhite @henrykironde @zhangcandrew 
I'll work on this issue.
Any example test or advice to start working on it ?

henrykironde:
This should be worked on after the PR #1042 . 
It will be much simpler. 

The mail goal is to use the python interface to load the dataset into a data frame like Pandas
https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html
http://www.datacarpentry.org/python-ecology-lesson/08-working-with-sql/

The retriever has several engines. We want to use the SQlite(prefered) or CSV engine to download a dataset and then use panda to to load it.

[...]

PR #1069, opened 2018-02-05
Add fetch to python Interface 

adhaamehab:
Hi @henrykironde Sorry, I was off schedule last days so I couldn't work on this issue as I told you.
This should solve #1019  but is this the right place for the method?


henrykironde:
Great start @adhaamehab, we just need to improve the design.

#### Parameters and Arguments

Let us assume we are given either `1` ,`2`or `3`  arguments (`dataset` name, `path`_to_the_`SQLite` _database, `table_name` ). Both `path_to_the_SQLite _database`, `table_name` are optional and can be initialized to `None`. These are the main connection parameters for the SQlite engine `retriever\retriever\engines\sqlite.py`. 

If only the `dataset` name is given, we use the default args which are:
**path_to_the_SQLite _database:** Defaults to working directory.
 Imported as` from retriever.lib.defaults import DATA_DIR` and created as os.path.join(DATA_DIR, "sqlite.db". In the PR you have used `os.path.join(HOME_DIR,'{}`, which would not be perfect for users. 

If table name is given then update the table name in the script object that belongs to that dataset, else use the default "{db}_{table}"

#### Edge cases

Avoid installation all the time by checking if the dataset has already been installed. 

#### Fetching only the dataset tables needed:

Right now, we are assume that the SQlite database will have only the tables that we want, however there will be many tables that are not associated to currently fetched dataset. You will have to access the script object to get the right table_names that need to be fetched.

adhaamehab:
Hi @henrykironde  Thanks for the explanation.
I already fixed all the points except the one about **_table name_**.

My problem with this is I didn't understand 
>If table name is given then update the table name in the script object that belongs to that dataset, else use the default "{db}_{table}"

As in many times, the dataset has multiple tables not only one

so, for example, the dataset **_airports_**  has 6 tables
so I tried to check what command like this will result 

` retriever install sqlite airports --table_name test_name`

when I viewed the resulted tables. There was only one table name test_name and it has the same content of  _only one_ table from the 6 tables which were supposed to be there.

So please if you have time tell me what I'm missing. Or is it an issue.

#### Result of 

`retriever install sqlite airports`

image


#### Result of

 `retriever install sqlite airports --table_name test33 `

image

adhaamehab:
@henrykironde  if you have time please check this PR

ethanwhite:
Thanks for your work on this @adhaamehab. I just went to test this out and ran into the fact that the function isn't being made publicly available. You'll need to add it to `lib/__init__.py` like the other functions so that it can then be called in the usual way.

retriever/lib/fetch.py
```diff
@@ -0,0 +1,30 @@
...
+from retriever.lib.defaults import HOME_DIR
+from retriever.lib.install import install_sqlite
+from retriever.lib.download import download
+
+
+
+def fetch(dataset):
+    """Install dataset into pandas dataframes.
+       each file in a separate dataframe
+    """
```
ethanwhite: Since this is a user facing function we want the docstring to be as descriptive as possible. I'd suggest something like:

```
"""Import a dataset into pandas data frames

Imports each table in the dataset into a data frame and stores the data frames
together in a list.

Parameters
----------
dataset : str
    Name of dataset

Returns
-------
list
    List of data frames, one for each table in the dataset

Example
-------
surveys, species, plots = rt.fetch('portal')
surveys.head()

"""
```

This is beyond what we have for some of our other functions, but it's good to start using the best docstrings possible as we add new functionality.

retriever/lib/fetch.py
```diff
@@ -0,0 +1,30 @@
...
+    cursor.execute("SELECT name FROM sqlite_master WHERE type='table';")
+    tables = [t[0] for t in cursor.fetchall()]
+    dataframes = [
+        pd.read_sql_query("SELECT * FROM {};".format(table), con) for table in tables
+    ]
+    # delete the sqlite file
+    os.remove(installation_path)
+    return dataframes
+    
+
```
ethanwhite: Remove the extra blank line at the end.

retriever/lib/fetch.py
```diff
@@ -0,0 +1,30 @@
...
+    """Install dataset into pandas dataframes.
+       each file in a separate dataframe
+    """
+    #download(dataset)
+    installation_path = os.path.join(HOME_DIR,'{}.sqlite'.format(dataset))
+    install_sqlite(dataset, file=installation_path)
+    con = sqlite3.connect(installation_path)
+    cursor = con.cursor()
+    cursor.execute("SELECT name FROM sqlite_master WHERE type='table';")
+    tables = [t[0] for t in cursor.fetchall()]
```
ethanwhite: It's generally better to avoid single character variables names even in cases like this. How about `table` instead of `t`?

retriever/lib/fetch.py
```diff
@@ -0,0 +1,30 @@
+
```
ethanwhite: Remove blank line.

retriever/lib/fetch.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+import sqlite3
+import pandas as pd
+from retriever.engines import sqlite
```
ethanwhite: Blank line before retriever imports.

retriever/lib/fetch.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+def fetch(dataset):
+    """Install dataset into pandas dataframes.
+       each file in a separate dataframe
+    """
+    #download(dataset)
+    installation_path = os.path.join(HOME_DIR,'{}.sqlite'.format(dataset))
+    install_sqlite(dataset, file=installation_path)
```
ethanwhite: Since we're just creating this and then deleting it what about an in memory database instead. It's quicker and doesn't require the clean up?

adhaamehab: Hi @ethanwhite  thanks for your review. I fixed all the issues pointed by your comments.
The only thing remaining is creating in memory db. Is there an already implemented method as the current method `install_sqlite` doesn't support in memory db

ethanwhite:
@adhaamehab - looks like the tests are failing because `pandas` isn't listed as a requirement. Can you add it to `requirements.txt` so that we can see if the tests are passing?

@henrykironde - do you think we should add this as a requirement to `setup.py` as well. I'm leaning in that direction. `pandas` is pretty heavy, but it's also something that almost anyone working with data in Python already has installed.

henrykironde:
I am going to have a thorough review on this PR today evening. 
Regarding the pandas as part of the requirements, I think with the fetch addition we should definitely add it. Like Ethan said, panda is heavy but most people doing data analysis in python have panda installed. So lets go for it.

ethanwhite:
Thanks @henrykironde. So, @adhaamehab, that means we should also add `pandas` to `install_requires` in `setup.py`.

adhaamehab:
Hi @henrykironde  @ethanwhite  any update about this PR?

ethanwhite:
@adhaamehab - this looks good to me in general. You're totally right about the in-memory issue. You can install into one, but it would be closed before we could access it and now isn't the time to engineer around that.

The one issue I found with local testing is that when installing datasets with only a single table we end up with 2 "tables" in the output. The second table is basically the name of the first table and it's length:

```
In [32]: mom[1]
Out[32]: 
                         name   seq
0  mammal_masses_MammalMasses  5731
```

This issue doesn't happen with datasets with more than one table. To fix this we'll need to either follow @henrykironde's recommendation of only getting specifically named tables, or confirm that this always (and only) happens with single table datasets and drop the second value from the list as part of `fetch`.

It would also be good to have at least one test added for this new function.

ethanwhite:
I've also noticed an issue where if I happen to have a file named `sqlite.db` in my home directory it causes installation errors.

henrykironde:
@adhaamehab, This is a great start, sorry for taking long to review this PR. Let us change the design. I think there is a cyclic dependency introduced. 

If I have a `script test_fetch.py` with the code
```python
import retriever as rt
data = rt.fetch('portal')
```
Running `python test_fetch.py` fails.

I think we can include the method fetch in install.py instead of creating a separate file or if we keep it in fetch.py, we should figure out where the imports fails.

Refactor the method.

Let us move part of the code to the sqlite engine
In the main engine, create a function that will be overwritten by the sqlite engine. 
So we can have some method in engine.py as `method_name_fetch_tables(self, args):` and  sqlite overrides this to include the logic in the code below. 

```python
    con = sqlite3.connect(installation_path)
    cursor = con.cursor()
    cursor.execute("SELECT name FROM sqlite_master WHERE type='table';")
    tables = [table[0] for table in cursor.fetchall()]
    dataframes = [
        pd.read_sql_query("SELECT * FROM {};".format(table), con) for table in tables
    ]
    # delete the sqlite file
    os.remove(installation_path)
```

So we shall have to provide the arguments like in the install_sqlite function 
```python
def install_sqlite(dataset, file= os.path.join(DATA_DIR, 'sqlite.db'),
                   table_name='{db}_{table}',
                   debug=False, use_cache=True):
    """Install datasets into sqlite."""
    args = {
        'command': 'install',
        'dataset': dataset,
        'engine': 'sqlite',
        'file': file,
        'table_name': table_name
    }
    _install(args, use_cache, debug)
```

In the SQlite engine, this function `method_name_fetch_tables(self, args):` should be able to query the tables for any dataset. 

As Ethan mentioned, we should create a dataframe where we map a table name and the data instead of a list of data.  `I just had another thought about this PR. What about, instead of storing the tables in a list, we store them in a dictionary with the keys being the table names? `

adhaamehab:
@henrykironde Thanks for your time.

I think this PR is ready now.
I also squashed the commits into one.

adhaamehab:
@henrykironde Can you check why `test_sqlite_integration` fail?

henrykironde:
Sent you a fix for the sqlite engine fails

adhaamehab:
@henrykironde  Thanks, solved :+1:

adhaamehab:
@henrykironde Any updates about this PR?

henrykironde:
I am still testing it. will get back to you today

retriever/engines/sqlite.py
```diff
@@ -106,3 +109,33 @@ def get_connection(self):
...
         return dbapi.connect(self.opts["file"])
+
+
+    @staticmethod
+    def fetch_tables(dataset, file= os.path.join(DATA_DIR, 'sqlite.db'),
+                table_name=None):
+        """Return sqlite dataset as list of pandas dataframe."""
+        import sqlite3
+        
+        con = sqlite3.connect(file)
```
henrykironde: This looks like hard coding.
We should use the script instance. We know the engine, so we can instantiate the right engine(sqlite). The engine has the connection methods that can provide us with the connection cursor. 

For now we should get the table names from the  table`s property in the script object. This will only work for `Json` datasets and will fail for `.py` datasets.

Later, we shall re design the `.py` datasets to be able to get the tables initialized like the `Json` datasets

adhaamehab:
Hi @henrykironde  If you can have a look here.

Some tests and their output

```python
import retriever as rt
air = rt.fetch('airports', table_name='air_{table}')
print(air.keys())
print(air['air_airports'].head())
```
```shell
Installing air_airports: 100%|███████| 52865/52865 [00:04<00:00, 11739.00rows/s]
Installing air_airport_frequencies:   9%| | 2699/28640 [00:00<00:00, 26987.97rowInstalling air_airport_frequencies:  17%|▏| 4779/28640 [00:00<00:01, 22509.91rowInstalling air_airport_frequencies:  26%|▎| 7474/28640 [00:00<00:00, 23679.81rowInstalling air_airport_frequencies:  36%|▎| 10170/28640 [00:00<00:00, 24576.23roInstalling air_airport_frequencies:  45%|▍| 12964/28640 [00:00<00:00, 25495.08roInstalling air_airport_frequencies:  54%|▌| 15522/28640 [00:00<00:00, 25519.17roInstalling air_airport_frequencies:  64%|▋| 18254/28640 [00:00<00:00, 26031.92roInstalling air_airport_frequencies:  73%|▋| 21000/28640 [00:00<00:00, 26428.05roInstalling air_airport_frequencies:  83%|▊| 23828/28640 [00:00<00:00, 26955.50roInstalling air_airport_frequencies:  93%|▉| 26592/28640 [00:01<00:00, 27156.96roInstalling air_airport_frequencies: 100%|█| 28640/28640 [00:01<00:00, 25949.35rows/s]
Installing air_runways: 100%|█████████| 40582/40582 [00:04<00:00, 9886.36rows/s]
Installing air_countries: 100%|██████████| 247/247 [00:00<00:00, 16848.16rows/s]
Installing air_regions: 100%|██████████| 3999/3999 [00:00<00:00, 16413.11rows/s]
dict_keys(['air_airports', 'air_airport_frequencies', 'air_runways', 'air_navaids', 'air_countries', 'air_regions'])
       id ident           type                                name  \
0    6523   00A       heliport                   Total Rf Heliport   
1  323361  00AA  small_airport                Aero B Ranch Airport   
2    6524  00AK  small_airport                        Lowell Field   
3    6525  00AL  small_airport                        Epps Airpark   
4    6526  00AR         closed  Newport Hospital & Clinic Heliport   

   latitude_deg  longitude_deg  elevation_ft continent iso_country iso_region  \
0     40.070801     -74.933601          11.0        NA          US      US-PA   
1     38.704022    -101.473911        3435.0        NA          US      US-KS   
2     59.949200    -151.695999         450.0        NA          US      US-AK   
3     34.864799     -86.770302         820.0        NA          US      US-AL   
4     35.608700     -91.254898         237.0        NA          US      US-AR   

   municipality scheduled_service gps_code iata_code local_code home_link  \
0      Bensalem                no      00A                  00A             
1         Leoti                no     00AA                 00AA             
2  Anchor Point                no     00AK                 00AK             
3       Harvest                no     00AL                 00AL             
4       Newport                no                                           

  wikipedia_link keywords  
0                          
1                          
2                          
3                          
4                    00AR  

```

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "python.linting.pylintEnabled": false,
```
henrykironde: do we need this file?

retriever/engines/sqlite.py
```diff
@@ -106,3 +108,27 @@ def get_connection(self):
         import sqlite3 as dbapi
         self.get_input()
         return dbapi.connect(self.opts["file"])
+
+    def fetch_tables(self, dataset, table_name='{db}_{table}', file=os.path.join(DATA_DIR, 'sqlite.db')):
```
henrykironde: @adhaamehab, I think we do not need all these parameters. self has everything you would need, including the connection cursor. And we should not select from `sqlite_master`. We should only use the table name function `self.table_name()` to get the name of the table we want.

henrykironde:
@adhaamehab, great work. I will do some testing and let you know.

adhaamehab:
@henrykironde Any updates about this PR?

henrykironde:
@adhaamehab, we put this to a hold because it would fail with certain python scripts until we get those working well. I am working on this with Andrew and once done, we shall let u know.

ethanwhite:
@adhaamehab - We recently merged this in f7dc96c22c80982cec81c86268e87c5ad76a6669 and it is now available in the new release. Thanks so much for you contribution and apologies for it taking so long to get merged. I'm personally super excited to have this functionality now and you helped make that happen! :clap: :clap: :clap:

PR #1070, opened 2018-02-05
Add support to generate csv for all tables of a dataset using to_csv

apoorvaeternity:


henrykironde:
Do not forget to check overridden methods in the specific engines

apoorvaeternity:
@henrykironde Please review. I have fixed the name of the csv files to be datasetname_tablename.csv .

henrykironde:
```
# -*- coding: latin-1  -*-
# """Integrations tests for Data Retriever"""

import json
import os
import shutil
import sys
from imp import reload

from retriever.lib.defaults import ENCODING

encoding = ENCODING.lower()

reload(sys)
if hasattr(sys, 'setdefaultencoding'):
    sys.setdefaultencoding(encoding)
import pytest
from retriever.lib.load_json import read_json
from retriever.lib.defaults import HOME_DIR
from retriever.engines import engine_list
from retriever.lib.engine_tools import file_2list
from retriever.lib.engine_tools import create_file

# Set postgres password, Appveyor service needs the password given
# The Travis service obtains the password from the config file.
if os.name == "nt":
    os_password = "Password12!"
else:
    os_password = ""


mysql_engine, postgres_engine, sqlite_engine, msaccess_engine, \
csv_engine, download_engine, json_engine, xml_engine = engine_list


def get_script_module(script_name):
    """Load a script module."""
    return read_json(os.path.join(HOME_DIR, "scripts", script_name))


dbfile = os.path.normpath(os.path.join(os.getcwd(), 'testdb.sqlite'))
sqlite_engine.opts = {'engine': 'sqlite', 'file': dbfile, 'table_name': '{db}_{table}'}

script_module = get_script_module("airports")
script_module.download(sqlite_engine)
script_module.engine.final_cleanup()
script_module.engine.to_csv()
```

henrykironde:
This sets the name based on the table name and db name 
```
    def table_name(self, name=None, dbname=None):
        """Return full tablename."""
        if not name:
            name = self.table.name
        if not dbname:
            dbname = self.database_name()
            if not dbname:
                dbname = ''
        return self.opts["table_name"].format(db=dbname, table=name)
```

henrykironde:
test_tocsv

henrykironde:
So I debugged the code and I have pasted the code that is needed for the a simple test run. The image shows that we can actually get the values from script.tables.. object.

I feel it is better than `list(self.script.urls.keys()):`
and if we use `def table_name(self, name=None, dbname=None):` with arguments, It will create the right names

henrykironde:
Let me know what you find out

henrykironde:
based on what i have found out, 
This will do the work in the engine.py for mysql, postgres and sqlite
```
        from retriever.lib.engine_tools import sort_csv
        for t_name in list(self.script.tables.keys()):
            table_name = self.table_name(name=t_name)
```

It is kind of the same logic as the other engines(xml, json, csv, engines)
```
    def to_csv(self):
        """Export table from json engine to CSV file"""
        for keys in list(self.script.tables):
            table_name = self.opts['table_name'].format(db=self.db_name, table=keys) 
```

retriever/lib/engine.py
```diff
@@ -680,17 +680,22 @@ def table_name(self, name=None, dbname=None):
...
 
-    def to_csv(self):
+    def to_csv(self, db=None):
         # Due to Cyclic imports we can not move this import to the top
         from retriever.lib.engine_tools import sort_csv
-        for _ in list(self.script.urls.keys()):
-            table_name = self.table_name()
+        for table in list(self.script.urls.keys()):
+            table_name = "{}_{}".format(self.database_name(), table)
+            if db == "postgres" or db == "mysql":
```
henrykironde: One thing that you should note, `class Engine` is the main class for all other engines and it should be generic as possible.
Each engine `(postgres, mysql, csv and etc)` inherits these methods/functions and if they differ from the general method, that engine overrides(customizes) that method.

Since engines `postgres and mysql`  want to implement something different from the main function `def to_csv()` in the main class, these engines classes `(postgres and mysql)` should instead override the entire method.

In this way we avoid checks for specific variables that do not apply to the general object. ` if db == "postgres" or db == "mysql":`

apoorvaeternity: Okay 👍

retriever/engines/postgres.py
```diff
@@ -141,3 +141,7 @@ def get_connection(self):
         db_encoding = encoding_lookup.get(encoding)
         conn.set_client_encoding(db_encoding)
         return conn
+
+    def to_csv(self):
```
henrykironde: we don't need this. When the code is executed, a call to the function that is not overriden will call the function in the main class

apoorvaeternity:
@henrykironde Please review.

retriever/engines/postgres.py
henrykironde: same applies to mysql,  Just delete  that part

henrykironde:
mysql

henrykironde:
perfect

henrykironde:
This is a very important addition. :+1: . So we  let us clean up. Squash all commit to 1 or 2 commits and we get this in. 
@shreyneil added some test and after this, his PR should be able to work fine

apoorvaeternity:
@henrykironde Done :)

retriever/lib/engine.py
```diff
@@ -683,8 +683,8 @@ def table_name(self, name=None, dbname=None):
     def to_csv(self):
         # Due to Cyclic imports we can not move this import to the top
         from retriever.lib.engine_tools import sort_csv
-        for _ in list(self.script.urls.keys()):
-            table_name = self.table_name()
+        for table in list(self.script.tables.keys()):
```
henrykironde: let us change the table name because we have table as a global. 
table_i or table_n or whatever you are fine with

apoorvaeternity:
@henrykironde Ready to merge. The AppVeyor build is queued.

Issue #1063, opened 2018-02-01
Excessive script downloads caused by name-matching function


zhangcandrew:
Right now there is a function called name_matches in engine_tools which helps users download scripts that are closely related to what they type in. I.e. if a user accidentally types `retriever install csv iriis`, the name_matches function should still download iris since it recognizes the user just made a simple error. 

The problem we have right now is that if we type `retriever install csv mapped-plant-quads-co` and only want mapped-plant-quads-co, retriever installs mapped-plant-quads-co, mapped-plant-quads-ks, mapped-plant-quads-id, and mapped-plant-quads-mt due to the name matching function.

ethanwhite:
My personal take is that we should match other package managers and:

1. Only install if the names match exactly
2. If the name doesn't match an existing dataset then check to see if there is something close using `name_matches`, tell the user that there is no dataset by that name, and return a "did you mean XXXX" message.

ss-is-master-chief:
@ethanwhite @zhangcandrew Rather than having different packaged functions like `install_csv`, `install_json`, etc. would it be scalable to have just `install` as a command for the CLI, followed by the type? Something like this : 
```
install --+
          |--- csv
          |--- json
          |--- xml
          |--- msaccess
          |--- postgres
          |--- sqlite
          |--- mysql
```

ethanwhite:
@ss-is-master-chief - I don't understand the relationship between the install functions and the name matching in this issue, so this may be best discussed in a separate issue.

ethanwhite:
Fixed in 8ad74192ce9ca1508452ff3dd6c1dfcf43cafab2

PR #1072, opened 2018-02-06
modified the name_matches function 

zhangcandrew:
name_matches function now only checks for distances among script names and NOT keywords. This was made to address #1063. 

If the name matches exactly, the scripts are downloaded. If not, suggestions are given, but no scripts are downloaded. 

The constant of .53 for the levenshtein distance was arbitrarily chosen based on tests done by misspelling vertnet.

### DIFF
--- a/retriever/lib/install.py
+++ b/retriever/lib/install.py
@@ -31,9 +31,8 @@ def _install(args, use_cache, debug):
                 if debug:
                     raise
     else:
-        message = "The dataset \"{}\" isn't available in the Retriever. " \
-                  "Run retriever.datasets()to list the currently available " \
-                  "datasets".format(args['dataset'])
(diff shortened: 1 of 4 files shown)


retriever/lib/install.py
```diff
@@ -31,9 +31,8 @@ def _install(args, use_cache, debug):
                 if debug:
                     raise
     else:
-        message = "The dataset \"{}\" isn't available in the Retriever. " \
-                  "Run retriever.datasets()to list the currently available " \
-                  "datasets".format(args['dataset'])
+        message = "Run retriever.datasets()to list the currently available " \
+                  "datasets.".format(args['dataset'])
```
henrykironde: I think we do not need the .format()

PR #1087, opened 2018-02-15
[wip]:Clean up commit history

henrykironde:
Due to un squashed commits that we introduced by mistake during merges.

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -147,23 +147,18 @@ page.
 Creating or Updating a Conda Release
 ====================================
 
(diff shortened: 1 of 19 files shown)


henrykironde:
@ethanwhite, we introduced some un squashed commits that distorted the commit organization/history.
I have rebased some of those commits.
Have a look and let me know if we should force a push to upstream.

ethanwhite:
If I understand correctly you're asking about rewriting the history in `master` to make the history more readable. We should never change the history in `master` since it can cause serious issues for folks working on forks/branches, so we'll just need to live with a little bit of messy history and try to catch these before merging in the future. If I'm misunderstanding or missing the full complexity stop by today and we can chat about it more.

Issue #979, opened 2017-07-22
Is `urls` property in json files needed


ethanwhite:
This isn't part of the spec and it looks to be like `lib\compile.py` assembles the urls (appropriately) from the individual `resources`. I removed this `mammal-life-hist` and it seems to run fine.

ShivamNegi:
I don't understand what you are trying to say here. Do you mean that the `json` files don't require `urls` field?

ethanwhite:
Yes.

zhangcandrew:
@henrykironde and I were just talking about this the other day. I think @goelakash introduced them in a previous version, so maybe he could elaborate on his reasoning for them?

zhangcandrew:
Finally got around to digging into this a little more, so it looks like the urls property has only two main uses right now: 

1. Checking to see if any scripts are already downloaded (found in engine.py): 
![image](https://user-images.githubusercontent.com/15080652/36169839-6ae1fb38-10cb-11e8-93e3-20d2028a347d.png)

2. Grabbing information from the scripts quickly for citations/descriptions (found in conf.py and templates.py):
conf.py:
![image](https://user-images.githubusercontent.com/15080652/36170034-1cf5e94c-10cc-11e8-920a-d6441147edee.png)
templates.py:
![image](https://user-images.githubusercontent.com/15080652/36170054-383a812c-10cc-11e8-9887-ee3b4eab42d1.png)

@ethanwhite is right though, removing the urls attribute from the script does not affect downloads in any way; based on our implementation in datapackage.py, we download datasets purely off of the urls from the tables within the resources attribute. 

I think having the urls attribute is the neatest/quickest way to implement those functions, although, neither of the functions are vital to functioning - just good to have. Any thoughts?

ethanwhite:
The one downside to keeping this is that we are storing the same information in two different places. These two places can get out of sync if someone only remembers to update one of them and this can be confusing (see #1078).

Seems like we could store the urls only with the tables and then recreate the urls attribute with something like:

``` 
script.urls = {table: script.tables[table].url for table in script.tables.keys()}
```

This would let us keep the functionality without needing to store the data twice.

zhangcandrew:
I think that's a good solution to address the consistency issue. 

I'll have a PR up for this soon!

zhangcandrew:
Closing this issue!

PR #1089, opened 2018-02-15
Deleted the url attribute

zhangcandrew:
Deleted the url attribute.

in relation to #979


henrykironde:
Thanks @zhangcandrew.

Issue #1091, opened 2018-02-18
Verbose Mode Overwritten by License/Keyword Search


zhangcandrew:
In a prior version of retriever, the user was allowed to run `retriever ls -l` to return a verbose list of all of our datasets. This functionality no longer exists since it was overwritten in a newer PR.

I think it might still be useful to have in the retriever though. 

See #488,  #1037.

adhaamehab:
Hi @zhangcandrew I will work in this issue

adhaamehab:
@zhangcandrew  could you review the PR?

ethanwhite:
Fixed in a6ddc9cbd697c0cae91db79ca2983c8e0b46eabe. Thanks @adhaamehab!

PR #1092, opened 2018-02-18
Add verbose mode to ls

adhaamehab:
Verbose Mode Overwritten by License/Keyword Search
so this commit re-add the verbose mode again
command added : `ls -v`

fix issue #1091

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -20,7 +20,7 @@ from retriever.lib.datasets import datasets, dataset_names, license
 from retriever.lib.defaults import sample_script, CITATION, ENCODING, SCRIPT_SEARCH_PATHS
 from retriever.lib.get_opts import parser
 from retriever.lib.repository import check_for_updates
-from retriever.lib.scripts import SCRIPT_LIST
+from retriever.lib.scripts import SCRIPT_LIST, get_script
 from retriever.lib.engine_tools import name_matches, reset_retriever
 
 encoding = ENCODING.lower()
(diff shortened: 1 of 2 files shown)


zhangcandrew:
Thanks @adhaamehab!

It's looking great! I only have a few suggestions: 

1. To keep consistent between the -k, -l, and -v flags, I think we should format the output so that the numbering is #. instead of #-. Adding on to that, it would also be good to have -v output formatted in a similar fashion to that of the -k and -l flags:
```
script.title
script.name
script.keywords
script.description
script.license
script.citation
```

2. As a user I think it might be useful to return the verbose of a specific dataset if I know which dataset I want to know more on, something like `retriever ls -v dataset1 dataset2 dataset3` If I only want to see those 3 datasets expanded. 

3. Since these are new flags, we might need to add them to the documentation (although, looking at the documentation right now, I think we're missing a lot about the command-line and other extended functionality, so this might be better for a separate issue and new PR where we looks over the docs again). 

Let me know what you think!

adhaamehab:
@zhangcandrew  awesome notes I'll do and push the new changes.
for the documentation. I think it's better opening a new issue for updating the docs

retriever/__main__.py
```diff
@@ -131,12 +131,33 @@ def main():
                 print("No scripts are currently available. Updating scripts now...")
                 check_for_updates(False)
                 print("\n\nScripts downloaded.\n")
-
-            if args.l is None and args.k is None:
-                all_scripts = dataset_names()
+            if not (args.l or args.k or (type(args.v) is list)):
+                all_scripts = datasetget_script_names()
```
zhangcandrew: I think you might have accidentally inserted "get_script" into this. I tried it with `all_scripts = dataset_names()` and it seems to work fine. 

Otherwise, this looks real good!

zhangcandrew:
This looks good! 

@henrykironde, wondering if you have any thoughts on this, otherwise I'll be merging it in!

zhangcandrew:
Looks great @adhaamehab! 

I'm merging it in now. Feel free to re-open or create a new issue if anything arises!

PR #1094, opened 2018-02-20
[WIP] Raster Support

ss-is-master-chief:
Code contains modules for:
1. ESRI Grid
2. GeoTIFF
3. HDF (under development)

@henrykironde @ethanwhite Please have a look

### DIFF
--- a/get_raster_files.py
+++ b/get_raster_files.py
@@ -1,11 +1,9 @@
-import sys
 import glob
-import numpy as np
-import sqlite3
-import csv
 import linecache
 import pandas as pd
-import os.path
+import os
+from tqdm import tqdm
+import sys
 
 # Resolve link : https://gis.stackexchange.com/questions/233654/install-gdal-python-binding-on-mac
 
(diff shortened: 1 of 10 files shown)


ss-is-master-chief:
Update -
1. Progress bar for ESRI Grid
2. Check for existing data files and updating them for ESRI

henrykironde:
Looks nice, there are some errors. I think you forgot to update this function call  ` load_data(file_name)`.
When reading files, we should provide a path and not assume that the working directory is the path where the files are always. The reading of all files can be handled in the same function.

henrykironde:
So the design of reading files should be as follows:

User provides either path to a single file that needs to be processed. If path is a file then we need to check if the file is a raster. If path is a directory, get a list of all files in that path directory and for each file we can check if it's a raster of not.

a sample design you could borrow some ideas from is below

```python
class Raster:
    """ processes the spatial files from a given location
    """

    raster_extensions = ['.gif', '.img', '.bil', '.jpg', '.tif', '.tiff','.hdf','.l1b']
    multi_file = ['.hdf','hd5']

    def __init__(self):
        self.data_sets = list()
        self.firm = []

    def read_source(self, pathx):
        """read source folder or file and return list of Raster files
        """
        if os.path.isfile(os.path.normpath(pathx)):
            self.firm = [os.path.normpath(pathx)]
        else:
            self.firm = [os.path.normpath(os.path.join(root, name))
                         for root, dirs, files in os.walk(pathx, topdown=False) for name in files]

    def get_bands(self):
        """get all the bands from the source files"""
        for filenames in self.firm:
            if ....
```

henrykironde:
What errors are you getting? 
What commands do you want me to run / examples of commands

ss-is-master-chief:
image

Here I have made two copies of the same `.asc` file just for testing purposes. The `.asc` files are located in a `virtual environment` and I install the `.csv` to `Desktop`. I will update the code the add support for other initial file types as well.

@henrykironde Please have a look and let me know of any error you might face while running locally.

Updates:
1. Adding working directory
2. Adding target directory
3. To install `all` files or a `specific` files

ss-is-master-chief:
@henrykironde I was refactoring the modules for GeoTIFF. When running through the CLI, I am able to install the data set. But while using the Python's interface, I am facing an error since I am not able to initialize spatial meta-data. I will keep on working on this. Please let me know your thoughts on this. 

The `load_tif_data` was wrongly called as `load_tif`.

henrykironde:
Clean up this code. 
First we do not need any of the `setting working directory` or 



Have a list of raster extensions
raster_extensions = [".asc",".tif", ".bil",.....]

Have a function that checks if the needed extensions are available

```python
check needed extensions(conn)
      # since we have connection conn
      # use conn to check for the needed extensions 
      if the extenion is not loaded can you load it and verify that its loaded(may be with a try block )
```

Have a function that installs any file given to it

Install raster function (): I think you have it as `load_tif_data` but it should not only load tif.
Let us make it generic

```
load_raster(path)
   os.system("rasterlite_load -d {}.sqlite -T {} 
````

@ss-is-master-chief why do we have two operations one with -t and  -v

 ```
   def load_tif_data(self,file_name):
        os.system("rasterlite_load -d {}.sqlite -T {} -D . -t".format(file_name,file_name))
        os.system("rasterlite_load -d {}.sqlite -T {} -D . -v".format(file_name,file_name))
```

We should have a test function
This should open the database and get the table names or table names and a  " SELECT * from [table name]  LIMIT 3"

henrykironde:
So when you test a raster with a new extension for example `filename.NEWEXTENSION` and you are able to install it, then add it to the raster_extensions. [".asc",".tif", ".bil","NEWEXTENSION",.....]

henrykironde:
In fact lets change the raster_extension to raster_format to avoid confusion.

ss-is-master-chief:
@henrykironde 

`why do we have two operations one with -t and -v` : I have used `rasterlite_load` with the `-t` flag for testing purposes only. I'll remove this from the final scripts.

`lets make it generic` : Yes I am currently working on it.

`In fact lets change the raster_extension to raster_format to avoid confusion` : Yeah I'll do that

`First we do not need any of the setting working directory` :  What about multi directory support?

`Have a function that checks if the needed extensions are available` : I'll add that to the script

ss-is-master-chief:
@henrykironde I do not think this is required.

```
  curs.execute("SELECT load_extension('mod_rasterlite2')")
  curs.fetchall()

  curs.execute("SELECT load_extension('mod_sqlite')")
  curs.fetchall()

  curs.execute("SELECT InitSpatialMetadataMeta(1)")
  curs.fetchall()
```
Upon creating an empty SQLite database using the ```spatialite``` tool, the database is initialised with Spatial Metadata. It wouldn't have been the case if I had used ```sqlite3``` directly to create the DB. Using sqlite3 requires to initialise a database with metadata later. I'll change the code accordingly. Please comment.

henrykironde:
Lets have a chat on hangouts to discus this further. Ping me whenever you are free.

ss-is-master-chief:
@henrykironde Please have a look.

The changes take care of the issues we were having with creating a DB and initializing it with MetaData. `JPG`,`JPEG`,`TIFF`,`PNG` file formats are taken care of by using `rasterlite_load`.

I have cleaned the code and removed directory support for now. Support can be added by changing the parameter `-D` in `rasterlite_load`. 

You can test the code by running `python get_raster_files.py`. I'll extend the script for `ESRI GRID` since I do not think `rasterlite_load` supports it.

Please let me know if things are working propoerly on your end.

henrykironde:
@thanks at @ss-is-master-chief. The code looks clean, I think its time we start testing the files supported.

ss-is-master-chief:
Thanks @henrykironde. I'll have the function ready over the next couple of days.

ss-is-master-chief:
@henrykironde 
Closing 
new ref: #1178

Issue #182, opened 2014-04-19
Add GloBI interaction data


ethanwhite:
https://github.com/jhpoelen/eol-globi-data/wiki

ethanwhite:
The single table tsv version of this data is available here:
https://depot.globalbioticinteractions.org/snapshot/target/data/tsv/interactions.tsv.gz

shreyneil:
Sir, i am unable to access this data as whenever i try to access it through a reader ,the software itself gets hanged due to the huge size, any other way on how to access  the data, that you might have in your mind?

ethanwhite:
To figure out the basic structure I would start by using the command line to grab a small piece of the data. E.g.,

```bash
head -200 interactions.tsv > small_interactions.tsv
```

would give you the first 200 lines of the file in `small_interactions.tsv`, which you could then easily open in the usual way. That will let you figure out the basic structure, which should let you write the `json` script. Since database management systems can work with data larger than RAM once it's loaded into a database it will be easier to work with.

PR #1096, opened 2018-02-22
adding globi interactions data

shreyneil:
This PR solves the following  issue https://github.com/weecology/retriever/issues/182 . Please review and get back.

### DIFF
--- a/scripts/globi-interaction.json
+++ b/scripts/globi-interaction.json
@@ -1,0 +1,233 @@
+{
+  "citation": "NA",
+  "description": "GloBI contains code to normalize and integrate existing species-interaction datasets and export the resulting integrated interaction dataset.",
+  "homepage": "https://github.com/jhpoelen/eol-globi-data/wiki",
+  "keywords": [
+    "source",
+    "target"
+  ],
+  "name": "globi_interaction",
+  "archived": "gzip",
+  "resources": [
+    {
+      "path": "interactions.tsv/data",
+      "dialect": {
+        "header_rows": 1
+      },
+      "name": "interactions",
+      "schema": {
+        "fields": [
+          {
+            "name": "sourceTaxonId",
+            "type": "char"
+          },
+          {
+            "name": "sourceTaxonIds",
+            "type": "char"
+          },
+          {
+            "name": "sourceTaxonName",
+            "type": "char"
+          },
+          {
+            "name": "sourceTaxonRank",
+            "type": "char"
+          },
+          {
+            "name": "sourceTaxonPathNames",
+            "type": "char"
+          },
+          {
+            "name": "sourceTaxonPathIds",
+            "type": "char"
+          },
+          {
+            "name": "sourceTaxonPathRankNames",
+            "type": "char"
+          },
+          {
+            "name": "sourceId",
+            "type": "char"
+          },
+          {
+            "name": "sourceOccurrenceId",
+            "type": "char"
+          },
+          {
+            "name": "sourceCatalogNumber",
+            "type": "char"
+          },
+          {
+            "name": "sourceBasisOfRecordId",
+            "type": "char"
+          },
+          {
+            "name": "sourceBasisOfRecordName",
+            "type": "char"
+          },
+          {
+            "name": "sourceLifeStageId",
+            "type": "char"
+          },
+          {
+            "name": "sourceLifeStageName",
+            "type": "char"
+          },
+          {
+            "name": "sourceBodyPartId",
+            "type": "char"
+          },
+          {
+            "name": "sourceBodyPartName",
+            "type": "char"
+          },
+          {
+            "name": "sourcePhysiologicalStateId",
+            "type": "char"
+          },
+          {
+            "name": "sourcePhysiologicalStateName",
+            "type": "char"
+          },
+          {
+            "name": "interactionTypeName",
+            "type": "char"
+          },
+          {
+            "name": "interactionTypeId",
+            "type": "char"
+          },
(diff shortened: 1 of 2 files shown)


henrykironde:
Could you please post and example run `retriever install sqlite globi_interaction`. I am wondering if you are able to download the file

zhangcandrew:
Hey @shreyneil, 

The script is failing for me. Returning:
```
HTTP Error: 403 Forbidden
```

shreyneil:
The link was provided in the issue itself by @ethanwhite , don't know why is it failing

scripts/globi-interaction.json
```diff
@@ -0,0 +1,233 @@
+{
+  "citation": "NA",
+  "description": "GloBI contains code to normalize and integrate existing species-interaction datasets and export the resulting integrated interaction dataset.",
+  "homepage": "https://github.com/jhpoelen/eol-globi-data/wiki",
+  "keywords": [
+    "source",
+    "target"
+  ],
+  "name": "globi_interaction",
+  "archived": "gzip",
```
zhangcandrew: This should be changed to "gz"

scripts/globi-interaction.json
```diff
@@ -0,0 +1,233 @@
...
+  "resources": [
+    {
+      "path": "interactions.tsv/data",
```
zhangcandrew: The path should be "interactions.tsv/interactions.tsv"

ethanwhite:
The url works fine in the browser, but 403's when called from inside Python. This is likely due to the site blocking some user agents, e.g.,:

https://stackoverflow.com/questions/34692009/download-image-from-url-using-python-urllib-but-receiving-http-error-403-forbid

I'll contact the GLOBI devs and see if they can help.

ethanwhite:
I chatted with the folks at GLOBI and it doesn't seem like this is user agent blocking. In fact, if I use `requests` this download works fine:

```
import requests

r = requests.get('https://depot.globalbioticinteractions.org/snapshot/target/data/tsv/interactions.tsv.gz', stream=True)

with open('interactions.tsv.gz', 'wb') as f:
    for chunk in r.iter_content():
            f.write(chunk)
```

So, this has something to do with `urllib` and something about that link that is abnormal (since `urllib` works fine with our 100s of other urls).

shreyneil:
So what do you suggest @ethanwhite ?

ethanwhite:
I'm not sure I have any great suggestions. Long-term it's probably worth moving over the `requests` for all of our downloading since it's become the gold standard. It will also make working with API based datasets easier. Short-term, if we really want to get this dataset added now I think we'd probably have to switch over to a Python script so that we could download the file manually using requests. I suspect that this would end up being too much work to justify the dataset.

zhangcandrew:
Digging a bit into this, it looks like the file is being downloaded from a redirected url: `https://s3.amazonaws.com/globi/snapshot/target/data/tsv/interactions.tsv.gz`. 

Replacing the url in the script with that seems to get the download started with a progress bar. The download itself takes a while and doesn't display anything for sqlite downloads (only csv downloads). 

I'm not sure why urllib wouldn't work with the URL that redirects though. I only found out it redirects after getting the forbidden error we were getting before, adding a Mozilla user agent header to our url request, getting a 302 redirect error, then using geturl with urlretrieve to find the redirect url. 


To fix this, we can: 
- change the url in the script
- do a bit of cleaning to the url within engine.py to get the final redirected url (this could help all url redirects).
or
- switch over to requests

I think switching over to requests would provide the best long term solution, but might take a little to implement.

ethanwhite:
Thanks @zhangcandrew. Great detective work!

I'd suggest that @shreyneil fix this for now by swapping out the url and that we add switching over to requests as an issue to solve the problem more generally.

shreyneil:
i have swapped the url for now, please review.

scripts/globi-interaction.json
```diff
@@ -0,0 +1,233 @@
+{
+  "citation": "NA",
+  "description": "GloBI contains code to normalize and integrate existing species-interaction datasets and export the resulting integrated interaction dataset.",
+  "homepage": "https://github.com/jhpoelen/eol-globi-data/wiki",
+  "keywords": [
+    "source",
+    "target"
+  ],
+  "name": "globi_interaction",
```
ethanwhite: Name field should be hyphenated, not underscored: `globi-interaction`

scripts/globi-interaction.json
```diff
@@ -0,0 +1,233 @@
+{
+  "citation": "NA",
```
ethanwhite: The citation should be:

Poelen, J.H., Simons, J.D. and Mungall, C.J., 2014. Global biotic interactions: an open infrastructure to share and analyze species-interaction datasets. Ecological Informatics, 24, pp.148-159.

scripts/globi-interaction.json
```diff
@@ -0,0 +1,233 @@
...
+            "type": "char"
+          }       
+          ]
+      },
+      "url": "https://s3.amazonaws.com/globi/snapshot/target/data/tsv/interactions.tsv.gz"
+    }
+ ],
+  "retriever": "True",
+  "retriever_minimum_version": "2.1.dev",
+  "title": "Globi_interaction_data",
```
ethanwhite: Title should be "Global Biotic Interactions (GloBI) data"

shreyneil:
Done with changes.Please review.

ethanwhite:
Perfect. Thanks @shreyneil & @zhangcandrew! In this goes.

Issue #639, opened 2016-09-10
Add plant growth form dataset


ethanwhite:
http://onlinelibrary.wiley.com/doi/10.1002/ecy.1569/full

henrykironde:
@ethanwhite are you working on this now? Just wondering because I was working on some data and i could look at this after?

ethanwhite:
No, I'm not working on it. Just saw it had come out and thought it would be useful to add. No rush on this.

henrykironde:
Ok, thanks. I was baffled by the date 27 August 2016, Cos I was pretty sure that we do not have data papers for 2016 yet.  based on[ Ecological Archives](http://esapubs.org/archive/search.php?journal=E&year=2016&firstAuthor=&isDP=0) wondering why this is not yet in the database here.

ethanwhite:
When the journal Ecology transitioned to being published by Wiley they stopped adding new data papers to Ecological Archives and started posting them as supplemental material on Wiley's site. So, the data for this paper is located at:

http://onlinelibrary.wiley.com/store/10.1002/ecy.1569/asset/supinfo/ecy1569-sup-0001-DataS1.zip?v=1&s=219c74fa85bfe6738e649bedd844f40428351802

I have no idea how stable this url is and in general I think this move to supplementary material is a terrible idea. I've asked about it with higher ups at ESA, but after initially telling me they were looking into it they've gone quite. We should experiment with this url over the next couple of weeks and see if it's stable so we know if we can support these or in their current state or not.

henrykironde:
okey thanks.

henrykironde:
This has issues with data, similar to `The thermal dependence of biological traits` by Dell.
I will look at it and enhance the script later

ethanwhite:
Sounds good. No rush on this one at all.

shreyneil:
Was this solved? and if not how to go about it @ethanwhite sir?

henrykironde:
Not yet. We had issues with this data back then. I had mad some script e9743b5, I am not sure if it would work now.

But you could try and test out a new json script

ethanwhite:
@shreyneil - definitely give it a go if you'd like. Thanks!

henrykironde:
@shreyneil,

```
{
[...]

PR #1102, opened 2018-02-28
Adding plant-growth form dataset

shreyneil:
The Following PR solves https://github.com/weecology/retriever/issues/639 , Please verify and get back.

### DIFF
--- a/scripts/plant-growth.json
+++ b/scripts/plant-growth.json
@@ -1,0 +1,103 @@
+{
+  "citation": "Engemann, K., Sandel, B., Boyle, B. L., Enquist, B. J., Jorgensen, P. M., Kattge, J., McGill, B. J., Morueta-Holme, N., Peet, R. K., Spencer, N. J., Violle, C., Wiser, S. K. and Svenning, J. .-C. (2016), A plant growth form dataset for the New World. Ecology. Accepted Author Manuscript. doi:10.1002/ecy.1569",
+  "description": "This dataset provides growth form classifications for 67,413 vascular plant species from North, Central, and South America.",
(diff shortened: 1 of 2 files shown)


scripts/plant-growth.json
```diff
@@ -0,0 +1,125 @@
...
+  "keywords": [
+    "Taxon",
+    "Plants"
+  ],
+  "name": "plant_growth_form_dataset",
+  "archived": "zip",
+  "resources": [
+    {
+      "path": "DataS1/GrowthForm_Final.txt",
+      "dialect": {
```
henrykironde: Add delimiter` \t`

henrykironde:
For this dataset  we may want to use a custom python script. 
The file `Growthform_Scheme.txt` was compiled in a format that would require some extra effort to clean up. In fact we may get a hard time reading this file using a custom script on both python 2 and 3 but we could try. ref e9743b52

shreyneil:
I realised that as even the encoding for Growthform_Scheme.txt is different as compared to other files.
So, do you want me to write a script for the other two files and Growthform_Scheme.txt seperately, as in , remove the Growthform_Scheme.txt from this PR and Launch another PR for Growthform_Scheme.txt

henrykironde:
Okey create a different PR for `Growthform_Scheme.txt`.   using a custom python file. You can use this as a ref `intertidal-abund-me`. The goal is to read that file line by line and remove/trim the extra spaces. If we can get that to work, we shall add all the tables to that script.

shreyneil:
Done with the changes,working on the Growthform_Scheme.txt part. Meanwhile , please review this one.

PR #1103, opened 2018-02-28
Avoid bulk inserts for all integration tests

ethanwhite:
I'm getting some local failures on MySQL for the few datasets that don't
avoid bulk inserts. It seems unlikely that these tests didn't avoid bulk
inserts on purpose so this makes all integration tests avoid them.


henrykironde:
I think the version of Mysql service used was updated.

PR #1107, opened 2018-03-02
Check test_regression.py

apoorvaeternity:


henrykironde:
@apoorvaeternity, That is very true. We can close this, I will debug it later to find what is really happening. 👍

apoorvaeternity:
Okay :+1:

Issue #151, opened 2014-01-19
Log errors and other issues


ethanwhite:
We should use the `logging` module to log anything that gets caught in an except statement during a Retriever session. This can then be used to help with debugging errors, in particular those that we can't replicate locally.

kapilkd13:
Hi @henrykironde As we were talking about using `logging` module. I would like to add [logging](https://docs.python.org/3/library/logging.html) to our tool. I have gone through the steps involved and what I am thinking is:
1. We can create two handler, for `file` and `stdout`.
2. Our logger object will be set at `debug` by default and will send all logs to both handlers.
3. `file` handler will be set to `debug` and would log everything in a retriever.log file.
4. For each session this log file will be cleared and rewritten.
5. we will put this file's name in `.gitignore`. So whoever files a issue can be requested to upload log file for better understanding.
6. `stdout` handler will be set to `info` or `warning` level which can be changed as per user specified `debug` flag.
What are your thoughts on it @ethanwhite @henrykironde

ethanwhite:
The general plan sounds good to me :+1:

adhaamehab:
@ethanwhite @henrykironde 
I'm working on this issue as discussed on gitter with @henrykironde

piyushchauhan:
Can I take up this issue? and if possible along with #1265  @henrykironde @ethanwhite

henrykironde:
Sure

henrykironde:
I do not think we still need this feature. In case it is needed later, we shall reopen the issue

PR #1110, opened 2018-03-05
[WIP] Logging errors and other issues

adhaamehab:
@ethanwhite @henrykironde 
I started with `__main__.py` , I still working on the remaining files in the project.
Right now I only log exceptions to `retriever.log`

Fixes:  #151

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -18,7 +18,7 @@ from imp import reload
 from retriever.engines import engine_list, choose_engine
 from retriever.lib.datapackage import create_json, edit_json, delete_json, get_script_filename
 from retriever.lib.datasets import datasets, dataset_names, license
-from retriever.lib.defaults import sample_script, CITATION, ENCODING, SCRIPT_SEARCH_PATHS
+from retriever.lib.defaults import sample_script, CITATION, ENCODING, SCRIPT_SEARCH_PATHS, LOGS_PATH
 from retriever.lib.get_opts import parser
 from retriever.lib.repository import check_for_updates
 from retriever.lib.scripts import SCRIPT_LIST, get_script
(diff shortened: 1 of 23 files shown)


henrykironde:
great

retriever/__main__.py
```diff
@@ -31,7 +32,13 @@
...
 
 def main():
-    """This function launches the Data Retriever."""
+    """This function launches the Data Retriever."""  
+    logging.basicConfig(
+        format="%(asctime)s [%(levelname)-5.5s] [%(filename)s:%(lineno)s - %(funcName)3s() ] %(message)s",
+        handlers=[
+            logging.FileHandler("retriever.log"),
+        ])
+    sys.argv[1:] = [arg.lower() for arg in sys.argv[1:]]
```
henrykironde: Remove this line

retriever/__main__.py
```diff
@@ -31,7 +32,13 @@
 
 
 def main():
-    """This function launches the Data Retriever."""
+    """This function launches the Data Retriever."""  
+    logging.basicConfig(
+        format="%(asctime)s [%(levelname)-5.5s] [%(filename)s:%(lineno)s - %(funcName)3s() ] %(message)s",
+        handlers=[
+            logging.FileHandler("retriever.log"),
```
henrykironde: Let's avoid hardcoding this file `retriever.log` since it's universal, we can include it in the defaults `retriever.lib.defaults`

henrykironde:
As we design and implement this feature, we should think about the tests.
I am envisioning a test in the test retriever module, that will trigger a known error and we should be able to check the log file for that particular error or log info

adhaamehab:
Yes I agree with you, I will consider that test and add it once done with adding logging to all the source code.

adhaamehab:
@henrykironde So I added logging to all scripts now. I'll move to work on `test_logging.py`

henrykironde:
Thanks you @adhaamehab for this really great addition, for now we shall close it and reference it in the future when we improve the logging underlying core foundation.

PR #1111, opened 2018-03-06
Geo-spatial Postgres

henrykironde:


henrykironde:
To test you will need postgis extension enabled

 
```
CREATE EXTENSION postgis;
CREATE EXTENSION postgis_topology;
```

Sample test run 

retriever install postgres usgs-elevation -b -94.98704597353938 39.027001800158615 -94.3599408119917 40.69577051867074

henrykironde:
@zhangcandrew could you test this and post what you found out(whenever you get free time). I think am have solved the issues.

zhangcandrew:
The proper error is being displayed!
image

I think I have PostGIS installed, but since it's giving me the error I'm assuming I need to do something more.

henrykironde:
Note: Currently the data link is down

To test this PR please visit the links and check if the data link is available
https://nationalmap.gov/3DEP/
If you can view the map then then it means the api is back to service.

henrykironde:
We can not Load Extensions using python, But we do some slight test to see if postgis is installed. 
We shall add documentation for users to install Postgis and enable the extensions

henrykironde:
Notes:

`On windows`
If you have PostgreSQL installed.  Look for the home directory of PostgreSQL. 
Look for the `stackbuilder.exe` and run it for the right server. 
If you run this executable. It shows you the additional packages that you would want to install for this particular server.
When installing, It will ask you to select components to install, 
- [ ] Posgis
- [ ]  Create spatial database 
You only need Posgis, no need of creating a database.

You will be asked to set your environmental variables for Gdal and some other programs. (i did not set that)
**Errors encountered:**
`Could not load library “C:\…\ rtpostgis-2.2.dll”`

 `http://www.postgresonline.com/journal/archives/364-PostGIS-2.2-Windows-users-hold-off-on-installing-latest-PostgreSQL-patch-release.html`
In my case, I  download the binaries for `9.5.12`  since I am using `9.4`. 

I renamed the current dll with a old tag `libcurl-4-old.dll` just to keep the original and then copied the 
`libcurl.dll` from `9.5/bin` to my postgres's bin
`libcurl-4-old.dll`

zhangcandrew:
On unix you can install postgres/postgis like so: 

#installing postgres
```
sudo add-apt-repository "deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main"
wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
sudo apt update
sudo apt install postgresql-9.6 postgresql-contrib-9.6
```

#installing postgis
```
sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable
sudo apt update
sudo apt install postgis postgresql-9.6-postgis-2.3
sudo -u postgres psql -c "CREATE EXTENSION postgis; CREATE EXTENSION postgis_topology;" gisdata
```

PR #1125, opened 2018-03-31
Make sure table name option is used correctly.

### LINKED ISSUE #1117: Error on custom table name.
The current state of the retriever uses the required_opts  `"table_name"` to customize the installation of dataset into dataset. This works fine for single table datasets however, it over rides the tables when the dataset contains multi tables. The best way to use this option is by using this custom `table_name` as a prefix to the table name. 

For example:

"{db}.{prefix_name}_{table}")

henrykironde:
if -t option is used without the {db}*{table} format,
the retriever will fail to install the data corectly

fixes #1117

### DIFF
--- a/retriever/engines/__init__.py
+++ b/retriever/engines/__init__.py
@@ -1,4 +1,6 @@
 """Contains DBMS-specific Engine implementations."""
+import os
+
 from retriever.lib.engine import Engine
 
 engines = [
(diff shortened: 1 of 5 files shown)


henrykironde:
I feel like this is a hacky way to solve this issue, but it is one of those edge cases.

As mention in the issue #1117, if users follow the format of the required opts for the table-name provided in the engine, we do not get this issue.

_For example_ 
**Mysql and Postgres:** `{db}.{table}`, `{db}.prefix{table}`
**Sqlite:** `{db}_{table}`, `{db}_prefix{table}`

When a user provides something like `retriever install csv portal -t march`, the engine will fail to maintain the format for the table name.

To avoid this, we are getting the default required opt for that engine `{db}_{table}.csv` and replacing the `{table}` value with `prefix{table}`.
The final format will look like `{db}_prefix{table}.csv`

henrykironde:
I will add some multi table tests as I add the fetch function. #1069

zhangcandrew:
How is the -t flag supposed to be correctly used? I can't seem to get it working correctly, but I want to look a bit more into this.

henrykironde:
When using the t tag, you have to follow the format of the required engine values for that engine.

For example installing in to csv

retriever install csv portal -t {db}_{table}.csv
retriever install csv portal -t {db}_new{table}.csv
retriever install csv portal -t {db}_old{table}.csv

zhangcandrew:
Looks good to me, tried it with a few cases and they seemed to work. Merging it in!

Issue #1114, opened 2018-03-15
Switch from urllib to requests


zhangcandrew:
Right now the retriever uses the urllib/urllib2 module, but the requests module has proven itself as a much friendlier url request handler. 

Switching over would save us a lot of trouble/code in the future. 

Reference: http://docs.python-requests.org/en/master/

apoorvaeternity:
Yes the requests module is a lot easier and error handling is also better.

blue-atom:
Can I do this??

henrykironde:
Go for it @blue-atom

blue-atom:
@henrykironde  Thanks for assigning it to me.....

PR #1134, opened 2018-04-07
Repair aquatic animal excretion

henrykironde:


henrykironde:
This needs to wait till we have #1114

PR #1142, opened 2018-05-06
Extract files from archive

henrykironde:


retriever/lib/engine.py
```diff
@@ -38,18 +34,19 @@ class Engine(object):
...
 
-    name = ""
-    instructions = "Enter your database connection information:"
-    db = None
-    table = None
     _connection = None
     _cursor = None
     datatypes = []
-    required_opts = []
+    db = None
```
henrykironde: Sorted these properties and included placeholder

retriever/lib/engine.py
```diff
@@ -75,12 +72,15 @@ def get_connection(self):
         raise NotImplementedError
 
     def add_to_table(self, data_source):
-        """This function adds data to a table from one or more lines specified
+        """Adds data to a table from one or more lines specified
         in engine.table.source."""
         if self.table.columns[-1][1][0][:3] == "ct-":
             # cross-tab data
```
henrykironde: Changed the style to follow pep8. I think it is still easily readable.

retriever/lib/engine.py
```diff
@@ -239,23 +251,25 @@ def auto_get_datatypes(self, pk, source, columns, column_values):
                             val = self.table.cleanup.function(
                                 val, self.table.cleanup.args)
 
-                        if val is not None and val.strip() is not '':
+                        if val and val.strip():
```
henrykironde: Use python python idioms

retriever/lib/engine.py
```diff
@@ -227,7 +238,8 @@ def auto_get_datatypes(self, pk, source, columns, column_values):
         # set default column data types as int
         column_types = [('int',)] * len(columns)
         max_lengths = [0] * len(columns)
-
+        if column_values:
```
henrykironde: This variable is redundant, I will consider revising the the parameters later to find out why I previously included it and never used it. For now I suggest keeping it here for a bit, if that is okey

retriever/lib/engine.py
```diff
@@ -4,12 +4,7 @@
 from future import standard_library
 
 standard_library.install_aliases()
-from builtins import object
-from builtins import range
-from builtins import input
-from builtins import zip
-from builtins import next
-from builtins import str
+from builtins import object, range, input, zip, next, str
```
henrykironde: Note: The changes in this file, from here till line [455](https://github.com/weecology/retriever/pull/1142/files#diff-ff05cd5f49d751e66a61636b498c70acR455), are related to either refactoring and python idioms.

retriever/lib/tools.py
```diff
@@ -3,10 +3,43 @@
 import io
 import os
 import sys
+import struct
 
+from gzip import FEXTRA, FNAME
 from retriever.lib.defaults import  ENCODING
 
 
+def get_gzip_filename(gzip_file):
```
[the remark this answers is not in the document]
henrykironde: I will do. This function gets the name of a gzip file that we downloaded without knowing the file name. Just the same case we saw when dealing with usgs data. We use bounding box but we do not know the exact name of the file in the archive. The library that handles gzip does not provide an attribute to get the name in the archive. We have to read some bytes to get to the location of the name and extract it.

PR #1149, opened 2018-05-28
Refactor extract methods from download from archive

henrykironde:
ref #1142


henrykironde:
@ethanwhite, based on the current schedule, I do not expect any reviews until we get back to the norm. 
The tests took me some time to get right hence the delay. I will create a copy to merge to the dev branch so that we can continue with the next steps for now.

henrykironde:
We are skipping some tests on `appveyor` py.test -v -k "not test_download_archive"

retriever/lib/engine.py
```diff
@@ -438,70 +448,56 @@ def download_file(self, url, filename):
...
+            self.download_file(url, archive_name)
+            if archive_type == 'tar' or archive_type == 'tar.gz':
+                file_names = self.extract_tar(archive_full_path, archive_dir, archive_type)
+            elif archive_type == 'zip':
+                file_names = self.extract_zip(archive_full_path, archive_dir)
+            elif archive_type == 'gz':
+                file_names = self.extract_gz(archive_full_path, archive_dir)
+            return file_names
+
+        if file_names:
```
ethanwhite: This is a little confusing since we define file_names in the previous `if` and if you miss the `return` it looks like this code should always run. I might use an `else` where instead.

PR #1135, opened 2018-04-14
[WIP] Add BUPA Liver Disorders Dataset

ss-is-master-chief:
https://archive.ics.uci.edu/ml/datasets/Liver+Disorders

@henrykironde Please have a look

PR #1150, opened 2018-05-30
Add bupa liver disorders dataset

henrykironde:
Ref: #1135


henrykironde:
Thanks @ss-is-master-chief for working on this dataset.

PR #1157, opened 2018-06-16
Add spatial data for spatial tests.

henrykironde:
harv_landrgb.zip contains raster data(3 bands)
regions_shape.zip contains vector data
regions_shape contains Italian regional data.


zhangcandrew:
Thanks for adding the test data Henry!

PR #1159, opened 2018-06-28
Remove extra incorrect citation from macroalgal comms dataset

KristinaRiemer:


ethanwhite:
Thanks @KristinaRiemer!

@henrykironde any idea why this is failing on Python 3 on AppVeyor? Doesn't seem like it should be an issue with this PR.

henrykironde:
Thanks @KristinaRiemer. Apveyor is failing, I believe they changed some images. My best guess would be the default values of Mysql like autocommit and character set. But this PR looks fine and ready.

PR #1162, opened 2018-06-29
Fix argument name in download in templates.py

apoorvaeternity:
In engine.py the argument name is `file_names` in `download_files_from_archive` so I have fixed it in templates.py.


henrykironde:
Thanks for that catch @apoorvaeternity

PR #1167, opened 2018-07-16
Update breed-bird-survey script to match new data structure

ethanwhite:
The USGS folks added a new field to the main table. This is a breaking
change and fits the definition of a major patch version in the frictionless
data specs, so I've bumped the version to 2.0.0


henrykironde:
Looks good to me. Apveyour wont pass on this PR for a known reason.

ethanwhite:
Thanks @henrykironde!

PR #1173, opened 2018-07-25
Documentation refinement

pranita-s:


docs/rdataretriever.rst
```diff
@@ -25,12 +25,219 @@ To install the development version, use ``devtools``
...
+as shown below:
+
+:: 
+
+  library(rdataretriever)
+  use_RetrieverPath("/home/<system_name>/.local/bin/")
+
+The ``which`` command in the terminal finds the exact location of ``Retriever``, but the path
+required by the function is actually the directory which contains ``Retriever``. Therefore, the
+final folder from the path (*retriever* in this case) needs to be removed before using it.
```
ethanwhite: I would rephrase to:

```
The ``which`` command in the terminal finds the location of ``retriever`` including the name
of the program, but the path required by the function is the directory that contains ``retriever``.
Therefore, the `retriever` needs to be removed from the path before using it.
```

docs/rdataretriever.rst
```diff
@@ -25,12 +25,219 @@ To install the development version, use ``devtools``
...
+when that particular environment is activated. To illustrate, assume the virtual environment is *py27*:
+
+::
+
+  $ conda activate py27
+  (py27) $ which retriever
+  /home/<system_name>/anaconda2/envs/py27/bin/retriever
+
+After the location has been found out, it can be used for ``rdataretriever`` after removing last directory
+from the path as follows:
```
ethanwhite: Rephrase to:

```
This path can be used for ``rdataretriever`` after removing `retriever` as follows:
```

docs/rdataretriever.rst
```diff
@@ -25,12 +25,219 @@ To install the development version, use ``devtools``
...
+**Example** :
+
+::
+
+  rdataretriever :: fetch(dataset = 'portal')
+
+download()
+^^^^^^^^^^
+**Description** : Used to download datasets directly without cleaning them and when user does not 
+have specific preference for format of data and the kind of database.
```
ethanwhite: "have a specific preference for the format of the data..."

docs/rdataretriever.rst
```diff
@@ -25,12 +25,219 @@ To install the development version, use ``devtools``
...
+
+**Example** :
+
+::
+
+  rdataretriever :: get_updates()
+
+reset()
+^^^^^^^
+**Description** : The function will Reset the components of rdataretriever using scope [ all, scripts, data, connection]
```
ethanwhite: We've dropped the storing of connection information in the core retriever. If we're still doing so in `rdataretriever` this is fine. If we've dropped it there as well then we should drop references to `connection`.

docs/rdataretriever.rst
```diff
@@ -56,3 +263,90 @@ To get citation information for the ``rdataretriever`` in R use ``citation(packa
 
 
 .. _Data Retriever: http://data-retriever.org
+
+
+RDataRetriever with PostgreSQL
```
ethanwhite: No capitals in `rdataretriever` here and below`

docs/rdataretriever.rst
```diff
@@ -56,3 +263,90 @@ To get citation information for the ``rdataretriever`` in R use ``citation(packa
...
+==============================
+
+RDataRetriever supports different DBMS for storing of downloaded datasets through the Retriever R API. It supports PostgreSQL, MySQL, SQLite and MS Access.
+
+
+Installation of PostgreSQL in Ubuntu
```
ethanwhite: This is really useful information, but I'm not sure if information on installing postgres into Ubuntu is within scope for our documentation. It seems like if we're going to include this we'd also want to include other DBMSs and OSs. What do you think @pranita-s & @henrykironde?

henrykironde:
@pranita-s I have open a PR pranita-s#2 to do more clean up. Merge it and it will ready for a reviewed.

docs/rdataretriever.rst
```diff
@@ -56,3 +263,90 @@ To get citation information for the ``rdataretriever`` in R use ``citation(packa
...
+
+>>> createdb -U vmsadmin vms;
+
+You will be again prompted for password. After the successful setup of PostgreSQL, it can now be used with R API for Retriever.
+
+Using PostgreSQL with RDataRetriever
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+While using PostgreSQL as a connection type for ``rdataretriever::install`` function, a file named ``postgres.conn`` is needed. It contains information for establishing connection with the requested DBMS, in this case, PostgreSQL. Default location of the file is the directory, through which RStudio is running. If it saved in some other location, its path needs to be given to the install function.
+In the above example, ``postgres.conn`` will look like below:
```
ethanwhite: The core retriever is now handling this through the official `.pgpass` file used by postgres. If this is currently also working in `rdataretriever`, then I think we should remove mention of this old approach in the docs. If it isn't then we can definitely leave this in for now, but it would be good to get `rdataretriever` using `.pgpass` instead.

docs/rdataretriever.rst
pranita-s: @ethanwhite @henrykironde I can include information for other DBMSs and OSs too. If there are any content enhancements needed, taking current information for "Installation for PostgreSQL" into consideration, it will be great to know !

docs/rdataretriever.rst
pranita-s: Yes, the ``rdataretriever`` is using ``.pgpass`` instead of ``conn`` file. I will make sure, the previous information is edited and also fulfill other requirements.

docs/rdataretriever.rst
ethanwhite: I don't think I see anything that needs to be added in general, but I also think it would be find to link to other installation guides and just mention anything that is specific to the retriever that needs to be done.

ethanwhite:
Hey @pranita-s - just wanted to check in on this. It's totally fine to not include installation instructions for all of the different database management systems, so if you have time to do the more minor cleanup then we can go ahead and get this in.

ethanwhite:
Thanks @pranita-s! The changes to `rdataretriever.rst` look great, but it looks like a bunch of other files got accidentally added to this commit as well. If you can clean up and rebase the PR I can get this in. Alternatively you could submit a fresh PR or I can clean things up locally and merge them that way. Just let me know what you'd like to do.

pranita-s:
@ethanwhite I will submit a new PR. Thank you !

henrykironde:
ref #1191

Issue #1174, opened 2018-07-26
Warnings in current release from new changes


ethanwhite:
```
ethan@omar:~$ retriever --version
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
Exception: name 'gz' is not defined 
Failed to load script: harvard_forests (/home/ethan/.retriever/scripts/)
Exception: invalid syntax (harvard_forests.py, line 14) 
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
Exception: name 'gz' is not defined 
Failed to load script: harvard_forests (/home/ethan/.retriever/scripts/)
Exception: invalid syntax (harvard_forests.py, line 14) 
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
Exception: name 'gz' is not defined 
Failed to load script: harvard_forests (/home/ethan/.retriever/scripts/)
Exception: invalid syntax (harvard_forests.py, line 14) 
v2.1.0
```

Will bumping the minimum version to `2.1.dev` fix this?

henrykironde:
Have you tried removing retriever reset scripts.

ethanwhite:
I did. Are you not seeing the same thing with 2.1.0?

```
ethan@omar:~$ retriever reset scripts 
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
Exception: name 'gz' is not defined 
Failed to load script: harvard_forests (/home/ethan/.retriever/scripts/)
Exception: invalid syntax (harvard_forests.py, line 14) 
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
Exception: name 'gz' is not defined 
Failed to load script: harvard_forests (/home/ethan/.retriever/scripts/)
Exception: invalid syntax (harvard_forests.py, line 14) 
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
Exception: name 'gz' is not defined 
Failed to load script: harvard_forests (/home/ethan/.retriever/scripts/)
Exception: invalid syntax (harvard_forests.py, line 14) 

This will remove existing scripts. 
Specifically it will remove the scripts folder in /home/ethan/.retriever.
Do you want to proceed? (y/N)
y
ethan@omar:~$ retriever --version
Downloading scripts...
Download Progress: [#######################################################################################################################################################################################################################################################] 100.00%
The retriever is up-to-date
Failed to load script: globi_interaction (/home/ethan/.retriever/scripts/)
[...]

PR #1176, opened 2018-07-26
Repair script backward compatibility.

henrykironde:
Harvard forest is a spatial dataset and not available in retriever version <=2.1.0.

Globi Interaction's `gz` format is not support in json in <=2.1.0.

This change excludes them from versions < 2.1.dev to avoid error messages. They will still appear and install in `master` and following releases.


ethanwhite:
Nice work solving this one @henrykironde. Closes #1174.

PR #1177, opened 2018-07-27
Fix and improve biotime script

ethanwhite:
* Fix off-by-one column error due to first column in query table lacking a heading
* Update types to match data more precisely and include large enough fields
* Properly handle NA missing values
* Update to newest data files


ethanwhite:
Good catch. I removed the comment instead so that this would stay available to the current release.

PR #1185, opened 2018-08-05
FIA url update

henrykironde:
FIA changed resource extensions from .ZIP
to .zip on August 4 2018

Change of PATCH version as per frictionlessdata versioning.


ethanwhite:
Once a year or so. Seems about right 😄

PR #1189, opened 2018-08-16
RDataRetriever documentation cleaned and updated

pranita-s:


ethanwhite:
Thanks for this @pranita-s. It looks great!

PR #1190, opened 2018-08-17
Add path parameter to to_csv()

apoorvaeternity:


henrykironde:
How about using path=None , I think it is better than path=' '

retriever/engines/jsonengine.py
```diff
@@ -117,11 +117,13 @@ def table_exists(self, dbname, tablename):
         tablename = self.table_name(name=tablename, dbname=dbname)
         return os.path.exists(tablename)
 
-    def to_csv(self, sort=True):
+    def to_csv(self, sort=True, path=None):
         """Export table from json engine to CSV file"""
         for table_item in self.script_table_registry[self.script.name]:
             header = table_item[1].get_insert_columns(join=False, create=True)
-            csv_outfile = json2csv(table_item[0], header_values=header)
+            outputfile = os.path.normpath(
```
henrykironde: Am thinking that we should test `if path:` somewhere.

retriever/lib/engine.py
```diff
@@ -129,11 +129,11 @@ def add_to_table(self, data_source):
                                 self.table.cleanup.args),
                             types[n])
                         for n in range(len(line_values))
-                        ]
+                    ]
```
henrykironde: @apoorvaeternity, Do not change. Looks like the indention is off

retriever/lib/engine.py
```diff
@@ -129,11 +129,11 @@ def add_to_table(self, data_source):
...
                 except Exception as e:
                     self.warning(
                         'Exception in line {}: {}'
-                        .format(self.table.record_id, e))
+                            .format(self.table.record_id, e))
```
henrykironde: Do not change. Looks like the indention is off

retriever/lib/engine.py
```diff
@@ -275,7 +275,7 @@ def auto_get_datatypes(self, pk, source, columns):
                                     val = int(val)
                                     if column_types[i][0] == 'int' and \
                                             hasattr(self, 'max_int') and \
-                                            val > self.max_int:
+                                                    val > self.max_int:
```
henrykironde: Do not change. Looks like the indention is off

PR #1203, opened 2018-10-31
Change docs toc

sdtaylor:
The retriever docs were always a bit confusing to me. I think changing around the TOC a bit and combining the dev stuff into a single page helps out.


henrykironde:
@sdtaylor, thanks for the changes 👍

PR #1204, opened 2018-10-31
Add note on using system Python for building DEB

ethanwhite:
The `build.sh` script doesn't work properly with conda. This commit
adds a note to notify the developer of this and recommend ways to
avoid the problem if conda is the first Python on the path.

### DIFF
--- a/docs/Release.rst
+++ b/docs/Release.rst
@@ -25,6 +25,10 @@ Start
 Linux
 -----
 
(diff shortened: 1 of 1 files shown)


ethanwhite:
I'm happy to rebase this on #1203 if we decide to merge those changes.

ethanwhite:
Replaced by #1206

PR #1209, opened 2018-11-06
Test builds

henrykironde:


henrykironde:
The failing looks like it is from the new release of pytests

PR #1210, opened 2018-11-11
Post release version update

henrykironde:


henrykironde:
@ethanwhite am not sure of the target version number.

ethanwhite:
This is fine for now. I suspect we'll end up going to 2.3, but it's always hard to know in advance, so any increment is fine.

Issue #1229, opened 2019-02-03
Restart required if scripts directory is empty


ethanwhite:
If the `.retriever/scripts` directory is empty when starting a Python session the retriever reports that there are no scripts even after successfully running `check_for_updates`.

```python
In [1]: import retriever                                                                                                                                                                                           

In [2]: retriever.dataset_names()                                                                                                                                                                                  
Out[2]: []

In [3]: retriever.check_for_updates()                                                                                                                                                                              
Downloading scripts: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 97/97 [00:33<00:00,  2.99files/s]

In [4]: retriever.dataset_names()                                                                                                                                                                                  
Out[4]: []
```

If you restart Python then everything works as expected.

See https://github.com/ropensci/rdataretriever/issues/173 for initial report.

ethanwhite:
This results from 398ca50484d290c59eecf0d64b9e1d5c4bc0c7ce. In general this commit is great - we don't want to constantly be reloading scripts, but we do need to be able to after we've checked for updates. I played with this a bit, but I found the current structure of the code related to `global_script_list`, `StoredScripts`, and `SCRIPT_LIST` to be a bit confusing and couldn't figure out how to call for a reloading of the scripts from `check_for_updates`.

@zhangcandrew - do you have any thoughts on how to do this?

henrykironde:
I also came across that https://github.com/henrykironde/retriever/issues/117
I think Andrew will be able to figure that out.

ethanwhite:
:+1: We should fix this and roll a new release before we release the new version of `rdataretriever` because this issue will happen for all first time users on R. It's confusing enough that I think it's worth fixing before we generate new users through advertising around the release.

zhangcandrew:
I think this should be fixed in #1231 

I agree, the structure is a bit confusing. We ran into the problem because reload_scripts was being called by the singleton on instantiation before check_for_updates was called later on in the code. As a result, the global singleton scripts were set to nothing after a retriever reset. 

I've made it now so that upon instantiation of the SCRIPT_LIST it will do a check_for_updates if necessary. I think this should work.

zhangcandrew:
fixed in #1231

PR #1231, opened 2019-02-05
fixes to retriever ls after reset

zhangcandrew:
I think this should fix the problems we've been having with ls after the retriever reset. 

Basically what was happening was the singleton class was getting created before doing a check for update which actually downloads everything. 

In this new format, we move the initial check for update logic into the singleton class instantiation. 

This is introduced in relation to #1229

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -27,14 +27,15 @@ def main():
 
     else:
         # otherwise, parse them
-
-        if not os.path.isdir(SCRIPT_SEARCH_PATHS[1]) and not \
-                [f for f in os.listdir(SCRIPT_SEARCH_PATHS[-1])
-                 if os.path.exists(SCRIPT_SEARCH_PATHS[-1])]:
-            check_for_updates()
-        script_list = SCRIPT_LIST()
-
         args = parser.parse_args()
+
+        if args.command not in ['reset', 'update'] \
+        and not os.path.isdir(SCRIPT_SEARCH_PATHS[1]) \
+        and not [f for f in os.listdir(SCRIPT_SEARCH_PATHS[-1])
+            if os.path.exists(SCRIPT_SEARCH_PATHS[-1])]:
+                check_for_updates()
+                reload_scripts()
+        script_list = SCRIPT_LIST()
 
         if args.command == "install" and not args.engine:
             parser.parse_args(['install', '-h'])
@@ -57,8 +58,8 @@ def main():
             return
 
         if args.command == 'update':
-            check_for_updates(False)
-            script_list = SCRIPT_LIST()
+            check_for_updates()
+            reload_scripts()
             return
 
         elif args.command == 'citation':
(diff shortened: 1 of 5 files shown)


ethanwhite:
Thanks @zhangcandrew. I see one use case that this might not cover and that's updating to get new scripts. I haven't tested this, but reading the code it looks to me like if I get an up-to-date set of scripts now and then in two months I see there's a new script that's been added that I want use. So, I open Python, run `check_for_updates` and then try to install the new script. I am correct in thinking that this new script won't be available for installation without a restart?

zhangcandrew:
I think you're partially correct. Everytime you open python and import retriever, check_for_updates will run and the scripts will be reloaded, so new scripts should always show up on a new session. If a new script is added, when you open python and import retriever, it should be shown there already. 

In the event that new scripts are added during your python session, we don't currently have a function to reload the scripts after doing a check_for_updates(), so I can add that into the branch. 

I'm wondering, was the intended use for importing retriever to always have the scripts be up to date or to only update when the user manually checks for updates and reloads the scripts?

ethanwhite:
Ah, OK, so `check_for_updates` is running at every startup. That alleviates my concern. It used to be that `check_for_updates` only ran on request and so that could have set up some issues.

I'm a little torn on the right behavior here. Running `check_for_updates` every time the package is imported makes the most sense for most users. It means if a dataset breaks and we fix the script then the user will automatically get the updated script, which is good. I initially thought that this was different from the behavior of most package managers, but I'm not sure it is. If I run `pip install retriever` I get the most up-to-date version. Apt is a bit different in that there is a separate `update` command.

Anyway, I'm leaning towards the approach of running `check_for_updates` on import, but would be curious to hear what you, @henrykironde, and others think is the right approach here.

zhangcandrew:
Sorry I think I misspoke. So lines 171-172 in scripts.py currently make it so that check_for_updates is only run when there is no scripts directory, but I can take that if statement out to make it run everytime. 

On the topic of running it everytime - I agree with running it on startup or import everytime. I think there is a use case for making that process manual (not run on startup), but it seems to be very specific (wanting outdated data or data from an old schema), and ultimately something that would probably be best bundled with the provisioning project. That way we can always provide the newest data to our majority user base, but still have the option of using old data for specific users instead of the other way around.

ethanwhite:
> Sorry I think I misspoke. So lines 171-172 in scripts.py currently make it so that check_for_updates is only run when there is no scripts directory, but I can take that if statement out to make it run everytime.

OK, thanks. That's what I thought I'd read. So in this case if new scripts get updated after the initial install then the user would only see new scripts if they updated and then restarted, which is what we want to avoid.

> On the topic of running it everytime - I agree with running it on startup or import everytime. I think there is a use case for making that process manual (not run on startup), but it seems to be very specific (wanting to keep outdated data or data from an old schema), and ultimately something that would probably be best bundled with the provisioning project. That way we can always provide the newest data to our majority user base, but still have the option of using old data for specific users instead of the other way around.

This sounds right to me. If @henrykironde agrees then I'd say update the PR so that `check_for_updates` always runs on load and then we'll get this merged.

henrykironde:
LGTM. Thanks @zhangcandrew and @ethanwhite 
The only issues I think could be rewording the line `print('No scripts are currently available. Creating scripts folder...')` > `print('No scripts are currently available. Updating scripts...')`
and refactoring this code so the condition operators come last, `and not \`

```python
if args.command not in ['reset', 'update'] and not \
```

```python
        if args.command not in ['reset', 'update'] \
        and not os.path.isdir(SCRIPT_SEARCH_PATHS[1]) \
        and not [f for f in os.listdir(SCRIPT_SEARCH_PATHS[-1])
            if os.path.exists(SCRIPT_SEARCH_PATHS[-1])]:
                check_for_updates()
                reload_scripts()
```
But am fine with leaving it like this.

henrykironde:
I will go ahead and merge since I do need to test some more features . We can revisit incase any changes are suggested.

Issue #1232, opened 2019-02-06
UnboundLocalError when deleting an item from keywords key in json scripts


harshitbansal05:
I tried deleting an item from the list of the key `keywords` in the dataset scripts, but it threw an exception `UnboundLocalError: local variable 'del_key' referenced before assignment`.

harshitbansal05:
@henrykironde if possible, I would like to work on it.

PR #1234, opened 2019-02-06
Solve minor delete key bug Fix: 1232

harshitbansal05:
This pull request aims to solve #1232. It solves the minor error while deleting values from the `keywords` list.

### DIFF
--- a/retriever/lib/datapackage.py
+++ b/retriever/lib/datapackage.py
@@ -356,7 +356,8 @@ def edit_dict(obj, tabwidth=0):
                         if del_val not in obj[key]:
                             print("Invalid value: Not found.")
                             continue
-                        print("Removed " + str(obj[key].pop(del_key)))
+                        obj[key].remove(del_val)
+                        print("Removed " + str(del_val))
 
                     elif selection == '3':
                         do_remove = clean_input(


henrykironde:
@harshitbansal05 which python version are you using?

henrykironde:
I think this was from the bug that Andrew is currently fixing because both methods do slightly the same thing.
pop calls `del`, but it first gets the value to be removed as the return deleted value
`removed = obj[key]; del obj[key]; return removed`
let me know what you think. We could alternatively wait till Andrew finishes up with fixing the bug and then retest this.

harshitbansal05:
Hi, @henrykironde I was using python 3, but I think this problem is independent of the python version. IMO, the error is that the variable `del_key` is not defined in that scope. I think someone has accidentally written it in the place of `del_val`. Also, `pop` takes the index of the item to be deleted from the list, while `remove` takes the value as the argument, and deletes the first occurrence of that value. Reference: https://stackoverflow.com/questions/11520492/difference-between-del-remove-and-pop-on-lists/11520540

retriever/lib/datapackage.py
```diff
@@ -356,7 +356,8 @@ def edit_dict(obj, tabwidth=0):
                         if del_val not in obj[key]:
                             print("Invalid value: Not found.")
                             continue
-                        print("Removed " + str(obj[key].pop(del_key)))
+                        obj[key].remove(del_val)
```
henrykironde: Can you check the class of obj using print(class(obj));
This seems like a dictionary not a list.

harshitbansal05: Yes, @henrykironde `obj` is a dictionary.

henrykironde: I do not think remove will work since it is a dictionary

harshitbansal05: @henrykironde but, `obj[key]` is a list, not a dictionary.

harshitbansal05: @henrykironde have you tried removing an item from the `keywords` list of a dataset? e.g, the value `plants` from the database iris.

henrykironde: @harshitbansal05, I have not tried running that part of code or tested it, I just read through the code. And it seems you are deleting something from a dictionary. I will have to run it and see what we are doing but it is clear that the initial value `del_key` was out of scope / not declared anywhere.

zhangcandrew: I think @harshitbansal05 has it right in this correction. obj is a dictionary of key->list mappings, so obj[key] will return a list. the pop function is used only for indices, so we should use remove to delete the value.

zhangcandrew:
Otherwise, I think this bug fixes what it sets out to do.

PR #1237, opened 2019-02-09
Shift from latin-1 to utf-8 in tests_retriever.py

henrykironde:
Pr passed, a merge failed for PR #1236 and #1231
Tests passed locally on mac.

ref:

https://github.com/weecology/retriever/compare/f12a2c5afcaf...16d4d0cf07d0

https://github.com/weecology/retriever/compare/b5a34a784b46...f12a2c5afcaf

PR #1238, opened 2019-02-09
Use tqdm 4.30.0

henrykironde:
Looks like this has solved the issue 
ref: #1237


zhangcandrew:
Just pulled it down and checked it down. Looks good to me!

Issue #1241, opened 2019-02-15
Add reset dataset option


harshitbansal05:
Currently, the `retriever reset` only works for all the datasets or scripts. There should be an option to reset a specific dataset or script as well.

PR #1242, opened 2019-02-15
Add reset specific dataset or script functionality Fixes: #1241

harshitbansal05:
This pull request aims to solve #1241. The reset functionality can now be used to delete a specific dataset and script as well. Sample usage: `retriever reset [all/data/scripts/dataset]`. The script, as well as the raw data of the dataset, would be deleted.

### DIFF
--- a/retriever/lib/engine_tools.py
+++ b/retriever/lib/engine_tools.py
@@ -94,11 +94,8 @@ def final_cleanup(engine):
     pass
 
 
-config_path = os.path.join(HOME_DIR, 'connections.config')
-
-
 def reset_retriever(scope="all", ask_permission=True):
-    """Remove stored information on scripts, data, and connections."""
+    """Remove stored information on scripts and data."""
     warning_messages = {
         'all': "\nThis will remove existing scripts and cached data."
                "\nSpecifically it will remove the scripts and raw_data folders"
@@ -112,21 +109,39 @@ def reset_retriever(scope="all", ask_permission=True):
     }
 
     path = os.path.normpath(HOME_DIR)
-    warn_msg = warning_messages[scope].format(path)
-    if ask_permission:
-        confirm = input(warn_msg)
-        while not (confirm.lower() in ['y', 'n', '']):
-            print("Please enter either y or n.")
-            confirm = input()
+    rw_dir = os.path.normpath(os.path.join(HOME_DIR, 'raw_data'))
+    sc_dir = os.path.normpath(os.path.join(HOME_DIR, 'scripts'))
+    if scope in ['all', 'scripts', 'data']:
+        warn_msg = warning_messages[scope].format(path)
+        if ask_permission:
+            confirm = input(warn_msg)
+            while not (confirm.lower() in ['y', 'n', '']):
+                print("Please enter either y or n.")
+                confirm = input()
+        else:
+            confirm = 'y'
+        if confirm.lower() == 'y':
+            if scope in ['data', 'all']:
+                if os.path.exists(rw_dir):
+                    shutil.rmtree(rw_dir)
+            if scope in ['scripts', 'all']:
+                if os.path.exists(sc_dir):
+                    shutil.rmtree(sc_dir)
     else:
-        confirm = 'y'
-    if confirm.lower() == 'y':
-        if scope in ['data', 'all']:
-            if os.path.exists(os.path.join(path, 'raw_data')):
-                shutil.rmtree(os.path.join(path, 'raw_data'))
-        if scope in ['scripts', 'all']:
-            if os.path.exists(os.path.join(path, 'scripts')):
-                shutil.rmtree(os.path.join(path, 'scripts'))
+        dataset_path = os.path.normpath(os.path.join(rw_dir, scope))
+        if os.path.exists(dataset_path):
+            shutil.rmtree(dataset_path)
+        script = scope.replace('-', '_')
+        script_path_py = os.path.normpath(os.path.join(sc_dir, script + ".py"))
+        script_path_json = os.path.normpath(os.path.join(sc_dir, script + ".json"))
+        if os.path.exists(script_path_py):
+            os.remove(script_path_py)
+            print("successfully removed the script {scp}".format(scp=scope))
+        elif os.path.exists(script_path_json):
+            os.remove(script_path_json)
+            print("successfully removed the script {scp}".format(scp=scope))
+        else:
+            print("can't find script {scp}".format(scp=scope))
 
 
 def json2csv(input_file, output_file=None, header_values=None):
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -70,8 +70,8 @@ citation_parser.add_argument('dataset', help='dataset name', nargs='?', default=
 license_parser.add_argument('dataset', help='dataset name', nargs='?', default=None, choices=script_list + [None])
 new_parser.add_argument('filename', help='new script filename')
 edit_json_parser.add_argument('dataset', help='dataset name', choices=json_list)
-reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections',
-                          choices=['all', 'scripts', 'data', 'connections'])
+reset_parser.add_argument('scope', help='things to reset: all, scripts or data').completer = \
+    ChoicesCompleter(script_list + ['all', 'scripts', 'data'])
 install_parser.add_argument('--compile', help='force re-compile of script before downloading', action='store_true')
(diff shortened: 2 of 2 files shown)


henrykironde:
Could we add some tests cases for each scenario

retriever/lib/engine_tools.py
```diff
@@ -97,12 +97,9 @@ def final_cleanup(engine):
 config_path = os.path.join(HOME_DIR, 'connections.config')
 
 
-def reset_retriever(scope="all", ask_permission=True):
+def reset_retriever(data, script, ask_permission=True):
```
ethanwhite: This is a backwards compatibility breaking change. Alternatively we could use an API like:

`def reset_retriever(scope="all", ask_permission=True, dataset=None):`

and then to reset a dataset call this function as:

`reset_retriever("dataset", dataset = "mammal-masses")`

I doubt `ask_permission` is in active use based on position, so we could also the more natural:

`def reset_retriever(scope="all", dataset=None, ask_permission=True,):`

Allowing:

`reset_retriever("dataset", "mammal-masses")`

harshitbansal05: sorry, @ethanwhite I was a bit occupied, so could not reply. So I would do it in the manner you stated.

harshitbansal05:
@ethanwhite @henrykironde @zhangcandrew could you please review it now?

retriever/lib/engine_tools.py
```diff
@@ -112,21 +112,30 @@ def reset_retriever(scope="all", ask_permission=True):
     }
 
     path = os.path.normpath(HOME_DIR)
-    warn_msg = warning_messages[scope].format(path)
-    if ask_permission:
-        confirm = input(warn_msg)
-        while not (confirm.lower() in ['y', 'n', '']):
-            print("Please enter either y or n.")
-            confirm = input()
+    if scope in ['all', 'scripts', 'data']:
```
henrykironde: Thanks @harshitbansal05, let us reduce the hard coding here ;
We can either create the path values at  144 
```python
    rw_dir = os.path.normpath(os.path.join(HOME_DIR, 'raw_data'))
    sc_dir = os.path.normpath(os.path.join(HOME_DIR, 'scripts'))
    script_py_form = os.path.normpath(os.path.join(sc_dir, "{script}.py"))
    script_json_form  = os.path.normpath(os.path.join(sc_dir, "{script}.json"))
    # Or after... script = scope.replace('-', '_')
    script_path_py = os.path.normpath(os.path.join(sc_dir, script + ".py"))
    script_path_json = os.path.normpath(os.path.join(sc_dir, script + ".json"))
```
 
or decide to 
import 
`from retriever.lib.defaults import HOME_DIR, SCRIPT_WRITE_PATH, DATA_WRITE_PATH`
In this case the rw_dir is same as DATA_WRITE_PATH and sc_dir = SCRIPT_WRITE_PATH.
If you decide to import make sure that there are no cyclic imports caused.
I think both ways are good enough

retriever/lib/get_opts.py
```diff
@@ -70,8 +70,8 @@
 license_parser.add_argument('dataset', help='dataset name', nargs='?', default=None, choices=script_list + [None])
 new_parser.add_argument('filename', help='new script filename')
 edit_json_parser.add_argument('dataset', help='dataset name', choices=json_list)
-reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections',
-                          choices=['all', 'scripts', 'data', 'connections'])
+reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections').completer = \
+    ChoicesCompleter(script_list + ['all', 'scripts', 'data', 'connections'])
```
henrykironde: Remove the `connections`, This is part is no longer supported.
Also line `https://github.com/weecology/retriever/pull/1242/files#diff-09b72f49b55e7226a62ae92a8a825f82L97`

harshitbansal05:
@henrykironde could you review it now?

retriever/lib/get_opts.py
```diff
@@ -70,8 +70,8 @@
 license_parser.add_argument('dataset', help='dataset name', nargs='?', default=None, choices=script_list + [None])
 new_parser.add_argument('filename', help='new script filename')
 edit_json_parser.add_argument('dataset', help='dataset name', choices=json_list)
-reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections',
-                          choices=['all', 'scripts', 'data', 'connections'])
+reset_parser.add_argument('scope', help='things to reset: all, scripts, data, or connections').completer = \
```
henrykironde: remove connections in comments

retriever/lib/engine_tools.py
```diff
@@ -94,9 +94,6 @@ def final_cleanup(engine):
     pass
 
 
-config_path = os.path.join(HOME_DIR, 'connections.config')
-
-
 def reset_retriever(scope="all", ask_permission=True):
     """Remove stored information on scripts, data, and connections."""
     warning_messages = {
```
henrykironde: """Remove stored information on scripts, data, and connections."""

henrykironde:
Added some comments, I am making final tests on my side on the PR

harshitbansal05:
@henrykironde, if the tests have passed, can I make the changes and force push it?

henrykironde:
For this PR, since there are not many changes and the reviews are all covered, it is ok to force push as part of cleaning up. Usually it is not good to force push but its ok to perform that at last as a clean up of the commits

harshitbansal05:
@henrykironde in general, if I want to make more changes and the pull request must be of a single commit, what other alternatives are present apart from force push?

henrykironde:
I think we are on the same page, I just wanted to emphasize that as we aim for one commit per issue, We can always add more commits as we develop(if there are many reviews, comments or changes) and then squash them when all reviews are ready. In fact, I want to think that we don't have to worry about a force push. GitHub always gives us the forced push history . So go ahead and push

henrykironde:
@zhangcandrew, this looks good to me after tests. @harshitbansal05 will changes the `connection` in comments  and it will be ready to go.

One small issue that we may under look, is trying to delete a script that is not available.
retriever reset no-sciprt-called-this 
I am ok with leaving it like this , or trying to check if that script actually exists. Both ways are fine with me

zhangcandrew:
Pulled it down, everything looks good. I agree with @henrykironde  about some output for no scripts. I think in addition to that, you might also want to look into outputting on a successful reset. This delete is a special case since the final reset keyword we are inputting is the name of a script, so it's possible to be run even if a script name is misspelled. 

Otherwise, it looks good!

harshitbansal05:
@zhangcandrew @henrykironde I have made the suggested changes. Could you please review it now?

henrykironde:
as we review this, "The dataset %s does not exists." % scope) >"The dataset %s does not exist" % scope)

harshitbansal05:
@henrykironde are there any more changes, or this is the only one?

henrykironde:
I think that is all. We shall have to test for cyclic dependencies since this function is importing SCRIPT_LIST `from retriever.lib.scripts import SCRIPT_LIST` and could be imported somewhere else.
Or we could just drop the check on script availability and just go with the first option. Let us wait and see what @zhangcandrew advices us to do.

harshitbansal05:
@zhangcandrew could you please comment on the above?

henrykironde:
@harshitbansal05, Andrew may not be available for a few days. So I suggest you remove the [part that checks scripts and](https://github.com/weecology/retriever/compare/fb69d035e42ef546287264b59acf648dfabfff0b..fedaa6d0e0600aee0d1339cc091689acd8ddea9e) I will get this in. We could create a new PR with the check scripts and we wait for what others think

henrykironde:
Alternatively, we could do something like this 

```python
if os.path.exists(script_path_py):
    os.remove(script_path_py)
else:
    print ("can't find script {scp}".format(scp=) )

if os.path.exists(script_path_json):
    os.remove(script_path_json)
 else:
    print ("can't find script {scp}".format(scp=) )
```
And let us use  " {}".format instead of "%"

harshitbansal05:
@henrykironde, imo the above method would also output the message when either of the json or py files does not exist, which would be slightly confusing to the user.

zhangcandrew:
Hey @harshitbansal05, my bad I was out for a bit. It shouldn't be a problem if it outputs either the json or py files (some scripts are made using json and some using python). To the end user, this distinction doesn't really matter.

harshitbansal05:
@zhangcandrew could you please review it now?

zhangcandrew:
Thanks @harshitbansal05! Looks good to me. Merging it in.

Issue #1137, opened 2018-04-15
Add a `directory` or `path` argument to flat file installs


ethanwhite:
It's not uncommon to want to install data files into a specific directory. We currently support this through the `table_name` function and `-t` CLI flag, but this requires being familiar with the somewhat tricky table name syntax, e.g.,

`retriever install csv mammal-masses -t datadir/{db}_{table}.csv`

A simpler and more common version of this would be something like

`retriever install csv mammal-masses -d datadir`

Related to #1117 and #1136.

harshitbansal05:
@ethanwhite can I work on this one?

ethanwhite:
Yes @harshitbansal05 that would be great!

ashishpriyadarshiCIC:
@harshitbansal05 We may to work together to solve the issues #1136 and this one. Can you tell me where we should communicate to work.

harshitbansal05:
@ashishpriyadarshiCIC I have tried to fix this issue and pushed it to my fork(the above branch). You can pull this and work on #1136 and open a pull request on my branch.

ashishpriyadarshiCIC:
@ethanwhite @harshitbansal05 I've fixed issue #1136 . Should I open a pull request here or @harshitbansal05's repository ?

PR #1245, opened 2019-02-18
Add path argument for altering install directories

harshitbansal05:
This pull request aims to fix: #1137. It adds an optional `data_dir` argument to provide a more straight forward custom directory for installation of datasets.

### DIFF
--- a/retriever/engines/csvengine.py
+++ b/retriever/engines/csvengine.py
@@ -26,7 +26,10 @@ class engine(Engine):
     required_opts = [
         ("table_name",
          "Format of table name",
-         os.path.join(DATA_DIR, "{db}_{table}.csv")),
+         "{db}_{table}.csv"),
+        ("data_dir",
+         "Install directory",
+         DATA_DIR),
     ]
     table_names = []
 
@@ -37,7 +40,7 @@ class engine(Engine):
     def create_table(self):
         """Create the table by creating an empty csv file"""
         self.auto_column_number = 1
-        self.file = open_fw(self.table_name())
+        self.file = open_fw(os.path.join(self.opts["data_dir"], self.table_name()))
         self.output_file = open_csvw(self.file)
         column_list = self.table.get_insert_columns(join=False, create=True)
         self.output_file.writerow([u'{}'.format(val) for val in column_list])
@@ -97,7 +100,9 @@ class engine(Engine):
     def table_exists(self, dbname, tablename):
         """Check to see if the data file currently exists"""
         tablename = self.table_name(name=tablename, dbname=dbname)
-        return os.path.exists(tablename)
+        tabledir = self.opts["data_dir"]
+        table_name = os.path.join(tabledir, tablename)
+        return os.path.exists(table_name)
 
     def to_csv(self, sort=True):
         """Export sorted version of CSV file"""
--- a/retriever/engines/jsonengine.py
+++ b/retriever/engines/jsonengine.py
@@ -30,7 +30,10 @@ class engine(Engine):
     required_opts = [
         ("table_name",
          "Format of table name",
-         os.path.join(DATA_DIR, "{db}_{table}.json")),
+         "{db}_{table}.json"),
+        ("data_dir",
+         "Install directory",
+         DATA_DIR),
     ]
     table_names = []
 
@@ -40,7 +43,7 @@ class engine(Engine):
 
     def create_table(self):
         """Create the table by creating an empty json file"""
-        self.output_file = open_fw(self.table_name())
+        self.output_file = open_fw(os.path.join(self.opts["data_dir"], self.table_name()))
         self.output_file.write("[")
         self.table_names.append((self.output_file, self.table_name()))
         self.auto_column_number = 1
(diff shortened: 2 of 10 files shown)


henrykironde:
Good start @harshitbansal05

zhangcandrew:
Thanks @harshitbansal05! Henry and I reviewed it over here, I included a PR into your branch with some fixings, but things look pretty good. Test them out, see if it feels good. We're trying to enforce it now so that the -t or -f functions ONLY deal with the table name or the file name, but don't have any path information whatsoever. This would save us a lot of headache and make the code a bit easier too. 

The -d flag would be responsible for all information on the path. See how it feels. We can create some test for it afterwards too.

harshitbansal05:
@henrykironde I have made the suggested changes, but I could not understand what you mean by adding an option value to the tests. Could you please explain it in a little more detail?

henrykironde:
@harshitbansal05, I could see why we are failing, we are taking table name as only a file name, but in the engines we are using the file name as a pull path.

```python
       ("table_name",
         "Format of table name",
         os.path.join(DATA_DIR, "{db}_{table}.csv")),
```
I think it should be 
```python
       ("table_name",
         "Format of table name",
         "{db}_{table}.csv"),
```
and In the sqlite i[nterface](https://github.com/weecology/retriever/pull/1245/files#diff-6eeaff0b60ef0bad51fbc6a033928c46R94) we are still using  file_dir. 
The same applies to fetch function.
However, we may still have SQlite fails on regression after these changes and we shall have to change things again in the tests. The tests will be a separate commit after we have figured out what exactly is happening

harshitbansal05:
@henrykironde I do not understand, what you mean by the problem caused by the `table_name`. As far as I understand, the error is only because of renaming the argument `file_dir` to `data_dir` for the sqlite and msaccess engines without the corresponding changes in the test files. Also, the travis error logs do not throw any error for csv tests.
Also, if I do not make the changes in the tests, the build would fail. So, is it fine to push with the builds failing?

harshitbansal05:
Also, presently, I am not changing the format of the argument `table_name` for csv, etc. It still is of the form `os.path.join(DATA_DIR, "{db}_{table}.csv"))`. So, the task of changing the table_name argument to the form ` "{db}_{table}.csv"` should also be done in this pr?

henrykironde:
> @henrykironde I have made the suggested changes, but I could not understand what you mean by adding an option value to the tests. Could you please explain it in a little more detail?

The first design did not need a change on tests since we are adding optional args to the function signature, which was only used when set, `install_example(dataset, file ="data.file_name", data_dir = default_DATA_DIR) `.  data_dir was defaulting to "." in the function, adding it would have the same behavior for test. However a new test would be required to show that if  data_dir was set to a different location `data_dir = "new_loc"`, we would still get the data needed.

In the new design, it seems like a lot is changing . We assume that table name, or file name should only be a name not a path. And data_dir should default to "." or DATA_DIR.

henrykironde:
> So, the task of changing the table_name argument to the form ` "{db}_{table}.csv"` should also be done in this pr?
I think we should, we forgot to add that. https://github.com/harshitbansal05/retriever/pull/2

harshitbansal05:
@zhangcandrew I have made the appropriate changes, however, am unable to understand one thing. The tests would fail with the corresponding changes. So should I push the commit with the tests failing or there is some other workaround to it?

zhangcandrew:
Hey @harshitbansal05! If you merge harshitbansal05#2 in, I think the checks should pass.

harshitbansal05:
@henrykironde @zhangcandrew could you please review it now?

zhangcandrew:
Cool. Just wondering, why did you add `def table_name()` to the xml, csv, and json engines? I think it should be able to work ok w/o those functions

harshitbansal05:
@zhangcandrew, the original implementation of the function `table_name` returned the entire name of the table along with the directory for xml, csv and json engines. So, I had overridden the method to make it consistent with the previous implementation.

henrykironde:
Thanks folks, good work @harshitbansal05,  I think the goal of the changes was to use the 
`data_dir` as the path and then the file names should only be the base name. After going through this with @zhangcandrew, we came up with that design and sent you the PR. This new design eliminates the return of table_name as a full path. Which you have correctly implemented but added 
`def table_name() `making it wrong. 

If you think that the design has something wrong, please let us know. :) we shall be more than happy to rethink of the design. Right now I think the code would be fine with out 
`def table_name()`

harshitbansal05:
@henrykironde, I agree. There were some dependencies on the function `table_name` which required the entire path of the table including the directory. Two of them were `table_exists` in the json and csv, and `open_fw` in json, csv and xml files(although the former has no usage). I have now made changes to those dependencies and removed the overridden methods. Hope this would work.

harshitbansal05:
@henrykironde do any more changes have to be done or should I squash the commits?

harshitbansal05:
@henrykironde @zhangcandrew could you please review it now?

henrykironde:
A big 👍 for getting this everyone who contributed. 
Thanks @ashishpriyadarshiCIC for getting this to the ready stage.

PR #1248, opened 2019-02-20
Create absent directories when installing to flat files

### LINKED ISSUE #1136: Create absent directories when installing to flat files
The flat file engines will currently error if asked to install into a directory that doesn't exist, e.g.,

```
$ retriever install csv mammal-masses -t testdata/{db}_{table}.csv
=> Installing mammal-masses
[Errno 2] No such file or directory: 'testdata/mammal_masses_MammalMasses.csv'
Done!
```

We should create these directories on the fly. This may require some refactoring to split of the path into directories and file names, but I haven't looked yet.

ashishpriyadarshiCIC:
This PR fixes #1136 . It will create directory on the fly if it does not exists.

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -196,6 +196,8 @@ def main():
             for dataset in scripts:
                 print("=> Installing", dataset.name)
                 try:
+                    if not os.path.exists(args.data_dir):
+                        os.makedirs(args.data_dir)
                     dataset.download(engine, debug=debug)
                     dataset.engine.final_cleanup()
                 except KeyboardInterrupt:


retriever/__main__.py
```diff
@@ -196,6 +196,8 @@ def main():
             for dataset in scripts:
                 print("=> Installing", dataset.name)
                 try:
+                    if not os.path.exists(args.data_dir):
```
henrykironde: So we looked at the logic and this works fine. we added it to the PR https://github.com/harshitbansal05/retriever/pull/1/files#diff-ff05cd5f49d751e66a61636b498c70acR730, So incase you @ashishpriyadarshiCIC and @harshitbansal05 collaborate, we should have one PR that is clean and has, may be two commits or one commit with two people.

Look up how to make a commit with two contributors:
One author commits  but adds the author name and email of the other. using --author.

ashishpriyadarshiCIC: Ok Sir

henrykironde:
Fixed in 796847d

Thanks @ashishpriyadarshiCIC for the contribution.

Issue #1240, opened 2019-02-12
edit_json cannot dump to file correctly


zhangcandrew:
Looks like the arguments json.dumps takes in have changed within the last year or so. It only takes in 1 positional argument now, whereas before it could take two. Looks like we might have to use json.dump instead. 

Currently when doing an edit_json, I get this error on Ubuntu 18/python 3.7. 

![image](https://user-images.githubusercontent.com/15080652/52660542-de0bcf00-2ecd-11e9-9eb6-e8cec10bd4c5.png)

Looks like it's caused by lines 429/430 in datapackage.py.

ashishpriyadarshiCIC:
Can I work on this?

zhangcandrew:
Sure thing @ashishpriyadarshiCIC!

ashishpriyadarshiCIC:
@zhangcandrew How should I test if the issue is fixed?

zhangcandrew:
The best way to test it is probably just to run through the edit_json. 

type in `retriever edit_json (script_name)` where (script_name) is any script we currently have. For example:

'retirever edit_json iris'

Try to make some changes using the edit_json interface and then when it gets to the end, it should dump out a json instead of erring out like above.

zhangcandrew:
fixed in #1249

PR #1249, opened 2019-02-23
edit_json cannot dump to file correctly

ashishpriyadarshiCIC:
json.dumps takes only 1 argument since Python 3.6 so it's replaced by json.dump
Fixes issue #1240


zhangcandrew:
This looks good to me. Thanks @ashishpriyadarshiCIC!

PR #1251, opened 2019-02-28
specify the encoding type

YuliZheng:
specify the encoding type to avoid UnicodeDecodeError when using retriever install.

default

When using retriever on Windows or other platform, default encoding type may not work on open (see pic above). Specify the encoding type to "utf8" makes retriever usable in many other platform.

### DIFF
--- a/retriever/lib/datapackage.py
+++ b/retriever/lib/datapackage.py
@@ -247,8 +247,8 @@ def create_json():
     file_name = contents['name'] + ".json"
     file_name = file_name.replace('-', '_')
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
-        json_str = json.dumps(contents, output_file, sort_keys=True, indent=4,
-                              separators=(',', ': '))
+        json_str = str(json.dump(contents, output_file, sort_keys=True, indent=4,
+                              separators=(',', ': ')))
         output_file.write(json_str + '\n')
         print("\nScript written to " + file_name)
         output_file.close()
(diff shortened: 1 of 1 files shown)


retriever/lib/scripts.py
```diff
@@ -60,7 +60,7 @@ def reload_scripts():
         files = [file for file in os.listdir(search_path)
                  if file[-3:] == ".py" and file[0] != "_" and
                  ('#retriever' in
-                  ' '.join(open(join(search_path, file), 'r').readlines()[:2]).lower())
+                  ' '.join(open(join(search_path, file), 'r', encoding='UTF8').readlines()[:2]).lower())
```
henrykironde: Could you try 
```Python
' '.join(open_fr(join(search_path, file), encode=ENCODING).readlines()[:2]).lower())
```

retriever/lib/scripts.py
```diff
@@ -60,7 +60,7 @@ def reload_scripts():
         files = [file for file in os.listdir(search_path)
                  if file[-3:] == ".py" and file[0] != "_" and
                  ('#retriever' in
-                  ' '.join(open(join(search_path, file), 'r', encoding='UTF8').readlines()[:2]).lower())
+                  ' '.join(open(join(search_path, file), 'r', encoding=ENCODING).readlines()[:2]).lower())
```
henrykironde:
```Python
' '.join(open_fr(join(search_path, file), encode=ENCODING).readlines()[:2]).lower())

```

YuliZheng: I found the problem when running
default

I think you mean `open_fr(join(search_path, file), encoding=ENCODING) `because the encode should be either True or False in open_fr.
And `' '.join(open_fr(join(search_path, file), encoding=ENCODING).readlines()[:2]).lower())` can give the correct result.
default


Actually, I think `' '.join(open_fr(join(search_path, file)).readlines()[:2]).lower())` is just fine because the default value for encoding in open_fr is ENCODING.

YuliZheng:
Also, I think I messed up the commits a little bit 😢  Should I start another pull request? Or I can reset and squash them.

henrykironde:
That's fine we can squash them after testing the PR. Will let you know after testing.

YuliZheng:
Sorry I did not see that. created another PR already 😢

PR #1254, opened 2019-02-28
Setting up automating infrastructure for the creation of some clean scripts

zhangcandrew:
Automates script creation through CLI. 

Example test directory from which scripts are created:
image

image

the command is `retriever autocreate` followed by the path to directory to be turned into scripts. Adding a -d flag will make it so that we only create scripts for directories within that directory and all single files within a directory are turned into tables under a script with the name of the directory. 

In the example above, nested.json is a script containing two tables: iris and sby.

Without the -d flag, all single data files are turned into individual scripts. 

Works by creating a directory and putting data files into it, then calling this CLI on that directory. Feel free to ask questions, or suggest improvements!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@
 [![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.1038272.svg)](https://doi.org/10.5281/zenodo.1038272)
 [![JOSS Publication](http://joss.theoj.org/papers/10.21105/joss.00451/status.svg)](https://doi.org/10.21105/joss.00451)
 [![Anaconda-Server Badge](https://anaconda.org/conda-forge/retriever/badges/version.svg)](https://anaconda.org/conda-forge/retriever)
+[![Version](https://img.shields.io/pypi/v/retriever.svg)](https://pypi.python.org/pypi/retriever)
+
 
 Finding data is one thing. Getting it ready for analysis is another. Acquiring,
 cleaning, standardizing and importing publicly available data is time consuming
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -17,6 +17,7 @@ from retriever.lib.engine_tools import name_matches, reset_retriever
 from retriever.lib.get_opts import parser
 from retriever.lib.repository import check_for_updates
 from retriever.lib.scripts import SCRIPT_LIST, reload_scripts, get_script
+from retriever.lib.create_scripts import create_package
 
 
 def main():
@@ -103,6 +104,13 @@ def main():
             # edit existing JSON script
             json_file = get_script_filename(args.dataset.lower())
             edit_json(json_file)
+            return
+
+        elif args.command == 'autocreate':
+            if sum([args.f, args.d]) == 1:
+                create_package(args.path, args.dt, args.f, args.d, args.o, args.skip_lines)
+            else:
+                print('Please use one and only one of the flags -f -d')
             return
 
         elif args.command == 'delete_json':
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -108,7 +108,8 @@ def main():
 
         elif args.command == 'autocreate':
             if sum([args.f, args.d]) == 1:
-                create_package(args.path, args.dt, args.f, args.d, args.o, args.skip_lines)
+                file_flag = True if args.f else False
+                create_package(args.path, args.dt, file_flag, args.o, args.skip_lines)
             else:
                 print('Please use one and only one of the flags -f -d')
             return
--- a/retriever/lib/__init__.py
+++ b/retriever/lib/__init__.py
@@ -1,5 +1,6 @@
 """retriever.lib contains the core Data Retriever modules."""
 
+from .create_scripts import create_package
 from .datasets import datasets
 from .datasets import dataset_names
 from .download import download
@@ -17,6 +18,7 @@ from .scripts import reload_scripts
 
 __all__ = [
     'check_for_updates',
+    'create_package',
     'datasets',
     'dataset_names',
     'download',
--- a/retriever/lib/create_scripts.py
+++ b/retriever/lib/create_scripts.py
@@ -1,0 +1,236 @@
+#! /usr/bin/env python
+import collections
+import os
+import json
+import io
+try:
+    from osgeo import ogr
+except ImportError:
+    pass
+from retriever.lib.engine import Engine
+from retriever.lib.models import Table
+from retriever.lib.datapackage import clean_input
+from retriever.lib.tools import open_fw
+
+def get_directory(path):
+    return_path = path
+    if not os.path.isdir(return_path):
+        return_path = os.path.dirname(path)
+    return return_path
+
+def create_package(path, datatype, fileFlag, dirFlag, write_out_flag_path, skip_lines):
+    path = os.path.expanduser(os.path.normpath(path))
+    if not os.path.exists(path):
+        print('Please enter a valid path.')
+
+    if write_out_flag_path is not None and write_out_flag_path == '':
+        write_out_flag_path = os.path.join(get_directory(path), 'scripts')
+    skip_lines = skip_lines[0] if skip_lines else 1
+
+    if datatype.lower() == 'vector':
+        create_vector_datapackage(path)
+    elif datatype.lower() == 'raster':
+        pass
+    elif dirFlag:
+        process_dirs(path, write_out_flag_path, skip_lines)
+    elif fileFlag:
+        process_singles(path, write_out_flag_path, skip_lines)
+
+
+def create_resources(file, skip_lines):
+    engine = Engine()
+    table = engine.auto_create_table(
+        Table(str(file), header_rows=skip_lines), filename=file, make=False)
+    cleanTable = table.__dict__
+    resourceDict = {}
+    pathToTable = os.path.basename(cleanTable['name'])
+    resourceDict['name'] = os.path.splitext(pathToTable)[0]
+    resourceDict['schema'] = {}
+    resourceDict['dialect'] = {}
+    resourceDict['schema']['fields'] = []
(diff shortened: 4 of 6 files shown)


henrykironde:
How about we get this script to run like:
retriever [`functionname`] [`datatype`] [`path`] and the datatype can either tabular, raster or vector. And we could include either it is compressed and requires extraction.

sdtaylor:
here are some help descriptions for autocreate I feel are more descriptive. Changes are in **

```
usage: retriever autocreate [-h] [-t {raster,vector,tabular}] [-d] path

positional arguments:
  path                  **path to the directory containing the data files**

optional arguments:
  -h, --help            show this help message and exit
  -t {raster,vector,tabular}
                        **datatype for files**
  -d                   **create a single script for all data files within the specified directory**
```

ethanwhite:
Feedback from testing in lab meeting this afternoon:

1. If a `scripts` directory doesn't exist in the working directory then `autocreate` fails. We should either fall back on creating the scripts in the working directory or create a scripts directory.
2. We had two OSX users get an error because OSX created a `.DS_store` file in the data directory. We should add a check and ignore for `.DS_store` files or try/catch to not fail on files that error and report the list of files that didn't get included.
3. The behavior of the `-d` functionality was counter intuitive. The API should probably be changed take the path to the directory with the files, not a path to a directory that includes a subdirectory with the files.
4. One dataset that someone tried to add had extra metadata lines at the start of the csv files. We should consider adding a `--skip-lines` argument that skips a set number of lines before autodetecting and adds the correct indicator to skip the lines to the resulting JSON script.

zhangcandrew:
The reasoning behind the -d flag was to allow for multiple directories in a single directory to be processed and turned into scripts. That way we could generate a large batch of directories where every directory is a script. I can definitely make it so that if it points to a directory with no subdirectories, it just makes that single directory a script though.

ethanwhite:
Thanks @zhangcandrew - I'd wondered if that's the case and I like the general idea. The confusion was around how to get a single directory of files to produce a single JSON script (instead of one script per file). What @sdtaylor found (if I'm recalling correctly) was that running on a directory with multiple files resulted in multiple JSON scripts, when the expected behavior was a single JSON script. The only way to get one script for a directory was to nest it inside another directory and then call `autocreate` on the top-level directory using -d.

So, I think the desire is for either a more intuitive way to get a single directory to yield a single script or clearer documentation of the current behavior.

zhangcandrew:
Ok cool, I added more stuff to be descriptive

image

I put in some mutually exclusive required flags that I'm not a huge fan of, but I think it's the easiest way to be the most descriptive and clear with what each is doing here. If there's a better way of doing this feel free to suggest it. 

There's also an -o flag to direct outputted files to a directory now. 

I also wanted to talk about the skip lines flag, I was thinking of restricting it to only be used with single files, or single directories. I'd imagine you probably wouldn't want to skip the same amount of lines (larger than 1) across an entire directory of unrelated files or directory tree, but I don't know how it is actually.

ethanwhite:
> I put in some mutually exclusive required flags that I'm not a huge fan of, but I think it's the easiest way to be the most descriptive and clear with what each is doing here. If there's a better way of doing this feel free to suggest it.

This is definitely better, though I agree it doesn't seem super smooth. One option would be to just restrict the number of options. E.g., we could just keep `-f` and `-d` and if there are subdirectories within the directory passed to `-d` turn each subdirectory into a separate script. I think this would cover >90% of use cases.

> I also wanted to talk about the skip lines flag, I was thinking of restricting it to only be used with single files, or single directories. I'd imagine you probably wouldn't want to skip the same amount of lines (larger than 1) across an entire directory of unrelated files or directory tree, but I don't know how it is actually.

I agree that this only makes sense for a single data source and could be either file or directory specific. So in line with the simplification proposed above I'd suggest just a single `--skip-lines` flag that treats all data files the same way. So `-f` would just apply to the single file and `-d` would apply to all files. If you wanted a bunch of subdirectories you'd have to not use this flag assuming the directories differed.

zhangcandrew:
Sounds good! Everything's been updated and should be working now

image

zhangcandrew:
Sweet!

I actually agree with having a python interface. Aside from consistency across the different interfaces, I think it could also be useful for anyone who wants to utilize autocreating in code. I've added the changes to expose the python interface. 

 I can get PR's in for the other two points. I think both are valid.

ethanwhite:
I can't seem to see the Python interface in the current PR. Probably just looking in the wrong spot when testing. Can you add an example call here?

zhangcandrew:
Sure! I think the change is only in __init__.py, but you should be able to import and use the new function in python like this: 
image

henrykironde:
```python
#! /usr/bin/env python
import collections
import os
import json
import io
try:
    from osgeo import ogr
except ImportError:
    pass
from retriever.lib.engine import Engine
from retriever.lib.models import Table
from retriever.lib.datapackage import clean_input
from retriever.lib.tools import open_fw
from retriever.lib.defaults import VERSION


def create_script_dict(allpacks, path, file_n, skip_lines):
    allpacks["name"] = "FILL"
    allpacks["title"] = "FILL"
    allpacks["description"] = "FILL"
    allpacks["citation"] = "FILL"
    allpacks["licenses"] = [{"name": "FILL"}]
    allpacks["keywords"] = []
    allpacks["homepage"] = "FILL"
    allpacks["version"] = "1.0.0"
    # take this out
    try:
        resources = create_resources(
            os.path.join(path, file_n), skip_lines)
    except:
        print('Package not created, {0} file has an error:'.format(file_n))
        return
    allpacks.setdefault('resources', []).append(resources)
    # allpacks["resources"] = resources
    allpacks["retriever"] = "True",
    allpacks["retriever_minimum_version"] = VERSION
    return allpacks


def create_resources(file, skip_lines):
    engine = Engine()
    table = engine.auto_create_table(
        Table(str(file), header_rows=skip_lines), filename=file, make=False)
    clean_table = table.__dict__
    resource_dict = {}
    path_to_table = os.path.basename(clean_table['name'])
    resource_dict['name'] = os.path.splitext(path_to_table)[0]
    resource_dict['schema'] = {}
    resource_dict['dialect'] = {}
    resource_dict['schema']['fields'] = []
    for cname, ctuple in clean_table['columns']:
        resource_dict['schema']['fields'].append(
            {'name': cname, 'type': ctuple[0]})
    resource_dict['url'] = "FILL"
    return resource_dict


def get_directory(path):
    path = os.path.expanduser(os.path.normpath(path))
    if not os.path.isdir(path):
        path = os.path.dirname(path)
    return path


def create_tabular_datapackage(path, out_path, skip_lines, source_flag):
    if source_flag == "dir_flag":
        process_dirs(path, out_path, skip_lines)
    elif source_flag == "file_flag":
        process_singles(path, out_path, skip_lines)


def create_package(path, data_type, source_flag, out_path, skip_lines):
    """# def create_package(path, data_type, file_flag, dir_flag,
    out_path, skip_lines):

    old add new doc string"""
    path = os.path.expanduser(os.path.normpath(path))
    if not os.path.exists(path):
        print('Please enter a valid path.')
        return
    if not out_path and out_path == '':
        out_path = os.path.join(get_directory(path), 'scripts')
    skip_lines = skip_lines[0] if skip_lines else 1

    if data_type.lower() == 'vector':
        create_vector_datapackage(path, data_type)
    elif data_type.lower() == 'raster':
        # create_raster_datapackage(path, out_path, data_type)
        pass
    elif data_type.lower() == 'tabular':
        create_tabular_datapackage(path, out_path, skip_lines, source_flag)


def write_out_scripts(scriptDict, path, write_out_flag_path):
    file_name = os.path.basename(path).split('.')[0] + '.json'
    path_dir = get_directory(os.path.expanduser(path))
    if write_out_flag_path is not None:
        path_dir = os.path.expanduser(write_out_flag_path)
        if not os.path.exists(path_dir):
            os.mkdir(path_dir)
    write_path = os.path.join(path_dir, file_name)

    if not (scriptDict and 'resources' in scriptDict):
        print(write_path + ' creation skipped because resources were empty.')
        return
    if os.path.exists(write_path):
        choice = clean_input(write_path + ' already exists. Overwrite the script? [y/n]')
        if choice == 'n':
            print(write_path + ' creation skipped.')
            return
    try:
        with open_fw(write_path) as outputPath:
            sortedDict = collections.OrderedDict(sorted(scriptDict.items()))
            jsonstr = json.dumps(sortedDict, sort_keys=True, indent=4)
            outputPath.write(jsonstr)
            print('Successfully wrote scripts to ' + os.path.abspath(write_path))
            outputPath.close()
    except Exception as e:
        print(write_path + ' could not be created. {}'.format(e.message))


def process_dirs(sub_dirs_path, write_out_flag_path, skip_lines):
    for path, subdirs, files in os.walk(sub_dirs_path):
        allpacks = collections.OrderedDict()
        for file_n in files:
            if file_n:
                try_create_dict =  create_script_dict(allpacks, path, file_n, skip_lines)
                if try_create_dict:
                    allpacks = try_create_dict
        write_out_scripts(allpacks, path, write_out_flag_path)


def process_singles(single_files_path, write_out_flag_path, skip_lines):
    if os.path.isdir(single_files_path):
        for path, subdirs, files in os.walk(single_files_path):
            for file_n in files:
                allpacks = collections.OrderedDict()
                if file_n:
                    allpacks = create_script_dict(allpacks, path, file_n, skip_lines)
                    filepath = os.path.join(path, file_n)
                    write_out_scripts(allpacks, filepath, write_out_flag_path)
    else:
        directory = os.path.dirname(single_files_path)
        file_name = os.path.basename(single_files_path)
        allpacks = collections.OrderedDict()
        allpacks = create_script_dict(allpacks, directory, file_name, skip_lines)
        write_out_scripts(allpacks, single_files_path, write_out_flag_path)

```

henrykironde:
Thanks @zhangcandrew for working on this and @ethanwhite for the input. this LGTM

Issue #1247, opened 2019-02-20
Improve output when connection is not made


henrykironde:
In both cases (Postgres and MySql), I have no databases up.
A run will finish with a **Done!**. I think we should catch the error and print something that tells the user of no installation made.

`retriever install postgres poker-hands`
```shell
=> Installing poker-hands
Creating database poker_hands...

could not connect to server: Connection refused (0x0000274D/10061)
        Is the server running on host "localhost" (::1) and accepting
        TCP/IP connections on port 5432?
could not connect to server: Connection refused (0x0000274D/10061)
        Is the server running on host "localhost" (127.0.0.1) and accepting
        TCP/IP connections on port 5432?

Installing into existing database
Downloading poker-hand-testing.data: 11.7kB [00:03, 3.52kB/s]
could not connect to server: Connection refused (0x0000274D/10061)
        Is the server running on host "localhost" (::1) and accepting
        TCP/IP connections on port 5432?
could not connect to server: Connection refused (0x0000274D/10061)
        Is the server running on host "localhost" (127.0.0.1) and accepting
        TCP/IP connections on port 5432?

Replacing existing table
could not connect to server: Connection refused (0x0000274D/10061)
        Is the server running on host "localhost" (::1) and accepting
        TCP/IP connections on port 5432?
could not connect to server: Connection refused (0x0000274D/10061)
        Is the server running on host "localhost" (127.0.0.1) and accepting
        TCP/IP connections on port 5432?

Done!
```
`C:\HenryGitHub\retriever [master +1 ~0 -0 !]> retriever install mysql  poker-hands`

```shell
=> Installing poker-hands
Creating database poker_hands...

(1045, "Access denied for user 'root'@'localhost' (using password: NO)")
Installing into existing database
(1045, "Access denied for user 'root'@'localhost' (using password: NO)")
Replacing existing table
(1045, "Access denied for user 'root'@'localhost' (using password: NO)")
Done!
```

ashishpriyadarshiCIC:
@henrykironde So it means to remove the message `Done!` with a descriptive message in case installations fails? If that is the case can I work on it?

henrykironde:
It should be something like connection failed. We  seem like we are not catching the connection failure error.
and we should not see that Done, coz we shall return on a failed connection.

ashishpriyadarshiCIC:
So it should run a loop until it connects or there is a keyboard interrupt.

henrykironde:
Not sure if that is what we should do. 
[...]

PR #1255, opened 2019-03-01
Improve output when connection is not made

ashishpriyadarshiCIC:
Solves issue #1247


henrykironde:
How did you test this ?

ashishpriyadarshiCIC:
The issue of `Done` showing up after an exception was there for every type of exception. So, I recreated some different type of errors like I disabled the internet, made some changes on the source code to create some errors, etc.

henrykironde:
Okey that is a good start. Now try to run the same commands in issue when you have stoped both the Postgres and Mysql server. Let me know what you get.

ashishpriyadarshiCIC:
Here is the output :- 
screenshot from 2019-03-02 11-44-38
screenshot from 2019-03-02 11-45-55

henrykironde:
Could we find away to catch that error?

ashishpriyadarshiCIC:
@henrykironde Are you talking about catching the error on the methods to connect to the server because the message `Done!` would have been there even if the error was caught at the method where the connection is made. I'll try to catch the error at the get connect method and will see if there is a difference.

ashishpriyadarshiCIC:
@henrykironde I've made a small change. Can you please review it?

henrykironde:
When you run the command without databases up, what do you get?

ashishpriyadarshiCIC:
Here it is :- 
screenshot from 2019-03-03 22-19-26

ashishpriyadarshiCIC:
@henrykironde Is this ok?

henrykironde:
I feel like we are handling it at a higher level. 
The Engine() member function `connection` gets to conn after `get_connection(): ` in the other inheriting engines, If I am correct.
From a quick look, I think we need to revisit def get_connection(self): in the Postgres and Mysql engines. Let me know if I am clear

ashishpriyadarshiCIC:
I might have to close this PR and open a new PR as I messed up many things while using git.

PR #1256, opened 2019-03-02
use open_fr instead of open

YuliZheng:
I found the problem when running this on windows
default
I think that is due to the default decoding type in windows is gbk, and I think we should explicitly use ENCODING.
After changing, it can give the correct result.
default

Thanks for your kindly reviewing


henrykironde:
Thanks @YuliZheng for getting this error reported and fixing it

PR #1267, opened 2019-03-03
Return OrderedDict in fetch_tables in retriever/engines/sqlite.py

### LINKED ISSUE #1266: fetch method should return an ordered dict (not a dict)
Hello,

I noticed that `fetch` method returns a dict of dataframes.
To preserve order of tables (which may have a meaning) it should return an ordered dict.
[Datapackage](https://frictionlessdata.io/specs/data-package/) ressources are stored in a list so it's ordered.

Kind regards

Jaffrey98:
Resolves #1266

- Default `dict` doesn't maintain order of key value pairs
  i.e. in fetch_tables function, all rows of the table are getting
  populated into a `dictionary`, but we may need the data in the order
  in which the rows actually exist in the database.
- To facilitate this, `OrderedDict` from `collections` has been used. It
  maintains the order in which the values are assigned to corresponding
  keys.

### DIFF
--- a/retriever/engines/sqlite.py
+++ b/retriever/engines/sqlite.py
@@ -45,10 +45,12 @@ class engine(Engine):
     def fetch_tables(self, dataset, table_names):
         """Return sqlite dataset as list of pandas dataframe."""
         connection = self.get_connection()
-        data = OrderedDict({table[len(dataset) + 1:]: pd.read_sql_query("SELECT * "
-                                                                        "FROM {};".format(table),
-                                                                        connection)
-                            for table in table_names})
+        sql_query = "SELECT * FROM {};"
+        data = OrderedDict({
+                    table[len(dataset) + 1:]
+                    :pd.read_sql_query(sql_query.format(table),connection)
+                    for table in table_names
+               })
         return data
 
     def get_bulk_insert_statement(self):
--- a/retriever/engines/sqlite.py
+++ b/retriever/engines/sqlite.py
@@ -48,7 +48,7 @@ class engine(Engine):
         sql_query = "SELECT * FROM {};"
         data = OrderedDict({
                     table[len(dataset) + 1:]
-                    :pd.read_sql_query(sql_query.format(table),connection)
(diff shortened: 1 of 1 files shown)


henrykironde:
Could we add so refactoring to look more readable 

In my case  I would do something like 
```python
sql_stmt =  "SELECT * ...... "
data = OrderedDict({table[len(dataset) + 1:]: pd.read_sql_query(sql_stmt,
```

Jaffrey98:
> Could we add so refactoring to look more readable
> 
> In my case I would do something like
> 
> ```python
> sql_stmt =  "SELECT * ...... "
> data = OrderedDict({table[len(dataset) + 1:]: pd.read_sql_query(sql_stmt,
> ```

Does this seem readable 😅 ?

```python
    def fetch_tables(self, dataset, table_names):
        """Return sqlite dataset as list of pandas dataframe."""
        connection = self.get_connection()
        sql_stmt = "SELECT * FROM {}"
        data = OrderedDict({
                    table[len(dataset) + 1:]
                    :pd.read_sql_query(sql_stmt.format(table),connection)
                    for table in table_names
               })
        return data
```

henrykironde:
sql_query 👍 . This looks really neat to me.

henrykironde:
@Jaffrey98  thank you for working in this issue.

scls19fr:
A test should always be present when fixing a bug so you will have a chance to notice that

        data = OrderedDict({
                    table[len(dataset) + 1:]
                    :pd.read_sql_query(sql_stmt.format(table),connection)
                    for table in table_names
               })

doesn't preserve order because you are building an `OrderedDict` from a `dict` (which is unordered)

You need a `list` of `tuple` `(key,  value)` to [build an OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict).

So I'd say that this should probably be

        data = OrderedDict([
                    (table[len(dataset) + 1:],
                    pd.read_sql_query(sql_stmt.format(table),connection))
                    for table in table_names
               ])

PR #1268, opened 2019-03-04
Fix flat file installs into non-default directories

### LINKED ISSUE #1263: Python interface errors and needs more tests.
We have currently added the `data_dir` parameter. This has caused some changes that introduced some errors. 

```python
Import retriever as rt
rt.install_json("iris", data_dir="test)
Installing iris_Iris.json: 100%|██████████████████████████████████████████| 151/151 [00:00<00:00, 18683.70rows/s]
[Errno 2] No such file or directory: 'iris_Iris.json'
```

To solve this we need to rewrite the disconnect functions for some of the engines like xml, json,

ethanwhite:
The new `data_dir` option wasn't added to the `table_names` list that is
used for disconnecting the flat file engines. This was leading to "No such file"
errors and files that weren't getting disconnected.

Fixes #1263

### DIFF
--- a/retriever/engines/csvengine.py
+++ b/retriever/engines/csvengine.py
@@ -40,11 +40,12 @@ class engine(Engine):
     def create_table(self):
         """Create the table by creating an empty csv file"""
         self.auto_column_number = 1
-        self.file = open_fw(os.path.join(self.opts["data_dir"], self.table_name()))
+        table_path = os.path.join(self.opts["data_dir"], self.table_name())
+        self.file = open_fw(table_path)
         self.output_file = open_csvw(self.file)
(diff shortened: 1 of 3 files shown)


henrykironde:
Wondering if `msaccess` is not affected. I have just looked at the files. Yet to test

ethanwhite:
`msaccess` is using the default `disconnect`, so as long as it's connecting properly it will get disconnected properly. The connection looks to be using `data_dir` correctly, so it doesn't seem like there should be an issue with `msaccess` here.

henrykironde:
Thanks @ethanwhite  for  getting this fixed

Issue #1271, opened 2019-03-04
fetch method should preserve order


scls19fr:
Following https://github.com/weecology/retriever/issues/1266 (because I can't reopen issue)
because https://github.com/weecology/retriever/pull/1267 doesn't fix it (I'm pretty sure)

See https://github.com/weecology/retriever/pull/1267#issuecomment-469147555

henrykironde:
Thanks @scls19fr, Could we get an example of the order to be preserved

scls19fr:
Sorry but I can't do it now, I'm moving (using my phone)

You just need to create a sample dataset with tables not alphabetically ordered to catch this.

harshitbansal05:
@henrykironde I think that scls19fr is absolutely right. Reference: https://stackoverflow.com/questions/15711755/converting-dict-to-ordereddict. The dictionary passed to the Ordered Dict as an argument is inherently not ordered, so the the `ordered dict` would also not be ordered. It should be solved by passing tuples of key, value pairs. If possible, can I work on it?

henrykironde:
I see what you mean and thank you for the explanation folks, I think what it is doing now is keeping the order of tables as they are given in the script. If we want to produce a new ordering, we can add a feature for that where we do some fetch or fetch ordered. Hope I am right. Feel free to put in a PR.

Jaffrey98:
@harshitbansal05 Yes I agree, it seems it slipped my mind that I was actually doing a `normal dictionary comprehension`  inside `OrderedDict()`. Thank you for the FIX 😄

Jaffrey98:
@harshitbansal05 
It seems it isn't necessary to wrap the `comprehension` in `square brackets` , not that it changes the execution. Just wanted to mention it 😅 

> Both work the same
```python
data = OrderedDict([
            (table[len(dataset) + 1:],
            pd.read_sql_query(sql_query.format(table), connection))
            for table in table_names
        ])
```
```python
data = OrderedDict(
            (table[len(dataset) + 1:],
            pd.read_sql_query(sql_query.format(table), connection))
            for table in table_names
        )
```

henrykironde:
I think what @scls19fr wants is to do a sort on the table here `for table in table_names`

harshitbansal05:
[...]

PR #1273, opened 2019-03-04
Solve minor error in fetch preserve method

harshitbansal05:
This pull request aims to fix: #1271. It solves the minor error in preserving the ordering of table names.

### DIFF
--- a/test/test_regression.py
+++ b/test/test_regression.py
@@ -91,6 +91,24 @@ fetch_tests = [
      }])
 ]
 
+fetch_order_tests = [
+    ('airports',
+     ['airports', 'airport_frequencies', 'runways', 'navaids', 'countries', 'regions']
+     ),
+    ('acton-lake',
+     ['ActonLakeDepth', 'ActonLakeIntegrated', 'StreamDischarge', 'StreamNutrients',
+      'SiteCharacteristics']
+     ),
+    ('forest-plots-michigan',
+     ['all_plots_1935_1948', 'all_plots_1974_1980', 'swamp', 'species_codes',
+      'upland_plots_1989_2007', 'sampling_history']
+     ),
+    ('veg-plots-sdl',
+     ['Plots', 'Plot_corners', 'Species', 'Seedling_counts', 'Count1906',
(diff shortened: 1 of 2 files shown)


henrykironde:
Thanks @harshitbansal05 Could we add a test for this

harshitbansal05:
ok, would do that @henrykironde

scls19fr:
Could you please consider #1274 before merging others PRs (and so this one) because #1274 involves a large number of files and merging others PR could lead to conflicts (which is a pain to solve)

test/test_regression.py
```diff
@@ -91,6 +91,24 @@
      }])
 ]
 
+fetch_order_tests = [
+    ('airports',
+     ['airports', 'airport_frequencies', 'runways', 'navaids', 'countries', 'regions']
+     ),
+    ('acton-lake',
+     ['ActonLakeDepth', 'ActonLakeIntegrated', 'StreamDischarge', 'StreamNutrients',
```
henrykironde: use two datasets  acton-lake and forest-plots-michigan

test/test_regression.py
```diff
@@ -246,3 +264,13 @@ def test_fetch():
                 first_row_data = list(data_frame[table_i].iloc[0])
                 assert expected_data == first_row_data
                 assert expected_column_values == column_values
+
+
+def test_fetch_order():
```
henrykironde: I think we can add more cleaning up.
use` @pytest.mark.parametrize("dataset, expected", fetch_order_tests)`
This should be the same as the for loop. `for dataset, expected in fetch_order_tests:`
Something like 

```python
@pytest.mark.parametrize("dataset, expected", fetch_order_tests)
def test_fetch_order():
    """Test fetch dataframe order"""
        data_frame_dict = fetch(dataset)
        assert data_frame_dict.keys() == expected
```
dict.keys returns list of keys and since we return ordered data_frame_dict should be order preserved.

harshitbansal05:
@henrykironde could you please review it now?

henrykironde:
Thanks to everyone who contributed to the issue and PR. 👍

PR #1269, opened 2019-03-04
PEP8

scls19fr:
Check PEP8 using Travis

Closes #1270

PR #1274, opened 2019-03-04
[WIP]: PEP8

### LINKED ISSUE #1270: PEP8
Hello,

following https://github.com/weecology/retriever/issues/279
Travis can check if the Python code in this repository conforms to PEP8.

See https://github.com/weecology/retriever/pull/1269

Some cleanup is probably required.
See https://travis-ci.org/weecology/retriever/jobs/501333622#L934

Kind regards

scls19fr:
Replaces #1269
Closes #1270


scls19fr:
Fixed a lot of PEP8 issues using [Black](https://black.readthedocs.io/)

```bash
$ black --version
black, version 18.9b0
$ black .
```

and a lot of others PEP8 issues manually

scls19fr:
Unfortunately `test_auto_get_datatypes` was failing.
I changed test expected values from `[101, 102, 104]` to `[103, 102, 106]`.
This is just a temporary workaround to check if all others changes are not breaking something else.

Now it seems that tests are passing with Python 3.5... but not with Python 2.7.
https://travis-ci.org/weecology/retriever/jobs/501670103#L862

Any idea what's going wrong?

henrykironde:
@scls19fr this really good. We could improve the PR by setting up rules that everyone / PR will follow.

I think a Flask8 and Black config settings' file would be nice.
Or include Flask8 and black setting in `setup.cfg`. 

We shall then include some logic to keep these rules.
and we could create a pre-commit hook to run the the tools before someone makes a commit 
 
I think we should first run black with `black . --skip-string-normalization `.  And once these changes are reviewed and tested, we could then create a PR without --skip-string-normalization

Otherwise this looks ready for testing.
I also have a feeling that the change of double quotes caused the error, I am not sure, but I will look into it and add a note

👍

henrykironde:
I went through the PR, and OMG somethings are really nice. 
But in some parts, like some lists, are handled in a way that I personally would not recommend. 

For example (https://github.com/weecology/retriever/pull/1274/files?utf8=%E2%9C%93&diff=split#diff-f9d401369dcefb545d08eb98d9b19d6aL58)

We can create a rule for black to change the list more horizontally in some cases.

I possible it would be nice to have separate PRs.  One for only the retriever folder, and another for the scripts' folder and the last one for the the test folder.  This is not necessary but would make testing code changes easy.

**Error: The tests failing** 

> Unfortunately test_auto_get_datatypes was failing.
> I changed test expected values from [101, 102, 104] to [103, 102, 106].
This is because, the encoding of the file has changed and this changed the bits in the  unicode characters. ```[["ö", 'bb', 'Löve']],``` to ``` None, [["�", "bb", "L�ve"]]```

**The scripts folder:**

The python scripts have a very distinct first line, `#retriever` in new version of retriever, the scripts can have a space in that line. In older versions that will not work and will break backward compatibility of the code.
If we make automation tests and auto code refactor using flask/black, that is something that we should consider. `#retriever`

scls19fr:
Thanks @henrykironde for the idea of processing by "chunk" of code (test / script / retriever folders)

I decided to tackle first the `test` folder

I haven't use black at all in this branch
https://github.com/scls19fr/retriever/tree/pep8_tests_no_black

after trying to fix only E402 module level import not at top of file I encountered the `test_auto_get_datatypes` error.

See https://travis-ci.com/scls19fr/retriever/jobs/182340623#L865

Aren't you using unicode utf8?

scls19fr:
the first line

    #retriever

raises this PEP8 error

    E265 block comment should start with '# '

Why not following PEP8 and have a first line like

    # retriever

Adding a pre-commit hook could be a nice idea... but there is still some work until reaching this!

> In older versions that will not work and will break backward compatibility of the code.

older versions? of Python or of retriever?

not sure I have a perfect view of the whole landscape.

henrykironde:
> older versions? of Python or of retriever?
 the older versions of the retriever.
Older version still use the scripts in the master. So that is why we keep then like that.

henrykironde:
The change in encoding happens when you use either an editor or terminal which does not have a good default encoding set that can support global localization but works well on any local requirements. Like being able to process Hindi or Japanese character sets but works well with any European characters.

scls19fr:
Closing this PR because closing #1276 and #1277 should probably be considered before #1270

henrykironde:
I have this idea, may be we could get this to a clean branch and we label that branch something like python3. So that we do all the work related to the shift to python3 on that branch. What do you think?
I am reopening this PR and adding a `[WIP]`. We do what to lose truck of this work.

henrykironde:
https://travis-ci.org/henrykironde/retriever/builds/503395616 
You could force a commit from https://github.com/henrykironde/retriever/commit/92be9e00019005bdf868c3e7349d389ce254f030.

henrykironde:
Ref to PR 1280

PR #1275, opened 2019-03-05
Added file logger

### LINKED ISSUE #151: Log errors and other issues

piyushchauhan:
Fixes #151 
Made file logger using `logging` module and used it for as many exceptions as possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
 # database setup
 before_script:
   - sqlite3 --version
-  - psql -c 'create database testdb;' -U postgres
+  - psql -c 'create database testdb_retriever;' -U postgres
 
 after_success:
   - codecov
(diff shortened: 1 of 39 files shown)


henrykironde:
Could you update your branch with master.
I think we should limit loggers to only areas where we expect some catastrophic failures. I am against all if statements having a logger.

henrykironde:
Lets not have loggers in the scripts directory and test directory

henrykironde:
Something worth reading through would be the last PR comments, In the Issue page there are two PR's that tried to solve this problem

harshitbansal05:
@piyushchauhan if possible can I contribute to it?

henrykironde:
You can contribute to it.

henrykironde:
@piyushchauhan I will close for now since we had to reorganize some the design when it comes to logging. Will open / reference this PR in the future

PR #1280, opened 2019-03-11
Apply pep8 

henrykironde:
Applies pep8 to the code as
Ref #1274, #1269


henrykironde:
> Fixed a lot of PEP8 issues using Black
> $ black --version
> black, version 18.9b0
> $ black .
> and a lot of others PEP8 issues manually

henrykironde:
Something to note:
Long lists need to be compressed so as they can appear more clear.
We need to change the pep8 setting for lists. https://github.com/weecology/retriever/pull/1280/files#diff-3ecb48437eee85df9eab30a681012e11R153

In the setting, we should also add `Maximum Line Length`.  I think a length of 80 could be okay, but suggestions otherwise are well much considered and welcomed

zhangcandrew:
Cool. I'm getting some other errors that black might have missed so I put in a PR here on your branch (https://github.com/henrykironde/retriever/pull/119). The other pep8 errors showing are module import and Line too long. The module imports are due to install_aliases() from the future library so I think that's ok. I'm fine with Line too long being set to 80, but we are getting quite a few of those errors.

henrykironde:
looks good, f0da960

PR #1281, opened 2019-03-11
Improve output when connection is not made

ashishpriyadarshiCIC:
Solves issue #1247

### DIFF
--- a/retriever/engines/postgres.py
+++ b/retriever/engines/postgres.py
@@ -249,11 +249,16 @@ CSV HEADER;"""
         import psycopg2 as dbapi
 
         self.get_input()
-        conn = dbapi.connect(host=self.opts["host"],
-                             port=int(self.opts["port"]),
-                             user=self.opts["user"],
-                             password=self.opts["password"],
-                             database=self.opts["database"])
+        try:
+            conn = dbapi.connect(host=self.opts["host"],
+                                 port=int(self.opts["port"]),
+                                 user=self.opts["user"],
+                                 password=self.opts["password"],
+                                 database=self.opts["database"])
+        except dbapi.OperationalError as _:
+            print("Connection with PostgreSQL Failed! Try to restart PostgreSQl Server")
+        except Exception as e:
+            print(e)
         encoding = ENCODING.lower()
         if self.script.encoding:
             encoding = self.script.encoding.lower()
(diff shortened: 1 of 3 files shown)


ashishpriyadarshiCIC:
@henrykironde Can you please review it

retriever/engines/postgres.py
```diff
@@ -249,11 +249,16 @@ def get_connection(self):
...
-                             user=self.opts["user"],
-                             password=self.opts["password"],
-                             database=self.opts["database"])
+        try:
+            conn = dbapi.connect(host=self.opts["host"],
+                                 port=int(self.opts["port"]),
+                                 user=self.opts["user"],
+                                 password=self.opts["password"],
+                                 database=self.opts["database"])
+        except dbapi.OperationalError as _:
```
henrykironde: Why do we two exception handlers ?

henrykironde:
Did you test this. And could you paste the test results

retriever/engines/postgres.py
ashishpriyadarshiCIC: I thought if there was another error it would not print the same message

henrykironde: I think we can put it on the same line except (dbapi.OperationalError, Exception) as err

ashishpriyadarshiCIC:
@henrykironde Here is the output :- 
Screenshot from 2019-03-12 00-18-23

ethanwhite:
It seems like ideally we should stop as soon as the connection fails rather than continuing with the following installation steps.

henrykironde:
Ethan is correct. If you trace the code from the get connection, you should guide the execution to return to main. So search for the functions that call get connection. and return if the connection is None. 
Try running `retriever install --debug  postgres  iris` to see the trace

ashishpriyadarshiCIC:
I'll start working on it as I was using `raise` initially to stop.

ashishpriyadarshiCIC:
@henrykironde I left this PR for too long, I will complete it now

ashishpriyadarshiCIC:
@henrykironde  Should I create a new PR as this branch is too outdated ?

henrykironde:
Sure go for it

ashishpriyadarshiCIC:
@henrykironde Closed this PR as another PR #1432  regarding the issue had been merged.

PR #1287, opened 2019-03-18
Add option for pretty json

### LINKED ISSUE #1285: JSON file generated using install json is not pretty
The .json file generated when installing datasets into json is not pretty.

For example, consider abalone-age-

The output file generated when using 'retriever install json abalone-age` looks like:
current_look

The size of this file is 726.5 kB.

The output file when setting the indentation of json to 2:
Screenshot from 2019-03-15 00-54-56

The size of this file is 810.1 kB.

It would be great if we have an option of setting the indentation and sorting the keys
 `retriever install json abalone-age --sort-keys --indent 2` to increase the readability of the json file.
I am aware that we keep the json files the way it is right now to reduce the file size but again these new options are optional. When installing the `abalone-age` to `xml` it takes 1.3 MB.
It would be nice if we let the user decide upon the tradeoff between file size and readability.

@ethanwhite @henrykironde Your views on this?
If you like it I will send a PR for this.

apoorvaeternity:
Fixes #1285

### DIFF
--- a/retriever/engines/jsonengine.py
+++ b/retriever/engines/jsonengine.py
@@ -12,7 +12,7 @@ from retriever.lib.engine_tools import json2csv, sort_csv
 
 
 class engine(Engine):
-    """Engine instance for writing data to a CSV file."""
+    """Engine instance for writing data to a JSON file."""
 
     name = "JSON"
     abbreviation = "json"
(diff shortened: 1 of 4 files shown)


ethanwhite:
Great first implementation @apoorvaeternity. This works perfectly from the CLI, but we also need it to work from the Python interface.

apoorvaeternity:
@ethanwhite Ready for a review.

retriever/engines/jsonengine.py
```diff
@@ -110,10 +110,14 @@ def insert_statement(self, values):
         else:
             newrows = values
         json_dumps = []
+        pretty = True if "pretty" in self.opts and self.opts["pretty"] is True else False
         for line_data in newrows:
             tuples = (zip(keys, line_data))
             write_data = OrderedDict(tuples)
-            json_dumps.append(json.dumps(write_data, ensure_ascii=False) + ",")
+            if not pretty:
```
henrykironde: I think we can move the [check](https://github.com/weecology/retriever/pull/1287/files#diff-4b22ee1a8c7630feb86e76d515b9dcf9R113) here instead of` if not pretty` to 
`if "pretty" in self.opts and not self.opts["pretty"]":`

apoorvaeternity: It's inside a loop so it will have to check the condition for every row though it is not needed.
This is why I have written the check before the loop in a variable.

apoorvaeternity: Though the condition is not very complex we can move it down for readability of code. As you say.

henrykironde: Thats fine, both will take constant time.

henrykironde:
@apoorvaeternity, thanks for the PR.

PR #1292, opened 2019-03-26
Use smaller samples of GIS data for testing

henrykironde:
Use sampled data from bioclim 1 and 2
and Ecoregions levels


zhangcandrew:
Looks good. Thanks for the updated tests.

PR #1293, opened 2019-03-26
Add Spatial data testing on Docker

### LINKED ISSUE #1295: Removing tests for python 2
Officially removing tests for Python2 and this is because of the fact that the underlying tools are breaking more often.

**`On Travis:`**
We are moving to dockers specifically for Geo spatial testing
Most Geo-spatial tools will not work with Python 2  straight from the box.
**`On Appveyor:`**
There must have been some updates or changes in the environment and this is causing encoding problems. 
When using Python2 tools/environment, some Unicode character encoding in a PR will result into a diff due to failed proper encoding/decoding of the bits for those chars.

henrykironde:
Use docker ubuntu:16.04 for retriever
Use kartoza/postgis:11.0-2.5 for posgis
	- This has Extensions loaded
Use mysql:5.7 image for MySQL

fixes #1295

### DIFF
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -100,10 +100,6 @@ for engine in engine_list:
         engine_parser.add_argument('-b', '--bbox', nargs=4,
                                    help='Set bounding box xmin, ymin, xmax, ymax',
                                    required=False)
-        if engine.name == "JSON":
-            engine_parser.add_argument('-p', '--pretty',
-                                       help='Add indentation to json file', action='store_true',
-                                       required=False)
 
     abbreviations = set('h')
 
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -100,6 +100,10 @@ for engine in engine_list:
         engine_parser.add_argument('-b', '--bbox', nargs=4,
                                    help='Set bounding box xmin, ymin, xmax, ymax',
                                    required=False)
+        if engine.name == "JSON":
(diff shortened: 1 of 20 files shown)


henrykironde:
Looks like we need a better way to catch errors.
Instead of `docker-compose  run --service-ports python_retriever py.test -v --cov=retriever 2>error.log && if [[ -s "error.log" ]]; then exit 1; else echo 0; fi`

retriever/lib/get_opts.py
```diff
@@ -100,10 +100,6 @@
         engine_parser.add_argument('-b', '--bbox', nargs=4,
                                    help='Set bounding box xmin, ymin, xmax, ymax',
                                    required=False)
-        if engine.name == "JSON":
-            engine_parser.add_argument('-p', '--pretty',
-                                       help='Add indentation to json file', action='store_true',
-                                       required=False)
```
ethanwhite: Is the removal of the pretty JSON stuff intentional?

retriever/lib/install.py
```diff
@@ -125,15 +125,14 @@ def install_msaccess(dataset, file='access.mdb',
 
 def install_json(dataset,
                  table_name='{db}_{table}.json',
-                 data_dir=DATA_DIR, debug=False, use_cache=True, pretty=False):
+                 data_dir=DATA_DIR, debug=False, use_cache=True):
```
ethanwhite: Same question about this removal being intentional?

ethanwhite:
Nice work @henrykironde!

PR #1297, opened 2019-03-28
Add USA eco legions

henrykironde:
Level III and IV Ecoregions of the Continental
United States

### DIFF
--- a/retriever/engines/postgres.py
+++ b/retriever/engines/postgres.py
@@ -45,6 +45,8 @@ class engine(Engine):
                       "{db}.{table}"),
                      ]
     spatial_support = True
+    # default postgres encoding
+    db_encoding = "Latin1"
 
     def auto_create_table(self, table, url=None, filename=None, pk=None):
         """Create a table automatically.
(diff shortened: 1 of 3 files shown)


scripts/us_eco_level_four.json
```diff
@@ -0,0 +1,250 @@
...
+    "Natural Resources", "spatial-data"
+    ],
+    "extent": {
+    "xMax": 2258224.8307000003,
+    "xMin": -2356069.457800001,
+    "yMax": 3172577.0456,
+    "yMin": 272048.49569999985
+    },
+    "version": "1.0.0",
+    "retriever_minimum_version": "2.1.0",
```
ethanwhite: minimum version should be 2.2.0 right?

henrykironde:
I separated them into different scripts because they have different citations for the levels.
But will add them and try to figure out the citation.
@ethanwhite in case you know the correct citation to the combined data help me and paste it here.

ethanwhite:
Ah, that makes sense. I still think it's more useful from the user perspective to package all of these together. I think the correct long-term solution is to allow `citation` to be a list and present multiple citations if there are more than 1. It's not uncommon for datasets to ask for multiple papers to be cited so it's a general thing to improve. For now. I think we could probably just include both of them in a single string back to back.

ethanwhite:
Once you've added the other citation back in let me know and I'll get this merged. I'll also open an issue about multiple citations.

henrykironde:
we have decided to use "table_citations": {} for multi citations.

PR #1305, opened 2019-04-15
 Use sub-dir to provide custom directory

henrykironde:


retriever/engines/download_only.py
```diff
@@ -37,37 +36,24 @@ def get_connection(self):
...
-        name already exist in the directory.
-
-        """
+        """Copies downloaded files to desired directory"""
         if hasattr(self, "all_files"):
             for file_name in self.all_files:
                 file_path, file_name_nopath = os.path.split(file_name)
-                subdir = os.path.split(file_path)[1] if self.opts['subdir'] else ''
+                subdir = self.opts['sub_dir'] if self.opts['sub_dir'] else ""
                 dest_path = os.path.join(self.opts['path'], subdir)
```
ethanwhite: Should this line have also changed `subdir` -> `sub_dir`?

henrykironde: We could do that, I just try to avoid using the same variable names and parameter names in cases like this

ethanwhite: Ah, so we have both `subdir` and `sub_dir` now. I wasn't clear.

henrykironde: should I use sub_dir_value

ethanwhite: After a good nights sleep I think this looks fine. The other option would be to use `get` to avoid the intermediate variable:

```python
dest_path = os.path.join(self.opts['path'], self.opts.get('sub_dir', ""))
```

If you prefer the original version let me know and I'll merge. If you want to make the change just ping me when it's ready.

henrykironde: I will have to change

henrykironde: Thanks @ethanwhite made the changes, thank you for that line .get()

PR #1306, opened 2019-04-15
Solve None error in json dump

harshitbansal05:
Presently, when a script is created or edited via `new_json` or `edit_json`, `None` is appended at the end. It is due to the line `output_file.write(json_str + '\n')`, where `json_str` is None. Currently, `json_str = str(json.dump(contents, output_file, sort_keys=True, indent=4, separators=(',', ': ')))` is called, but `json.dump()` itself writes the json formatted stream to the file. Thus, it returns None (which eventually gets assigned to `json_str`). This pull request aims to fix this. Please correct me if I am wrong.

### DIFF
--- a/retriever/lib/datapackage.py
+++ b/retriever/lib/datapackage.py
@@ -247,9 +247,9 @@ def create_json():
     file_name = contents['name'] + ".json"
     file_name = file_name.replace('-', '_')
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
-        json_str = str(json.dump(contents, output_file, sort_keys=True, indent=4,
-                              separators=(',', ': ')))
-        output_file.write(json_str + '\n')
+        json.dump(contents, output_file, sort_keys=True, indent=4,
+                  separators=(',', ': '))
+        output_file.write('\n')
         print("\nScript written to " + file_name)
         output_file.close()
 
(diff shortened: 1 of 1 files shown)


henrykironde:
We solved something like this recently, Are you using python3?

harshitbansal05:
yes, @henrykironde

henrykironde:
Okey will test it though it looks fine. We shall be done by tomorrow.

retriever/lib/datapackage.py
```diff
@@ -247,9 +247,9 @@ def create_json():
     file_name = contents['name'] + ".json"
     file_name = file_name.replace('-', '_')
     with open(os.path.join(HOME_DIR, 'scripts', file_name), 'w') as output_file:
-        json_str = str(json.dump(contents, output_file, sort_keys=True, indent=4,
-                              separators=(',', ': ')))
-        output_file.write(json_str + '\n')
+        json.dump(contents, output_file, sort_keys=True, indent=4,
+                  separators=(',', ': '))
+        output_file.write('\n')
         print("\nScript written to " + file_name)
```
henrykironde: looks fine. 
Could you change `for key, val in obj.items():` to `for key, val in obj.copy().items():`
https://github.com/weecology/retriever/pull/1306/files#diff-7cb8d9949f19aa110e816b90e44ffb2dR261

harshitbansal05: Done.

henrykironde:
Thanks @harshitbansal05 👍

PR #1310, opened 2019-04-16
Remove urls attribute from sample script

apoorvaeternity:
We no longer use the `urls` attribute in the json scripts so I have removed it from the sample script too.


ethanwhite:
Looks good!

Issue #1303, opened 2019-04-10
Support multiple citations for datasets


ethanwhite:
Some datasets have multiple citations and sometimes we want to combine multiple datasets into a single dataset (see #1297). It would be nice to be able to support this by effectively displaying more than one citation. We'll need to check if there are issues with the Frictionless Data specs, but from a first glance it looks like `citation` isn't actually in their specs (surprisingly) so we can probably do whatever we want.

henrykironde:
We can use a dictionary
```yaml
  "citation": [
    {
      "table1": "citation for table one.",
      "table2": "citation for table two."
    }
  ]
```

or we could just provide a list of the all the citations and leave it to the user to identify which citation belongs to which data

```yaml
  "citation": [
         "citation for table one.",
         "citation for table two.",
  ]
```

henrykironde:
**Note:** This will break backward compatibility, hence when designing we need to put that into account.
A cleaner version, I would say, is introducing a new keyword. In this case, we can use `"table_citations": {}` or `"table_citations": []`

harshitbansal05:
@henrykironde does it require that we add a new field `table_citations` to all the scripts, which would be a list, initially populated with the present citations. Also, make corresponding changes when creating a new json file?

henrykironde:
We do not have to add the field to all scripts, we have to just process it if present. look for places where we process citation and thinking of how to add `table_citations`
https://github.com/weecology/retriever/search?l=Python&p=1&q=citation.

A good example is https://github.com/weecology/retriever/blob/16d4d0cf07d08221f50756e169c9b35deaebf99a/retriever/__main__.py#L65

To be able to test this you will need a script with more than one citation. you can create a dummy one for the testing. I will get one in later tomorrow.

harshitbansal05:
ok, I would try to look into it @henrykironde.

henrykironde:
Adding a different data type for citation worked fine using both string "", and list []. We should allow users to add a list of citation when creating scripts. Proposed by @harshitbansal05

coolalexzb:
Does this issue still have any problem? It seems that multiple citations can be supported by using `citations: ["citatition for one", "citation for two"]`

henrykironde:
No we have not yet supported this

coolalexzb:
I want to make sure that I fully understand the issue.
Do you mean that the scripts need to support multiple citations using citations list
e.g `"citations": ["citatition for one", "citation for two"]`
[...]

PR #1313, opened 2019-04-17
Add table_citations field to support multiple citations

harshitbansal05:
This pull request tries to fix: #1303. It adds an extra field `table_citations`, to support multiple citations for a script.

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -70,7 +70,12 @@ def main():
                 scripts = name_matches(script_list, args.dataset)
                 for dataset in scripts:
                     print("\nDataset:  {}".format(dataset.name))
-                    print("Citation:   {}".format(dataset.citation))
+                    if hasattr(dataset, 'table_citations') and dataset.table_citations:
+                        print("Citations:   ")
+                        for index, citation in enumerate(dataset.table_citations):
+                            print("{count}. {citation}".format(count=index+1, citation=citation))
+                    else:
+                        print("Citation:   {}".format(dataset.citation))
                     print("Description:   {}\n".format(dataset.description))
 
             return
@@ -137,7 +142,7 @@ def main():
                     print(
                         "{count}. {title}\n {name}\n"
                         "{keywords}\n{description}\n"
-                        "{licenses}\n{citation}\n"
+                        "{licenses}"
                         "".format(
                             count=count,
                             title=script.title,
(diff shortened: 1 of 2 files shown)


ethanwhite:
Code looks good to me. Thanks @harshitbansal05!

I know I'm a bit late to the discussion on this, but I'm wondering about the need to add a new `table_citations` field instead of just using `citation`. It was mentioned that allowing lists in `citation` would break backward compatibility, but I'm not sure it would. `print("Citation:   {}".format(dataset.citation))` should produce something reasonable even if `dataset.citation` is a list. E.g.,

```python
In [2]: citation = ['first citation', 'second citation']

In [3]: print("Citation:   {}".format(citation))
Citation:   ['first citation', 'second citation']
```

Given that I'm wondering if we don't want to just use `citation` and check it's length/type to change the output format.

Apologies for bringing this up after a decision was made and I'm happy if you all decide to stick with this version. No need to convince me otherwise, just go for it. Just thought it was worth a quick mention.

henrykironde:
Thanks @ethanwhite for the input, after testing this I think we go with the proposed design.
@harshitbansal05 could you instead convert the previous` table_citations` for `ecoregions-us` to `citation = ['first citation', 'second citation']`

henrykironde:
I have put in the changes. Feel free to open and issue if we are wised something.

harshitbansal05:
@henrykironde, I agree with @ethanwhite. I was thinking, that if we now support multiple citations, we should also allow the user to provide them when creating a new json script. I would like to know your views on this.

henrykironde:
@harshitbansal05, that is something we should add. Nice catch. I will add an issue

harshitbansal05:
@henrykironde can the user input be taken in a manner similar to keywords, i.e. `;` separated citations. I following the same method in this pr.

henrykironde:
I was proposing that, but I think a simple list is okey. I have updated it to list

harshitbansal05:
sorry @henrykironde, I didn't quite understand. Could you explain it, please?

henrykironde:
@harshitbansal05, I have made the change [here](https://github.com/weecology/retriever/pull/1315/files#diff-fc5dbe2da6ff17615660b76e0fb766cdR8)

harshitbansal05:
@henrykironde may I work on the user input part?

henrykironde:
Go a head 👍

PR #1319, opened 2019-04-24
Change user input to support multiple citations

harshitbansal05:
This pull request works towards changing the user prompt to support multiple citations.

### DIFF
--- a/retriever/lib/datapackage.py
+++ b/retriever/lib/datapackage.py
@@ -160,7 +160,8 @@ def create_json():
 
     contents['title'] = clean_input("title: ", ignore_empty=True)
     contents['description'] = clean_input("description: ", ignore_empty=True)
-    contents['citation'] = clean_input("citation: ", ignore_empty=True)
+    contents['citation'] = clean_input("citations (separated by ';'): ",
+                                       split_char=';', ignore_empty=True)
     contents['homepage'] = clean_input("homepage (for the entire dataset): ", ignore_empty=True)
     contents['keywords'] = clean_input("keywords (separated by ';'): ",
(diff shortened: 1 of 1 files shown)


henrykironde:
@harshitbansal05 acton-lake is down.

retriever/lib/datapackage.py
```diff
@@ -160,7 +160,13 @@ def create_json():
 
     contents['title'] = clean_input("title: ", ignore_empty=True)
     contents['description'] = clean_input("description: ", ignore_empty=True)
-    contents['citation'] = clean_input("citation: ", ignore_empty=True)
+    citations = clean_input("citations (separated by ';'): ",
+                            split_char=';', ignore_empty=True)
+    contents['citation'] = ""
+    if len(citations) == 1:
```
henrykironde: I don't get what you are trying to check here
```python
>>> c="a"
>>> len(c)
1
```

harshitbansal05: @henrykironde, `citations` would return a list of user entered citations. If the user entered only a single citation, then instead of storing a list of only one element, the `citations[0]` would be stored. Else, the entire list would be stored.

henrykironde: In this case, I think it is good to assume that or enforce all new scripts' citations as lists.

henrykironde: In which case we may not need any checking

harshitbansal05: @henrykironde done.

harshitbansal05:
@henrykironde is it fine now?

henrykironde:
Thanks @harshitbansal05  for this

PR #1320, opened 2019-04-24
Solve minor error in setup.py

harshitbansal05:
This pull request corrects a minor error in `setup.py` as suggested by @henrykironde.


henrykironde:
Since acton-lakes is not available we could skip it using pytest -v -k "not test_fetch_order" or  commenting out the entire test function test_fetch_order. We should be passing for us to release.

harshitbansal05:
@henrykironde I have commented out the test function.

henrykironde:
Thanks for this.

henrykironde:
👍

Issue #1208, opened 2018-11-05
Improve efficiency of CSV installs using SQLite strategy


ethanwhite:
For the SQLite engine we read and right in chunks when possible. This makes large installs like `breed-bird-survey` really quick. The CSV engine is currently much slower (~15x) and I'm wondering if we might be able to use a similar strategy to the SQLite engine to get some big speed ups.

coolalexzb:
Hello, I have a question about this issue. When installing datasets by sqlite, it improves its efficiency by the use of `cursor.executemany`. When installing datasets by csv, it writes to a .csv file by `executemany` function in csvengine.py. In` executemany` function, when I print the value of statement or values(these two variable have the same value using csv engine), data is also divided in chunk and it only takes several times to write to a file. I think it makes use of a similar strategy with sqlite already. Is the result of a slower speed in csv caused by writing to a file directly rather than using database? @ethanwhite @henrykironde

henrykironde:
In the CSV engine, `executemany` still reads line by line and tries to write several of these. It is like `writelines` vs write for a single line. 
However, SQL engines can load data into the database without processing line by line. When data is clean we do not need to process the line by line. we need to more less load the whole file to the engine. I think for postgres that function is COPY and for sqlite we perform the same using read chuck as in the ref below
Some ref on sqlite: https://github.com/weecology/retriever/commit/8fff6cddb8253a1f396d1ad492214e478d8b4da3#diff-31ca15953e7058ab0b80525b04a2de29

coolalexzb:
I have improved the speed of large csv installs like ` breed-bird-survey` successfully, while I cannot pass pr test. I feel difficult to understand the error information. It seems a encoding problem. Could you give me some help? Besides, are there some ways to test locally rather than committing several times? The test functions need some data as parameters but it seems that I cannot provide it locally. Thanks! @henrykironde  @ethanwhite

coolalexzb:
I have improved the speed of large csv installs like breed-bird-survey successfully ＃1323, while I cannot pass pr test. I feel difficult to understand the error information. It seems a encoding problem. Could you give me some help? Besides, are there some ways to test locally rather than committing several times? The test functions need some data as parameters but it seems that I cannot provide it locally. Thanks! @henrykironde @ethanwhite

coolalexzb:
[...]

Issue #1322, opened 2019-04-28
[Bug] Data transformations with install_csv


ha0ye:
I noticed that data was importing slightly different on linux than on my Mac laptop for a particular data file and it seems to be an issue with how the CSV file is being generated on my Mac.

In all cases, I'm using retriever v2.3.0, and running from within Python3.
1. I use retriever to download dataset "veg-plots-sdl":
```python
import retriever
retriever.install_csv("veg-plots-sdl")
```

2. Then I look at the "Photo_info" file. On my mac, line 9 of the data (line 10 of the file) shows:
```
9,375,11/14/1988,Turner,9:30,Graphlex XL,,58,,280,up,20,1.45,2,0,0,,0
```
and on linux, it shows:
```
9,375,11/14/1988,Turner,9:30,Graphlex XL,,58,,280,up,2_0,1.45,2,0,0,,0
```
**notice the entry in column 12, which shows `20` vs. `2_0`**

3. Looking at the `veg_plots_sdl.json` script, I see the entry for Photo_info as:
```json
            "name": "Photo_info",
            "schema": {},
            "url": "https://ndownloader.figshare.com/files/5625354"
```

When I use that link to download the file directly, it seems to match up with the version of the file when downloading on Linux:
```
9,375,11/14/1988,Turner,9:30,Graphlex XL,NA,58,,280,up,2_0,1.45,2,0,0,NA,0
```

**Here, the NAs are explicitly present, but are not present when examining the CSV files downloaded by retriever.**

This suggests to me that there is some auto-conversion of CSVs operating under the hood of retriever, and that it does things slightly differently depending on OS. Perhaps, the missing values in the first rows of data cause that column to be set differently as a numeric or character column by default, and so the underscore gets dropped in one OS, but not another.

henrykironde:
Thank you @ha0ye, this needs some redesign.
```Python
Python 3.6.5
>>> 2_0
20
>>> 
Python 2.7.15 
>>> 2_0
  File "<stdin>", line 1
    2_0
      ^
```
I guess we shall have to change the way we detect the data type in https://github.com/weecology/retriever/blob/master/retriever/lib/engine.py#L265

ethanwhite:
For future reference (since I couldn't believe that 2_0 == 20): https://www.python.org/dev/peps/pep-0515/

PR #1323, opened 2019-04-29
Improve efficiency of CSV installs using SQLite strategy

coolalexzb:
Hi, I have improved the speed of making large installs like `breed-bird-survey` #1208 by using pandas' reading and writing functions and setting chunksize parameter . When installing and merging tables for `breed_bird_survey_counts.csv`, I do not add indexes into .csv file. I am glad that you can provide suggestion to my code. Thank you!  @henrykironde  @ethanwhite


henrykironde:
Hi @coolalexzb, this is a good contribution. A failure enables us to learn a lot more about the tools we are working on, hence I am opening the PR again and I will review it in detail later. Usually tools fail in this manner a lot and its good practice that you have tried so hard  👍and you have given it some good amount of effort.

coolalexzb:
Thanks for your help! I think another two problems are also need to be fixed. One is the problem of Nan value mentioned in #1322 . In some case Nan is seen as an empty string(''). The other may be an encoding problem. The two problems can be described clearly by the following prictures. @henrykironde

coolalexzb:
A question I want to ask is that if I need to wait and test my code in this request until all the problems are solved and new changes are merged into master branch. Thanks! @henrykironde

henrykironde:
@coolalexzb, usually we do need a PR to be working as good as the master branch. For example, if the master branch is passing on one platform i.e Travis, the PR should also be able to pass on that same platform. Sometimes PRs do wait for a long time before they start to pass and get merged. For GSOC, this gets to be part of your proposal's "adds on" since you are solving a problem that is part of your proposal. I believe this could take sometime to solve.

henrykironde:
Thanks @coolalexzb for this PR. I will close it because the Sqlite engine seems to be working in the same way

PR #1327, opened 2019-04-29
Fix and test Python underscore in values

### LINKED ISSUE #1322: [Bug] Data transformations with install_csv

henrykironde:
Python2 reads a csv value 2_0 as string
while Python3.6<current reads the same as int 20
change major version number because this changes
data installed by this  data package

Add Tests for with underscores
 
    
    

fixes #1322

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -401,6 +401,47 @@ change_header_values = {
     'expect_out': ['aa,bb,c_c', '1,2,3', '4,5,6']
 }
 
+underscore_csv = {
+    'name': 'underscore_csv',
+    'raw_data': ['a,b,c',
+                 '1,2,y',
+                 '4,5,2_0'],
+    'script': {"name": "underscore_csv",
+               "resources": [
+                   {"dialect": {"do_not_bulk_insert": "True"},
+                    "name": "underscore_csv",
+                    "schema": {
+                        "fields": [
+                            {
+                                "name": "record_id",
+                                "type": "pk-auto"
+                            },
+                            {
+                                "name": "a",
+                                "type": "int"
+                            },
+                            {
+                                "name": "b",
+                                "type": "int"
+                            },
+                            {
+                                "name": "c",
+                                "type": "char"
+                            }
+                        ]
+                    },
+                    "url": "http://example.com/underscore_csv.txt"
+                    }
+               ],
+               "retriever": "True",
+               "retriever_minimum_version": "2.0.dev",
+               "version": "1.0.0",
+               "urls": {"underscore_csv": "http://example.com/underscore_csv.txt"}
+               },
+    'expect_out': ['record_id,a,b,c', '1,1,2,y', '2,4,5,2_0']
+}
+
+
 tests = [
     simple_csv,
     comma_delimiter,
@@ -412,7 +453,9 @@ tests = [
     autopk_crosstab,
     skip_csv,
     extra_newline,
-    change_header_values]
+    change_header_values,
+    underscore_csv
+]
 
 # Create a tuple of all test scripts with their expected values
 test_parameters = [(test, test['expect_out']) for test in tests]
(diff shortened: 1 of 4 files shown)


test/test_integration.py
```diff
@@ -401,6 +401,86 @@
...
+               "urls": {"underscore_csv": "http://example.com/underscore_csv.txt"}
+               },
+    'expect_out': ['record_id,a,b,c', '1,1,2,y', '2,4,5,2_0']
+}
+
+underscore_int_csv = {
+    'name': 'underscore_int_csv',
+    'raw_data': ['a,b,c',
+                 '1,2,3',
+                 '4,5,20'],
```
ethanwhite: Were you thinking that the last value here should be `2_0` instead of `20`? If so I'm wondering whether or not we really want to read in a `2_0` from a csv file as `20`. I don't think in most cases that will be the intended value.

henrykironde: With the way things are moving, a new pep like "2_0", we shall def see files with such data in the near feature. I just wanted to be sure that we are handling both.

henrykironde: But I could remove that test in case we think it may not fit for now

ethanwhite: While I see Python folks using that notation in code I don't see it becoming standard usage in csv's since Python is the only language that has introduced that notation. Personally I feel like we should be treating `2_0` as a string regardless, but maybe this is a larger discussion that deserve's its own issue. So, let's remove that test for now, merge this, and open an issue to discuss what to do with that notation.

henrykironde: Sounds perfect

ethanwhite:
In it goes!

ha0ye:
I tried re-installing retriever from github, but am having the same issue.

Can you all check if there's something going on with that dataset I referenced in #1322? It has a lot of missing values on the first line, which I suspect may cause the data type to default to integer, even though the actual value shows up later with underscores.

henrykironde:
Are we talking about the same field that the 2_0 was

ha0ye:
Yes, it still shows as `20` for me.

ethanwhite:
@ha0ye - did you run `retriever update` to make sure you have the updated script?

ha0ye:
@ethanwhite Yep, still the same.

ha0ye:
Ok, @henrykironde helped me out - needed some combination of reseting the scripts and cached data, I think?

PR #1330, opened 2019-05-01
Change to python utf-8

henrykironde:
This is a work in progress that I wanted to be done by today.
I have tested with some of the unicode dataset like `elton-traits `, `mt-st-helens-veg`.

Added two new dataset that help us show that the platform works well with new characters sets
`catalogos-dados-brasil`, `transparencia-dados-abertos-brasil`

Feel free to add and test this PR. 
I think there is something small in the `engine_tools.py` model  or `to_csv` function that is not letting us pass.


retriever/lib/engine.py
```diff
@@ -811,7 +815,10 @@ def set_engine_encoding(self):
 
     def set_table_delimiter(self, file_path):
         """Get the delimiter from the data file and set it."""
-        dataset_file = open_fr(file_path)
+        # if self.script.encoding:
```
henrykironde: will delete this

henrykironde:
ref: #1337

PR #1333, opened 2019-05-02
Add datasets

henrykironde:
catalogos-dados-brasil
transparencia-dados-abertos-brasil

### DIFF
--- a/scripts/catalogos_dados_brasil.json
+++ b/scripts/catalogos_dados_brasil.json
@@ -1,0 +1,77 @@
+{
+    "citation": "Augusto Herrmann, Catálogos de dados abertos no Brasil, (2015), https://github.com/dadosgovbr/catalogos-dados-brasil",
+    "description": "Um catálogo de dados é uma coleção curada de metadados a respeito de conjuntos de dados",
+    "encoding": "utf-8",
+    "homepage": "https://github.com/dadosgovbr/catalogos-dados-brasil",
+    "keywords": ["brazil"],
+    "licenses": [{"name": "MIT"}],
+    "name": "catalogos-dados-brasil",
+    "resources": [
+        {
+            "dialect": {},
+            "name": "solucao",
+            "schema": {
+                "fields": [
+                    {
+                        "name": "solução",
+                        "type": "char"
+                    },
+                    {
+                        "name": "tipo_de_licença",
+                        "type": "char"
+                    },
+                    {
+                        "name": "url",
+                        "type": "char"
+                    },
+                    {
+                        "name": "conceito",
+                        "type": "char"
+                    }
+                ]
+            },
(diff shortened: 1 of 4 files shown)


scripts/catalogos_dados_brasil.json
```diff
@@ -0,0 +1,77 @@
...
+                    {
+                        "name": "solução",
+                        "type": "char"
+                    }
+                ]
+            },
+            "url": "https://raw.githubusercontent.com/dadosgovbr/catalogos-dados-brasil/master/dados/catalogos.csv"
+        }
+    ],
+    "retriever": ["True"],
```
ethanwhite: This is a string not a list in most of the other scripts. Should it be the same here?

scripts/catalogos_dados_brasil.json
```diff
@@ -0,0 +1,77 @@
...
+    "retriever_minimum_version": "2.1.0",
```
ethanwhite: Since this involves the new encoding work does the minimum version need to be "2.3.2" or "2.4.0" depending on what version bump you're using?

scripts/transparencia_dados_abertos_brasil.json
```diff
@@ -0,0 +1,73 @@
...
+                    {
+                        "name": "observação",
+                        "type": "char"
+                    }
+                ]
+            },
+            "url": "https://raw.githubusercontent.com/augusto-herrmann/transparencia-dados-abertos-brasil/master/data/portais-municipais.csv"
+        }
+    ],
+    "retriever": ["True"],
```
ethanwhite: This is a string not a list in most of the other scripts. Should it be the same here?

scripts/transparencia_dados_abertos_brasil.json
```diff
@@ -0,0 +1,73 @@
...
+    "retriever_minimum_version": "2.1.0",
```
ethanwhite: Since this involves the new encoding work does the minimum version need to be "2.3.2" or "2.4.0" depending on what version bump you're using?

scripts/catalogos_dados_brasil.json
```diff
@@ -0,0 +1,77 @@
+{
+    "citation": "NA",
```
ethanwhite: I think we should use a citation like the one in this answer since they don't indicate a preferred citation:
https://academia.stackexchange.com/a/14015

scripts/transparencia_dados_abertos_brasil.json
```diff
@@ -0,0 +1,73 @@
+{
+    "citation": "NA",
```
ethanwhite: I think we should use a citation like the one in this answer since they don't indicate a preferred citation:
https://academia.stackexchange.com/a/14015

henrykironde:
@ethanwhite this is ready for review

PR #1337, opened 2019-05-29
Use latin-1 as default 

henrykironde:
Python 3 uses 'UTF-8' as the default encoding, however some scripts
can not be processed under this protocol. 
The Data retriever maintains the use of latin-1 as default
based on performance of installation of unicode characters in some scripts


henrykironde:
@ethanwhite, about the issue I wanted to discus, I did finally resolve and I guess looks good to go after this version change commit

ethanwhite:
Awesome. Merging now. Release when ready.

PR #1338, opened 2019-05-30
Add functions for committing dataset

apoorvaeternity:


retriever/lib/engine_tools.py
```diff
@@ -199,7 +199,7 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
     return outputfile
 
 
-def getmd5(data, data_type='lines', encoding='utf-8'):
+def getmd5(data, data_type='lines', encoding=ENCODING):
```
henrykironde: I do not recommend changing this line.

apoorvaeternity: It was causing encoding issues when trying to find the `md5` of raw data of `aquatic-animal-excretion`.

henrykironde: If you are dealing with this script, you may want to start debugging from 
`retriever install --debug csv  aquatic-animal-excretion`

apoorvaeternity: errr
It is giving the same encoding error. Working fine if I change the encoding to `ISO-8859-1` for the script.

retriever/lib/provenance.py
```diff
@@ -0,0 +1,95 @@
...
+from zipfile import ZipFile
+
+from retriever import datasets
+from retriever.engines import choose_engine
+from retriever.lib.defaults import HOME_DIR, VERSION, ENCODING
+from retriever.lib.engine_tools import getmd5
+
+
+def package_details():
+    details = {}
```
henrykironde: details = {'retriever': VERSION[1:]}

retriever/lib/provenance.py
```diff
@@ -0,0 +1,95 @@
...
+    return info
+
+
+def commit(dataset, commit_message='', path='.', quiet=False):
+    """
+    Commit dataset to a zipped file.
+    """
+    if type(dataset) == str:
+        # if dataset is not a dataset script object find the right script
+        dataset = [script for script in datasets() if script.name == dataset][0]
```
henrykironde:
```python
paths_to_zip = {'script': dataset._file,
                'raw_data': []}
```

apoorvaeternity:
CLI usage:
`retriever commit abalone-age -m "commit message" -p path_to_store_archive_file`

henrykironde:
Thanks @apoorvaeternity, hope it covers all added functionality. For example, if I give `retriever commit abalone-aged, I should be able to get some good help messages. I have not tested yet.

henrykironde:
Just tested this and it looks like the commit are good. We should discuss more when you are free.

retriever/lib/provenance.py
```diff
@@ -0,0 +1,92 @@
...
+                    zipped.write(data_file, data_file.replace(raw_dir, ""))
+                with open('metadata.json', 'w') as json_file:
+                    json.dump(info, json_file, sort_keys=True, indent=4)
+                zipped.write(os.path.abspath(json_file.name), 'metadata.json')
+                os.remove('metadata.json')
+        if not quiet:
+            print("Successfully committed.")
+    except Exception as e:
+        print("Dataset could not be committed.")
+        raise (e)
```
henrykironde: Avoid raising, in case we have to use other programs, the raise will interfere with the python interpreter. Let's print the error e and use return.

retriever/lib/provenance.py
```diff
@@ -0,0 +1,92 @@
...
+            'time': datetime.now(timezone.utc).strftime("%m/%d/%Y, %H:%M:%S"),
+            'version': dataset.version}
+    return info
+
+
+def commit(dataset, commit_message='', path='.', quiet=False):
+    """
+    Commit dataset to a zipped file.
+    """
+    if type(dataset) == str:
```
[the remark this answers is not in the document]
apoorvaeternity: If we don't pass a dataset script object but only the name of dataset it should be able to handle it.

retriever/lib/provenance.py
apoorvaeternity: Okay :+1:

apoorvaeternity:
@henrykironde How does it look now?

henrykironde:
Looks good. I think we can make the tests more comprehensive . Lets discuss more when you get time, feel free to call me.

apoorvaeternity:
Once #1343 gets merge I will update the URLs for the raw data of sample-dataset that is being used for testing.

henrykironde:
Update urls

apoorvaeternity:
@henrykironde Do I have to squash the commits?

apoorvaeternity:
For python interface use the following syntax
```python
from retriever import commit
commit("abalone-age", commit_message="First Commit", path=".", quiet=False)
```

apoorvaeternity:
@henrykironde Don't know why the some other test on AppVeyor failed :thinking:

apoorvaeternity:
Passed after re-triggering the build.

henrykironde:
Add docs for these current changes, and I think we should move this to `retriever-recipes-dev` instead if `master`

apoorvaeternity:
As discussed above, closing this in favour of #1346 .

PR #1339, opened 2019-05-30
Use encoding latin-1

henrykironde:


henrykironde:
@ha0ye and @diazrenata this should work fine one we merge the PR

henrykironde:
Urgently needed change.

PR #1341, opened 2019-06-07
Make changes for the retriever scripts being shipped to a separate repository

harshitbansal05:


hooks/pre-commit
```diff
@@ -47,7 +47,7 @@ def get_script_version(file_name):
 
 # check and get script directory changes
 subprocess.check_output(
-    'git fetch https://github.com/weecology/retriever.git master', shell=True)
+    'git fetch https://github.com/harshitbansal05/retriever.git development', shell=True)
```
henrykironde: Do we want to keep this here `harshitbansal05`

hooks/pre-commit
```diff
@@ -63,8 +63,12 @@ modified_staged_scripts = to_string(subprocess.check_output(
 added_deleted_staged_scripts = to_string(subprocess.check_output(
     "git diff --cached --name-only --diff-filter=AmD scripts", shell=True))
 
+if added_deleted_staged_scripts != "":
+	print("\nThe legacy scripts of Retriever cannot be added to or removed. To add scripts,\
+	use the retriever-recipes repository.")
+	exit(1)
```
henrykironde: Avoid using exit, instead return would be better. exit(1) will cause the tests to fail in later edge cases especially in the Retriever.Julia package and in the rdataretriever package.

hooks/pre-commit
```diff
@@ -88,10 +92,9 @@ if modified_staged_scripts != "":
...
-# only runs version.py if we have modified scripts which have had their versions updated,
-# or we have no modified scripts but either deleted or added scripts.
-if (modified_staged_scripts and versions_updated) or (added_deleted_staged_scripts != ""):
+# only runs version.py if we have modified scripts which have had their versions updated.
+if (modified_staged_scripts and versions_updated):
     print("Updating version.txt")
     subprocess.call(['python', 'version.py'])
     print("version.txt added to commit")
-    subprocess.call(['git', 'add', 'version.txt'])
+    subprocess.call(['git', 'add', 'version.txt'])
```
henrykironde: Line end

hooks/pre-commit
harshitbansal05: Right now, I kept this only for testing. Later, we can change it to retriever repository.

zhangcandrew: I think it's better if we change it to the retriever-recipes repository right now instead of your temporary git. We can wait for [retriever-recipes PR 2](https://github.com/weecology/retriever-recipes/pull/2) to go in first before merging this in.

retriever/lib/get_opts.py
```diff
@@ -52,9 +49,6 @@
 default_parser = subparsers.add_parser('defaults', help='displays default options')
 update_parser = subparsers.add_parser('update', help='download updated versions of scripts')
 new_parser = subparsers.add_parser('new', help='create a new sample retriever script')
```
zhangcandrew: Is it possible to make this new parser create scripts within the scripts folder

retriever/lib/repository.py
```diff
@@ -69,9 +70,14 @@ def check_for_updates():
...
-                                              os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)))
+                                              os.path.normpath(os.path.join(SCRIPT_WRITE_PATH, script_name)),
+                                              repo)
                 except Exception as e:
                     print(e)
                     pass
     except:
         raise
+
+    if repo == RETRIEVER_REPOSITORY:
```
zhangcandrew: Can you run me through your logic for checking if the repo is a scripts repo vs just the normal retriever repo?

retriever/lib/scripts.py
```diff
@@ -161,6 +163,61 @@ def to_str(object, object_encoding=sys.stdout, object_decoder=ENCODING):
...
+        if json_file_encoding:
+            file_obj = open_fr(json_file, encoding=json_file_encoding)
+        else:
+            file_obj = open_fr(json_file)
+        json_object = json.load(file_obj)
+        file_obj.close()
+    except ValueError:
+        pass
+
+    if type(json_object) is dict and "version" in json_object.keys():
```
zhangcandrew: perhaps: 

`return json_object.get("version", None)`?

zhangcandrew:
Thanks @harshitbansal05, It's looking good, I'm going to Pull it down and test some parts out when I get the chance!

hooks/pre-commit
harshitbansal05: Ok, I would change it as soon as recipes pr is merged.

hooks/pre-commit
harshitbansal05: Would do so.

retriever/lib/get_opts.py
harshitbansal05: We can so it, should I open a separate pr for it or do it in this?

hooks/pre-commit
harshitbansal05: Noted.

retriever/lib/repository.py
harshitbansal05: @zhangcandrew in the method `check_for_updates`, it is done from both the repositories, retriever (for the five legacy scripts) and recipes (for the rest of the scripts). I just used the above  logic to call the same method with different repositories (to update from both of them).

retriever/lib/scripts.py
harshitbansal05: Noted.

harshitbansal05:
@henrykironde is it fine now?

retriever/lib/get_opts.py
zhangcandrew: Let's open a separate PR for this

retriever/lib/repository.py
zhangcandrew: Ok I see, so this is a placeholder for now, got it!

setup.py
```diff
@@ -139,3 +135,7 @@ def read(*names, **kwargs):
     compile()
 except:
     pass
+
+if os.path.exists(".git/hooks"):  # check if we are in git repo
```
zhangcandrew: Is there a reason for moving this to the bottom?

test/test_modified_scripts.py
```diff
@@ -37,10 +32,9 @@ def setup_module():
 def get_modified_scripts():
     """Get modified script list, using version.txt in repo and master upstream"""
 
-    os.chdir(retriever_root_dir)
     modified_list = []
-    version_file = requests.get("https://raw.githubusercontent.com/weecology/retriever/master/version.txt")
-    local_repo_scripts = get_script_version()  # local repo versions
+    version_file = requests.get("https://raw.githubusercontent.com/harshitbansal05/retriever/development/version.txt")
```
zhangcandrew: same as above, let's make this the retriever-recipes-dev branch that was just merged in

test/test_modified_scripts.py
```diff
@@ -52,11 +46,8 @@ def get_modified_scripts():
         local_script, local_version = item.lower().split(",")
         # check for new scripts or a change in versions for present scripts
         # repo script versions compared with upstream.
-        if local_script not in upstream_versions.keys():
+        if LooseVersion(local_version) != upstream_versions[local_script]:
```
zhangcandrew: I think this should be greater than since if we have an outdated script locally, it probably wasn't modified and we probably don't care to test it. Maybe @henrykironde has more to say on this though

test/test_modified_scripts.py
```diff
@@ -108,13 +95,13 @@ def install_modified():
                   'table_name': '{db}.{table}'},
 
         "xml": {'engine': 'xml',
-                'table_name': 'output_file_{table}.xml'},
+                'table_name': '{db}_{table}.xml'},
```
zhangcandrew: @henrykironde is this formatting ok? I know we had trouble in the past with some output names but I'm not knowledgeable on the current status of how this works.

hooks/pre-commit
zhangcandrew: alright, recipes pr should be merged

zhangcandrew:
I think it's coming together. One thing I forgot to mention, can you output a line in the retriever ls function that says something like 

"For a full list of all script available to download, go to: [retriever-recipes-link] here". 

Also, on update, is it possible to note a distinction of which scripts are donwloading? (like downloading default scripts, downloading upstream scripts) something like that. It currently shows like this: 
image

hooks/pre-commit
harshitbansal05: @zhangcandrew the above url refers to the retriever repository, which has not been merged. I have updated the url for the recipes scripts repository.

setup.py
harshitbansal05: No, I don't know how. I would correct it.

test/test_modified_scripts.py
harshitbansal05: Again, I have updated the url for the recipes repository. I didn't update this one so the pr could be tested prior to merging it in the retriever repo.

test/test_modified_scripts.py
harshitbansal05: I agree with you @zhangcandrew.

test/test_modified_scripts.py
henrykironde: You should checkout the format from https://github.com/weecology/retriever/blob/961974cb18940e26d5dd86f5f04ed157d08b3c7f/test/test_integration.py#L546

setup.py
zhangcandrew: I think you can try `git checkout HEAD -- setup.py` to reset it to head.

test/test_modified_scripts.py
zhangcandrew: sure that makes sense, but is there a reason why you didn't use the official weecology retriever-recipes version.txt? (https://raw.githubusercontent.com/weecology/retriever-recipes/retriever-recipes-dev/version.txt)

image

hooks/pre-commit
zhangcandrew: ahh ok I see what you mean by this then. I see you have 6 commits, let's split this PR into two PR's one with just the changes up to 78f5d64eeb09a9194dc5a75cadba3d029a9ee702, and we can have another with commits after. That way we can get the first half in and then commit the second half successfully without having to reference an intermediary repo like this.

test/test_modified_scripts.py
zhangcandrew: ok just double checking, this looks about right

zhangcandrew:
Closing PR in reference to #1345. Feel free to re-open if anyone feels the need

PR #1343, opened 2019-06-13
Add raw data for sample-dataset

apoorvaeternity:
Please merge this before #1338 because this data is needed for testing in #1338


henrykironde:
Is it possible to call this something else instead of `Portal_rodents_19772002.csv`, and also change the name of the script?

apoorvaeternity:
Yeah. What should I name it to? I have named the script as `sample-dataset`.

henrykironde:
May be `sample_dataset.csv`

apoorvaeternity:
Okay. So keeping the name of the script as `sample-dataset`?

henrykironde:
I think that works or even something like `sample-dataset-provenance` or` dataset-provenance` or `test-dataset-provenance`.

PR #1344, opened 2019-06-17
Move scripts from retriever to recipes repository

harshitbansal05:
@henrykironde @zhangcandrew the small pr for shifting scripts to the recipes repository.


zhangcandrew:
I think this looks good, merging this in to stop it from blocking

PR #1345, opened 2019-06-18
Move script editing functions to the recipes repository

harshitbansal05:
@henrykironde @zhangcandrew another subpart of the original pr. It shifts the script's adding and editing functionalities to the recipes repository.


henrykironde:
So are we closing the original PR? #1341

harshitbansal05:
Yes, @henrykironde we would close the original pr, as Andrew said it would be better to split the pr into smaller pr's.

zhangcandrew:
This looks good to me, can you make appropriate changes (either in this PR or another PR) to the retriever docs to remove any reference to editing jsons etc?

harshitbansal05:
@zhangcandrew, the doc changes would be large. O think we can do it in a separate pr.

harshitbansal05:
Also, should I simply remove the docs related to addition of scripts?

zhangcandrew:
If you want you can try to migrate them over to retriever-recipes since the recipes repo will end up needing a new set of doc

henrykironde:
@zhangcandrew, @harshitbansal05, where do you think we should put `autocreate_parser` functionalities? 
At this point we are have stand alone functions, I think we should make `retriever-recipes` a package with both CLI and interface, what do you think.

harshitbansal05:
@henrykironde @zhangcandrew I removed the docs related to adding/editing scripts. However, more discussion is needed regarding the section 'Datasets Available'.

zhangcandrew:
@henrykironde, I agree with that sentiment. I think retriever-recipes should have its own CLI to run, which I think can be implemented in a later PR after we've leveraged the changes we're trying to put in currently

harshitbansal05:
@zhangcandrew @henrykironde I agree too, and I have begun the work to add a CLI interface to recipes repo.

zhangcandrew:
This looks fine. Merging in, going to test out the retriever-recipes CLI

Issue #1353, opened 2019-06-27
Codec error while installing some datasets


harshitbansal05:
Highlighted by @zhangcandrew, the installation of some datasets raises the following error: ![](https://user-images.githubusercontent.com/15080652/60234319-c37d3580-9871-11e9-95a5-12c8baf63fcc.png)

henrykironde:
Try to include `"encoding": "latin-1"`, in `mt-st-helens-veg's` json file.

PR #1354, opened 2019-06-27
Fix codec error

harshitbansal05:
Fixes: #1353


henrykironde:
Thanks @harshitbansal05 and @zhangcandrew

PR #1356, opened 2019-06-27
Install committed datasets

apoorvaeternity:


apoorvaeternity:
Usage example:
`retriever install sqlite portal-b8b17d.zip`

apoorvaeternity:
@henrykironde It is working for the datasets with `JSON` scripts. For installing committed datasets with `python` scripts I will have to modify the download function of the python scripts of the dataset.
How to move forward with this?

henrykironde:
@apoorvaeternity could you discus this with @harshitbansal05, 
I had a discussion with him about this. 
I will be busy the next few days but will try to look into it.

apoorvaeternity:
@henrykironde Turns out my AppVeyor build was failing because for paths inside zip file I was using `os.path.join` and on a Windows OS it would generate the path according to Windows but according to Zip standards a zip file uses UNIX style file path. The build is passing now. The PR is ready for a review.

retriever/__main__.py
```diff
@@ -180,26 +181,29 @@ def main():
...
-                try:
-                    dataset.download(engine, debug=debug)
-                    dataset.engine.final_cleanup()
-                except KeyboardInterrupt:
-                    pass
-                except Exception as e:
-                    print(e)
-                    if debug:
-                        raise
+            if args.dataset.endswith('.zip'):
```
henrykironde: @apoorvaeternity, I guess we are testing if a file is a zip, but would this be the best or optimal way to do that. What if I create a text file as `textile.zip`, what would happen?

apoorvaeternity: It will raise `zipfile.BadZipFile: File is not a zip file` error. I think we can print a better message.

henrykironde: Also something I noted is that we are raising, this will stop the whole process especially when we interface with other languages like R or Julia. I think we could make good use of `return` in this case.

> It will raise zipfile.BadZipFile: File is not a zip file error. I think we can print a better message.

I will test `textile.zip`, and let you know what is happening.

apoorvaeternity: Yes. I will make the change so that we do not raise the errors.

apoorvaeternity:
@henrykironde Ready for a review.

retriever/__main__.py
```diff
@@ -180,26 +181,29 @@ def main():
             sys.tracebacklimit = 0
 
         if hasattr(args, 'debug') and args.not_cached:
-            engine.use_cache = False
+            use_cache = False
```
henrykironde: Why are we changing this? I think we should leave it the way it was

retriever/__main__.py
```diff
@@ -180,26 +181,29 @@ def main():
...
-                    dataset.download(engine, debug=debug)
-                    dataset.engine.final_cleanup()
-                except KeyboardInterrupt:
-                    pass
-                except Exception as e:
-                    print(e)
-                    if debug:
-                        raise
+            if args.dataset.endswith('.zip'):
+                _install(vars(args), debug=debug, use_cache=use_cache)
```
henrykironde: `_install(vars(args), debug=debug, use_cache=engine.use_cache)`

retriever/lib/provenance.py
```diff
@@ -88,3 +93,98 @@ def commit(dataset, commit_message='', path='.', quiet=False):
         print(e)
         print("Dataset could not be committed.")
         return
+
+
+def get_metadata(path_to_archive):
+    "Returns a dictionary after reading metadata.json file of a committed dataset"
+    with ZipFile(os.path.normpath(path_to_archive), 'r') as archive:
+        try:
+            metadata = json.loads(archive.read('metadata.json').decode('utf-8'))
```
henrykironde: Why are we decoding ?

retriever/lib/provenance.py
```diff
@@ -88,3 +93,98 @@ def commit(dataset, commit_message='', path='.', quiet=False):
...
+        except Exception as e:
+            print(e)
+            return
+    return metadata
+
+
+def commit_info_for_installation(metadata_info):
+    info = {'commit_message': metadata_info['commit_message'], 'time': metadata_info['time'], 'package_not_found': {},
```
henrykironde: Lines are too long

retriever/lib/provenance.py
```diff
@@ -88,3 +93,98 @@ def commit(dataset, commit_message='', path='.', quiet=False):
...
+
+
+def get_script(path_to_archive):
+    """
+    Reads script from archive.
+    """
+    with ZipFile(os.path.normpath(path_to_archive), 'r') as archive:
+        try:
+            commit_details = get_metadata(path_to_archive)
+            workdir = mkdtemp(dir=os.path.dirname(path_to_archive))
```
henrykironde: Let's use  'work_dir`

henrykironde:
Thanks @apoorvaeternity.

PR #1357, opened 2019-06-29
Changing os.system to subprocess.call

zhangcandrew:
Noticed that os.system wasn't working in our setup.py. 

This is likely due to the missing ```#!/usr/bin/env python``` at the top of setup.py, but we can keep the top of setup.py the same if we use subprocess calls in place of os.system. Subprocess calls seem like the python standard for making shell calls, and this might be a bigger discussion about changing the rest of our leftover os.system calls to subprocess.call calls.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 import os
 import re
 import platform
+import subprocess
 
 from pkg_resources import parse_version
 from setuptools import setup, find_packages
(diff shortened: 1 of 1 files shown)


ethanwhite:
This change seems fine to me. I think `subprocess` is generally preferred now anyway if I recall correctly. Can you check the failing test on Appveyor to see what's going on there?

henrykironde:
@apoorvaeternity could you confirm/check with this. Andrew was able to get appveyor working, let me know if it's the same issue

zhangcandrew:
@henrykironde @apoorvaeternity For some reason appveyor was not running because of the test_reset_retriever test. Making no changes, I made a new commit and force pushed it up and it worked. This may have been a fluke (although I'm sure I only pushed up the changes I made and none more), but I don't know if anyone else has also run into problems with test_reset_retriever. 

image

henrykironde:
I think this is good to go in.

harshitbansal05:
@zhangcandrew I think this happened, because`test_reset_retriever` picks a random dataset from the offline scripts and deletes it. In the above case, that random dataset happened to be test-raster-bio2, which is not available upstream. Thus, after deletion it could not be downloaded again, causing the tests to fail.

Issue #1290, opened 2019-03-24
[discussion] Extract Scripts into a separate system


harshitbansal05:
Hi, I was looking to work on the GSoC idea "Extract Scripts into a separate system". I researched for the project, and as far as I understand, it requires moving the scripts directory into a new repository for ease of maintenance(users downloading from source have to download the scripts directory as well because it comes with retriever). For moving it into a new repository, there have to be certain changes in the codebase where scripts are updated, edited, etc. However, I find this project somewhat small. Is there something am I missing? Also, please correct me if my understanding of the problem statement is incorrect.

harshitbansal05:
@ethanwhite @henrykironde @zhangcandrew sorry I was a bit involved in my midterm examinations, so could not create an issue earlier. Could you please add your suggestions here?

henrykironde:
I will add Andrew's comments here.

>  So the "Extract Scripts into a seperate system" project has a few more moving parts to it that we'd like to do. It deals a bit with how the retriever currently downloads and keeps track of scripts, how scripts are versioned and listed, portability to some of our other wrapper projects (R and Julia retrievers), and some other bits and pieces I can go into more detail about. If you want to start a discussion issue on retriever, we can talk about it a bit more in depth.

zhangcandrew:
Thanks @harshitbansal05! 

Some of the items we had hoped to tackle with this included:
- Moving all scripts to a separate area (new repository/cloud storage/whatever else sounds good) where it could be separately added to, versioned, and controlled
- Keeping track of scripts that would now be split across three different areas. Retriever looks in two local folders (~/.retriever and local retriever git repo) for scripts, and with this addition, the upstream repo. Among these three areas, it would be useful to keep track of which ones can be used offline (the local folders) and which ones would need to be downloaded (new upstream repo). This is something that was started and could be expanded upon from our retriever dash project (https://github.com/weecology/retrieverdash). 
- Change retriever ls to reflect scripts in different places (see above)
- Restructuring code to not redownload everything unless needed. Right now retriever creates a singleton class which by default downloads all scripts. With the addition of separate scripts, we should be able to check which scripts are available locally and upstream without downloading all the scripts, and ONLY download a script when we want to install from it. Some of our CLI commands also need other information that is embedded in the script which we would ideally be able to retriever without having to download all scripts still. 
- Our Julia and R projects rely on our retriever python project, so there may be some work that needs to be done on those once we move the scripts over. 
[...]

PR #1358, opened 2019-07-03
Download scripts only when specifically needed

harshitbansal05:
@zhangcandrew In the current pr, I have implemented `retriever ls` in the method I mentioned you, I am working on the method you mentioned.


test/test_modified_scripts.py
```diff
@@ -46,7 +46,7 @@ def get_modified_scripts():
         local_script, local_version = item.lower().split(",")
         # check for new scripts or a change in versions for present scripts
         # repo script versions compared with upstream.
-        if LooseVersion(local_version) != upstream_versions[local_script]:
+        if LooseVersion(local_version) > upstream_versions[local_script]:
```
harshitbansal05: @zhangcandrew @henrykironde the change from the previous pr that we needed.

test/raw_data_gis/scripts/test_vector.json
```diff
@@ -5,7 +5,7 @@
     "format": "vector",
     "spatial_ref": "ED50",
     "citation": "Test citation",
-    "license": [{"name": "CC Tests"}],
+    "licenses": [{"name": "CC Tests"}],
```
harshitbansal05: retriever uses the keyword `licenses` to recognize licenses.

test/raw_data_gis/scripts/test_us_eco.json
```diff
@@ -5,7 +5,6 @@
   "format": "vector",
   "spatial_ref": "NAD83",
   "citation": "N/A",
-  "license": [],
```
harshitbansal05: If no license is to provided, it must simply not be mentioned, providing an empty list results in error when any operation like `citation` is done.

test/raw_data_gis/scripts/test_raster_bio2.json
```diff
@@ -13,7 +13,6 @@
   "extract_all": "True",
   "archived": "zip",
   "driver": "GTiff",
-  "licenses": [],
```
harshitbansal05: If no license is to provided, it must simply not be mentioned, providing an empty list results in error when any operation like `citation` is done.

test/raw_data_gis/scripts/test_eco_level_four.json
```diff
@@ -5,7 +5,6 @@
   "format": "vector",
   "spatial_ref": "PROJCS[\"NAD83 / Conus Albers\",GEOGCS[\"NAD83\",DATUM[\"North_American_Datum_1983\",SPHEROID[\"GRS 1980\",6378137,298.257222101,AUTHORITY[\"EPSG\",\"7019\"]],TOWGS84[0,0,0,0,0,0,0],AUTHORITY[\"EPSG\",\"6269\"]],PRIMEM[\"Greenwich\",0,AUTHORITY[\"EPSG\",\"8901\"]],UNIT[\"degree\",0.0174532925199433,AUTHORITY[\"EPSG\",\"9122\"]],AUTHORITY[\"EPSG\",\"4269\"]],PROJECTION[\"Albers_Conic_Equal_Area\"],PARAMETER[\"standard_parallel_1\",29.5],PARAMETER[\"standard_parallel_2\",45.5],PARAMETER[\"latitude_of_center\",23],PARAMETER[\"longitude_of_center\",-96],PARAMETER[\"false_easting\",0],PARAMETER[\"false_northing\",0],UNIT[\"metre\",1,AUTHORITY[\"EPSG\",\"9001\"]],AXIS[\"X\",EAST],AXIS[\"Y\",NORTH],AUTHORITY[\"EPSG\",\"5070\"]]",
   "citation": "Test data",
-  "license": [],
```
harshitbansal05: Mentioned below.

test/raw_data_gis/scripts/test_raster_bio1.json
```diff
@@ -13,7 +13,6 @@
   "extract_all": "True",
   "archived": "zip",
   "driver": "GTiff",
-  "licenses": [],
```
harshitbansal05: Mentioned below.

setup.py
```diff
@@ -133,9 +133,7 @@ def read(*names, **kwargs):
 
 try:
     from retriever.compile import compile
-    from retriever.lib.repository import check_for_updates
 
-    check_for_updates()
```
harshitbansal05: Should not download all the scripts during installation.

retriever/lib/scripts.py
```diff
@@ -107,7 +108,101 @@ def get_script(dataset):
...
     else:
-        raise KeyError("No dataset named: {}".format(dataset))
+        if dataset in RETRIEVER_DATASETS:
+            read_script = get_script_upstream(dataset, repo=RETRIEVER_REPOSITORY)
+        else:
+            read_script = get_script_upstream(dataset)
+        if read_script is None:
+            raise KeyError("No dataset named: {}".format(dataset))
+        else:
+            return read_script
```
harshitbansal05: First checks if the script is available locally, if present returns it. Then, checks if it is a `retriever legacy` script or not. Then downloads it from the according upstream repository.

retriever/lib/scripts.py
```diff
@@ -107,7 +108,101 @@ def get_script(dataset):
...
+            setattr(read_script, "_name", script)
+            return read_script
+        else:
+            file, pathname, desc = imp.find_module(script, [SCRIPT_WRITE_PATH])
+            new_module = imp.load_module(script, file, pathname, desc)
+            setattr(new_module.SCRIPT, "_file", os.path.join(SCRIPT_WRITE_PATH, script_name))
+            setattr(new_module.SCRIPT, "_name", script)
+            return new_module.SCRIPT
+    except:
+        raise
```
harshitbansal05: Downloads the script from the upstream repo (mentioned in the argument), and then returns it as an instance of `BasicTextTemplate` taking into consideration whether it is a json or a python script.

retriever/lib/scripts.py
```diff
@@ -107,7 +108,101 @@ def get_script(dataset):
...
+                r = requests.get(search_url.format(query=k))
+                r = r.json()
+                for index in range(r['total_count']):
+                    script = r['items'][index]['name']
+                    script = '.'.join(script.split('.')[:-1])
+                    script = script.replace('_', '-')
+                    result_scripts.add(script)
+            except:
+                raise
+    return sorted(result_scripts)
```
harshitbansal05: If no keywords or licenses argument is provided, it reads the `version.txt` of the according repo and returns an array of script names with a maximum length of 100 to prevent flooding of terminal when doing `retriever ls`. When a query is provided, Github's search api is used (based on initial discussion with @henrykironde) to search scripts with the specified license or citation query parameter. Github's search api can be used to used to search all the script files in the `scripts` folder for the query param. However, it only searches the master branch, and thus may not produce desired results as of now, but would work when the branch is merged into master.

retriever/lib/install.py
```diff
@@ -17,9 +15,6 @@ def _install(args, use_cache, debug):
     engine.use_cache = use_cache
 
     script_list = SCRIPT_LIST()
-    if not (script_list or os.listdir(SCRIPT_WRITE_PATH)):
-        check_for_updates()
-        script_list = SCRIPT_LIST()
```
harshitbansal05: Should not download all the scripts if no offline script is present.

retriever/lib/download.py
```diff
@@ -23,9 +20,6 @@ def download(dataset, path='./', quiet=False, sub_dir='', debug=False, use_cache
     engine.use_cache = use_cache
 
     script_list = SCRIPT_LIST()
-    if not script_list or not os.listdir(SCRIPT_WRITE_PATH):
-        check_for_updates()
-        script_list = SCRIPT_LIST()
     scripts = name_matches(script_list, args['dataset'])
```
harshitbansal05: Should not download all the scripts if no offline script is present.

retriever/lib/engine_tools.py
```diff
@@ -75,6 +75,16 @@ def name_matches(scripts, arg):
...
+    from retriever.lib.scripts import get_script_upstream
+
+    if arg in RETRIEVER_DATASETS:
+        read_script = get_script_upstream(arg, repo=RETRIEVER_REPOSITORY)
+    else:
+        read_script = get_script_upstream(arg)
+
+    if read_script is not None:
+        return [read_script]
+
```
harshitbansal05: First checks if the script is available locally, if present returns it. Then, checks if it is a retriever legacy script or not. Then downloads it from the according upstream repository.

retriever/lib/get_opts.py
```diff
@@ -28,6 +28,10 @@
                 if dict_items['name']:
                     licenses_list.append(dict_items['name'])
 
+script_list.extend(get_dataset_names_upstream(repo=RETRIEVER_REPOSITORY))
+script_list.extend(get_dataset_names_upstream())
+script_list = list(set(script_list))
```
harshitbansal05: Add the upstream scripts to the choices list.

retriever/lib/datasets.py
```diff
@@ -27,19 +30,24 @@ def datasets(keywords=None, licenses=None):
...
                 for k in keywords:
                     if script_keywords.find(k.lower()) != -1:
-                        result_scripts.add(script)
+                        result_scripts_offline.add(script)
                         break
-    return sorted(list(result_scripts), key=lambda s: s.name.lower())
+    result_scripts_offline = sorted(list(result_scripts_offline), key=lambda s: s.name.lower())
+    result_retriever_scripts_online = get_dataset_names_upstream(keywords, licenses, repo=RETRIEVER_REPOSITORY)
+    result_scripts_online = get_dataset_names_upstream(keywords, licenses)
+    return dict({'online': result_retriever_scripts_online + result_scripts_online, 'offline': result_scripts_offline})
```
harshitbansal05: Search both online and offline scripts.

harshitbansal05:
@henrykironde @zhangcandrew don't know why the appveyor tests fail. The travis test fails because in the function `test_reset_retriever`, we call
```
dataset = random.choice(rt.dataset_names()['offline'])
rt.reset_retriever(dataset)
``` 
and it accidentally deletes one of the test dataset. Thus `rt.get_script_upstream(dataset)` won't work as that dataset is not present upstream.
I ran the tests locally and they ran just fine.

harshitbansal05:
@zhangcandrew in the latest commit, I tried to do `retriever ls` using different colors for online and offline scripts. Also, I think the reason for Travis tests failing is that with this pr, scripts would no longer be downloaded while installation. So, when running tests, the datasets in `test/raw_data_gis/scripts/` are copied to `~/.retriever`. Now, when `test_reset_retriever` is run, it has to delete a random offline script, which has to be one of these test scripts (no script apart from these scripts is available locally). After deletion, it attempts to download the script from upstream, and since the script is not available upstream, the test fails. @henrykironde.

harshitbansal05:
@henrykironde I edited the `test_reset_retriever` test function to never reset a test script, and the tests pass.

zhangcandrew:
Thanks @harshitbansal05, just wanted to throw the recent addition to this discussion here #1290, about pushing downloading scripts when specifically needed down.

harshitbansal05:
@zhangcandrew sorry, but I didn't quite understand. Could you please explain?

zhangcandrew:
Just wanted to reference the issue so we have Ethan's comment in this PR if we look back on this later on. 
 

> Given the length of the To Do list I thought I'd mention that I think that "Restructuring code to not redownload everything unless needed" might be something to consider placing lower down on the priority list. Other package managers (e.g., apt on Debian/Ubuntu) require an update step that pulls down up-to-date information on packages locally and takes a while to run (certainly longer than our updates). We already only update changed scripts, so the only time there's a meaningful download is on the initial setup or after a reset. For most users this will be very infrequent. Given that I think this is the lowest priority on the list. That said, this is more of an outsider's take, so whatever @zhangcandrew @henrykironde & @harshitbansal05 think is best is good with me.

setup.py
zhangcandrew: Hey Harshit, I think this was what was mentioned by Ethan. So he envisioned us still downloading everything since we have a relatively small scripts size still. On first install, everything should get installed, so lets reverse this.

retriever/lib/install.py
zhangcandrew: Same as above, we can just leave it as downloading everything for now.

retriever/lib/download.py
zhangcandrew: same as above.

retriever/lib/datasets.py
zhangcandrew: This is cool :^)

test/test_retriever.py
```diff
@@ -681,13 +688,22 @@ def test_reset_retriever(tmpdir):
...
+     'test-raster-multi-bands', 'test-raster-single-band', 'test-us-eco', 'test-vector']
+    offline_datasets = [dataset for dataset in offline_datasets if dataset not in test_datasets]
+    if len(offline_datasets) == 0:
+        return
+    dataset = random.choice(offline_datasets)
     rt.reset_retriever(dataset)
     rt.reload_scripts()
-    assert dataset not in rt.dataset_names()
-    rt.check_for_updates()
+    assert dataset not in rt.dataset_names()['offline']
```
henrykironde: Looks like we can make two test out of this, online and offline. In most cases if the test code goes past 10 line, try to refactor it, unless otherwise

zhangcandrew:
This is some good stuff! It seems to be working well. Some things:

1. Can you add a blurb of text either above or below ls referencing the retriever-recipes repo somehow (something like to see a full list of available scripts, visit the recipes-repo with link). Also, adding some mention about what the new red color means would help (red means we could not find the script locally, but they are available upstream or something). 

2. I am getting inconsistent results sometimes, where a reset will make some scripts appear red and others not. These commands were all run outside the git repo, cd-ing into a repo will change the results. I think this is a matter of hard coding somewhere, and likely needs to be changed. We should be checking ./retriever/scripts, retriever/scripts, and retriever-recipes/scripts at this point. It should output the same results regardless of our cwd. 

In this scenario, it does not work as expected. We reset predicts then ls and predicts is shown offline (even though it's in the retriever repo). 
image


In this scenario, it appears as expected. We reset some scripts yet they are still online because they are in the retriever-recipes repo. 
image

3. This is likely suited for another PR (and probably in retriever-recipes), but as a feature request, it would be good to add a command to show where the script is located (In the ./retriever/scripts cache, retriever/scripts, or retriever-recipes/scripts).

test/test_retriever.py
```diff
@@ -681,13 +688,22 @@ def test_reset_retriever(tmpdir):
     pwd_name = os.getcwd()
     workdir = tmpdir.mkdtemp()
     workdir.chdir()
-    dataset = random.choice(rt.dataset_names())
+    offline_datasets = rt.dataset_names()['offline']
+    test_datasets = ['test-eco-level-four', 'test-raster-bio1', 'test-raster-bio2',
+     'test-raster-multi-bands', 'test-raster-single-band', 'test-us-eco', 'test-vector']
+    offline_datasets = [dataset for dataset in offline_datasets if dataset not in test_datasets]
```
henrykironde: > I think this happened, becausetest_reset_retriever picks a random dataset from the offline scripts and deletes it. In the above case, that random dataset happened to be test-raster-bio2, which is not available upstream. Thus, after deletion it could not be downloaded again, causing the tests to fail.

@harshitbansal05, how about 
```python
offline_datasets = [dataset for dataset in offline_datasets if not dataset.startswith("test-")]
```

setup.py
harshitbansal05: ok, would do so.

test/test_retriever.py
harshitbansal05: yes, this would be good. Thanks @henrykironde.

harshitbansal05:
@zhangcandrew I worked on the comments you mentioned above and also made changes to consider the recipes path, which we wrote (the other pr is open in the recipes repo). Could you please review it?
@henrykironde ^.

retriever/__main__.py
```diff
@@ -109,21 +109,36 @@ def main():
...
+                    all_scripts_combined.append((dataset, True))
+                for dataset in all_scripts['online']:
+                    if dataset in all_scripts['offline']:
+                        continue
+                    all_scripts_combined.append((dataset, False))
+                all_scripts_combined = sorted(all_scripts_combined, key = lambda x: x[0])
+                print("Available datasets : {}\n".format(len(all_scripts_combined)))
+                lscolumns.printls(all_scripts_combined)
+                print("\nred: online datasets")
+                print("To see the full list of available datasets, visit "
```
henrykironde: Do we need the red?

retriever/__main__.py
```diff
@@ -140,7 +155,20 @@ def main():
...
                     count += 1
-
+                count = 1
+                if not args.v:
+                    print("Online datasets : {}\n".format(len(online_scripts)))
+                offline_scripts = [script.name for script in all_scripts]
+                for script in online_scripts:
+                    if script in offline_scripts:
+                        continue
+                    print(
```
henrykironde: I think this can fit on a single line

retriever/__main__.py
harshitbansal05: The word `red`?

retriever/__main__.py
harshitbansal05: ok, noted.

retriever/__main__.py
zhangcandrew: @henrykironde This was added in to show which scripts were "offline" when you retriever ls. I think it will make a bit more sense when @harshitbansal05 adds in some other keys and comments for the ls. Basically we want the user to be able to know which scripts they have locally and which ones they do not.

test/test_retriever.py
```diff
@@ -681,13 +688,20 @@ def test_reset_retriever(tmpdir):
     pwd_name = os.getcwd()
     workdir = tmpdir.mkdtemp()
     workdir.chdir()
-    dataset = random.choice(rt.dataset_names())
+    offline_datasets = rt.dataset_names()['offline']
+    offline_datasets = [dataset for dataset in offline_datasets if not dataset.startswith('test-')]
+    if len(offline_datasets) == 0:
```
henrykironde: `if not offline_datasets` should be the same as `if len(offline_datasets) == 0:`

harshitbansal05:
@henrykironde done.

PR #1362, opened 2019-07-19
Prompt user if a newer version of script is available

harshitbansal05:
@zhangcandrew @henrykironde, the pr that always attempts to install the latest version of scripts.

### DIFF
--- a/retriever/lib/engine_tools.py
+++ b/retriever/lib/engine_tools.py
@@ -10,7 +10,6 @@ from future import standard_library
 
 standard_library.install_aliases()
 
-import difflib
 import json
 import platform
 import shutil
@@ -19,7 +18,7 @@ import warnings
 
 from hashlib import md5
 from io import StringIO as NewFile
-from retriever.lib.defaults import HOME_DIR, ENCODING, RETRIEVER_REPOSITORY, RETRIEVER_DATASETS
+from retriever.lib.defaults import HOME_DIR, ENCODING
 
 from retriever.lib.models import *
 import xml.etree.ElementTree as ET
@@ -52,56 +51,6 @@ def create_home_dir():
                 print("The Retriever lacks permission to "
                       "access the ~/.retriever/ directory.")
                 raise
-
-
-def name_matches(scripts, arg):
-    """Check for a match of the script in available scripts
-
-    if all, return the entire script list
-    if the exact script is available, return that script
-    if no exact script name detected, match the argument with keywords
-    title and name of all scripts and return the closest matches
-    """
-
-    if not arg:
-        raise ValueError("No dataset name specified")
-    if arg.endswith('.zip'):
-        from retriever.lib.provenance import get_script
-        script = get_script(arg)
-        return [script]
-
-    arg = arg.strip().lower()
-    matches = []
-
-    if arg == 'all':
-        return scripts
-
-    for script in scripts:
-        if arg == script.name.lower():
-            return [script]
-
-    from retriever.lib.scripts import get_script_upstream
-
-    if arg in RETRIEVER_DATASETS:
-        read_script = get_script_upstream(arg, repo=RETRIEVER_REPOSITORY)
-    else:
-        read_script = get_script_upstream(arg)
-
-    if read_script is not None:
-        return [read_script]
-
-    for script in scripts:
-        script_match_ratio = difflib.SequenceMatcher(None, script.name, arg).ratio()
-        if script_match_ratio > .53:
-            matches.append((script.name, script_match_ratio))
-
-    matches.sort(key=lambda x: -x[1])
-
-    print("\nThe dataset \"{}\" "
-          "isn't currently available in the Retriever.".format(arg))
-    if matches:
-        print("Did you mean:"
-              " \n\t{}".format("\n\t".join([i[0] for i in matches])))
 
 
 def final_cleanup(engine):
--- a/retriever/lib/scripts.py
+++ b/retriever/lib/scripts.py
@@ -7,13 +7,16 @@ import csv
 import imp
 import io
 import os
+import re
 import sys
 import json
 import requests
+import difflib
 from os.path import join, exists
 from collections import OrderedDict
 
 from pkg_resources import parse_version
+from distutils.version import LooseVersion
 
 from retriever.lib.defaults import REPOSITORY, RETRIEVER_REPOSITORY, SCRIPT_SEARCH_PATHS, VERSION, ENCODING, SCRIPT_WRITE_PATH, RETRIEVER_SCRIPTS, RETRIEVER_DATASETS
 from retriever.lib.load_json import read_json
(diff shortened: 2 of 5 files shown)


harshitbansal05:
@zhangcandrew not sure why the newer builds are failing.

henrykironde:
This Pr has some changes from the previous PR 1358, Let me know if am missing something.

harshitbansal05:
@henrykironde the 2 commits are from the previous pr only. The last commit of this pr adds new changes, which is prompting the user if a newer version of the script is available.

henrykironde:
So do we need to close the other PR?

harshitbansal05:
no @henrykironde, the other pr would download a script if it is not present locally and we need it. This pr would further attempt to always use the latest version of a script. Example, if a script is present locally, this pr would further check upstream if a newer version of that script is available.

henrykironde:
@harshitbansal05 clean this up and we shall be ready to merge

harshitbansal05:
@henrykironde done.

retriever/lib/engine_tools.py
```diff
@@ -19,6 +19,7 @@
 
 from hashlib import md5
 from io import StringIO as NewFile
+from distutils.version import LooseVersion
 from retriever.lib.defaults import HOME_DIR, ENCODING, RETRIEVER_REPOSITORY, RETRIEVER_DATASETS
```
henrykironde: remove ENCODING

retriever/lib/engine_tools.py
```diff
@@ -76,11 +77,34 @@ def name_matches(scripts, arg):
...
-
-    from retriever.lib.scripts import get_script_upstream
+            local_version = script.version
+            if arg in RETRIEVER_DATASETS:
+                upstream_version = get_script_version_upstream(arg, repo=RETRIEVER_REPOSITORY)
+            else:
+                upstream_version = get_script_version_upstream(arg)
+            if upstream_version is None or LooseVersion(local_version) >= LooseVersion(upstream_version):
+                return [script]
+            prompt = "A newer version of {dataset} is available. Would you like to download it? (y/N): ".format(dataset=arg)
```
henrykironde: line to long

retriever/lib/scripts.py
```diff
@@ -150,59 +176,82 @@ def get_script_upstream(dataset, repo=REPOSITORY):
...
-                r = r.json()
-                for index in range(r['total_count']):
-                    script = r['items'][index]['name']
-                    script = '.'.join(script.split('.')[:-1])
-                    script = script.replace('_', '-')
-                    result_scripts.add(script)
-            except:
-                raise
-    return sorted(result_scripts)
+    try:
```
zhangcandrew: If you want to catch requests exceptions, I think you should catch the base class (requests.exceptions.RequestException). There are a bunch of requests errors that might slip through this check (HTTPError, TimeoutError). 

Now that we are switching over to calls and searches to things upstream, I think this is a good idea to add try catches. I think a better design though would be to abstract retriever specific calls in a new function that solely deals with calls upstream and returning information. This way when something goes wrong with requests, we can report it in retriever without having to try/catch every piece of retriever making calls upstream. The new function could be leveraged by get_dataset_names_upstream and also get_script_version_upstream

zhangcandrew:
It is weird that the checks are failing. Can you make the changes and force push a new commit up? I know ethan was having problems with tests earlier that seemed to clear up so you may have caught some bad timing.

henrykironde:
@harshitbansal05 if you rebase with retriever-recipes-dev the tests will pass.

retriever/lib/engine_tools.py
harshitbansal05: @henrykironde sorry, but ENCODING is used in the file.

retriever/lib/scripts.py
harshitbansal05: @zhangcandrew sorry, your idea is great but could we please do this later in a separate pr?

retriever/lib/engine_tools.py
```diff
@@ -76,11 +77,35 @@ def name_matches(scripts, arg):
...
+                upstream_version = get_script_version_upstream(arg)
+            if upstream_version is None or LooseVersion(local_version) >= LooseVersion(upstream_version):
+                return [script]
+            prompt = "A newer version of {dataset} is available. Would you like to download " \
+                     "it? (y/N): ".format(dataset=arg)
+            should_download = input(prompt)
+            while not (should_download.lower() in ['y', 'n', '']):
+                print("Please enter either y or n.")
+                should_download = input()
+            if should_download.lower() == 'y':
```
henrykironde:
```python
            if should_download.lower() == 'y':
                if arg in RETRIEVER_DATASETS:
                    read_script = get_script_upstream(arg, repo=RETRIEVER_REPOSITORY)
                else:
                    read_script = get_script_upstream(arg)
                if read_script is None:
                    print("Unable to download {dataset}.".format(dataset=arg))
                    return [script]
                else:
                    return [read_script]
            else:
                return [script]
```
Change  how return in if statements is done

```python
                if read_script is None:
                    print("Unable to download {dataset}.".format(dataset=arg))
                    return [script]
                return [read_script]
```

harshitbansal05:
@henrykironde done.

retriever/lib/scripts.py
```diff
@@ -150,59 +174,84 @@ def get_script_upstream(dataset, repo=REPOSITORY):
...
+                script = '.'.join(script.split('.')[:-1])
+                script = script.replace('_', '-')
+                scripts.append(script)
+                if len(scripts) == max_scripts:
+                    break
+            return sorted(scripts)
+
+        result_scripts = set()
+        if repo == RETRIEVER_REPOSITORY:
+            search_url = "https://api.github.com/search/code?q={query}+in:file" \
```
henrykironde:
```python
search_base_url = "https://api.github.com/search/code?q={query}+in:file" \
                 "+path:scripts+repo:weecology/"
if repo == RETRIEVER_REPOSITORY:
    search_url = search_base_url + "ret...."
else:
    search_url = search_base_url + "retriever-re.."
```

harshitbansal05: thanks, @henrykironde.

retriever/lib/engine_tools.py
```diff
@@ -76,12 +77,34 @@ def name_matches(scripts, arg):
     if arg == 'all':
         return scripts
 
+    from retriever.lib.scripts import get_script_version_upstream, get_script_upstream
```
henrykironde: from retriever.lib.scripts import get_script_version_upstream, get_script_upstream
why do we have the import here?

retriever/lib/scripts.py
zhangcandrew: Sure if you're more comfortable with that. I think you can add in the exception change relatively quick, and we can make a note about tackling this in a different PR.

henrykironde:
@zhangcandrew , This looks good but we have a problem with the cyclic dependency and that is why we are putting the imports here https://github.com/weecology/retriever/pull/1362/files#diff-09b72f49b55e7226a62ae92a8a825f82R80 do you have any solution?

zhangcandrew:
I see, I think we can solve this cyclic dependency better if we move the entire name_matches functions into scripts.py. That should stop it. Any thoughts @henrykironde ?

henrykironde:
Let's try that and see

harshitbansal05:
@henrykironde @zhangcandrew I moved the method. Is it fine now?

zhangcandrew:
Fixes the issue and looks good to me! Merging in

PR #1363, opened 2019-07-19
Use newer Python for testing by updating to Ubuntu 18.04

### LINKED ISSUE #1361: Travis builds are failing
We are using `python==3.5.2` on Travis and `python==3.5.4` on AppVeyor. The AppVeyor builds are passing. 
If we look at the logs on Travis the issue seems to be with `pandas`.
Possibly related to https://github.com/pandas-dev/pandas/issues/27247

ethanwhite:
Due to Pandas no longer supporting older Python versions the Travis builds are now failing.
This updates the base image to Ubuntu 18.04, which brings Python 3 up to 3.6.8 where everything
works as expected.

This requires some additional changes to allow the `apt` package `tzdata` to be installed without user interaction. 

Fixes #1361.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,13 @@
-# Download base image ubuntu 16.04
-FROM ubuntu:16.04
+# Download base image ubuntu 18.04
+FROM ubuntu:18.04
 
 MAINTAINER Weecology "https://github.com/weecology/retriever"
 
 RUN apt-get update && apt-get install -y --no-install-recommends apt-utils
+
+# Manually install tzdata to allow for non-interactive install
+RUN apt-get install -y --force-yes tzdata
+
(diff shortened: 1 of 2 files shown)


ethanwhite:
I'm not sure if this is the right approach to the issue that @apoorvaeternity opened, but it was the most straightforward I could think of and it has the added benefit of getting us testing on the most recent Ubuntu long-term service release.

henrykironde:
Thanks @ethanwhite for the fix

apoorvaeternity:
@ethanwhite @henrykironde I think we should also update the python version to `3.6.8` on AppVeyor so that the only difference between the two builds is of the platform.

ethanwhite:
+1 to updating appveyor as well.

PR #1364, opened 2019-07-20
Minor proposals

kant:
* update URL


ethanwhite:
Thanks for the contribution  @kant!

PR #1366, opened 2019-07-20
Hiatus added on paragraph 73 (minor proposal)

kant:
* days.The retriever > days. The retriever


henrykironde:
Thanks you @kant for this PR.

PR #1370, opened 2019-08-04
Add installation of committed datasets stored in provenance directory

apoorvaeternity:
Assumption: If a hash is provided when installing dataset it means it has to be installed from the 
provenance directory.
The hash value of a commit can be obtained using `retriever log dataset_name`

Usage for installation:
```bash
retriever install sqlite abalone-age --hash 02ee77
```
```python
from retriever import install_csv
install_csv('abalone-age', hash='02ee77')
```

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -232,8 +232,9 @@ def main():
         else:
             raise Exception("no dataset specified.")
         if scripts:
-            if args.dataset.endswith('.zip'):
+            if args.dataset.endswith('.zip') or args.hash:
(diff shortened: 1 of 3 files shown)


apoorvaeternity:
@henrykironde Ready for a review.

henrykironde:
This looks good I will merge it, we shall have to update docs and doc strings in the PR to master. 
We should keep all the command samples and we shall include the same as comments when we open the new PR to master. This will enable everyone to test the program easily.

PR #1371, opened 2019-08-09
Allow installation of committed dataset with python script

apoorvaeternity:
This will allow installation of committed datasets which have `python` scripts.
I tried to verify several datasets(including vertnet) and it is working fine.
@henrykironde Can you give it a test at your end.

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -232,7 +232,7 @@ def main():
         else:
             raise Exception("no dataset specified.")
         if scripts:
-            if args.dataset.endswith('.zip') or args.hash:
+            if args.dataset.endswith('.zip') or args.hash_value:
                 _install(vars(args), debug=debug, use_cache=use_cache)
                 return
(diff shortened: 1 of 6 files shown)


apoorvaeternity:
@henrykironde Looks like `flensburg-food-web` has changed causing the build to fail.

henrykironde:
`flensburg-food-web` did not change and it passes on other engines. It looks like there may be changes in the version of sqlite used in the tests

apoorvaeternity:
@henrykironde I seem to have found the issue. Looks like retriever use the same engine instance throughout. So when a committed dataset is installed by setting the path using `data_path` attribute it remains set. So when another dataset uses the same engine it still tries to find the raw data in the `data_path`. So setting the `data_path` attribute back to `None` after installing the committed dataset fixes the issue.

henrykironde:
Do you think we can come up with a test on this error.

PR #1389, opened 2019-09-25
Repair auto detect and fetch upstream scripts

henrykironde:
Repair auto detect
    
Include the size of the type where needed.
Use length of encoded string to account for encoded chars
Do not hard code total number of scripts

### DIFF
--- a/retriever/lib/create_scripts.py
+++ b/retriever/lib/create_scripts.py
@@ -72,12 +72,27 @@ def create_resources(file, skip_lines):
     clean_table = table.__dict__
     resource_dict = {}
     path_to_table = os.path.basename(clean_table["name"])
-    resource_dict["name"] = os.path.splitext(path_to_table)[0]
+    print("Processing... {file_name}".format(file_name=path_to_table))
+    resource_dict["name"] = os.path.splitext(path_to_table)[0].lower()
+    resource_dict["path"] = path_to_table
     resource_dict["schema"] = {}
-    resource_dict["dialect"] = {}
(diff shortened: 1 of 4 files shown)


retriever/lib/create_scripts.py
```diff
@@ -90,6 +105,7 @@ def create_script_dict(allpacks, path, file, skip_lines):
     allpacks["citation"] = "FILL"
     allpacks["licenses"] = [{"name": "FILL"}]
     allpacks["keywords"] = []
+    allpacks["archived"] = "fill or remove this field if not archived"
```
ethanwhite: I wouldn't be immediately clear on what this field should be. Is it a logical? If so what about "FILL: 'True' if data file is compressed and 'False' if it is not"?

ethanwhite: Alternatively we could set it to False which will be the most common case and then include why/how to switch it in the docs.

ethanwhite:
Looks great. One minor comment for discussion.

retriever/lib/create_scripts.py
henrykironde: This is only used when the dataset is stored as an archive. It takes on the value of the type of archive. ie, `zip`, `gz`, etc. I think we have this in docs somewhere. We can discus this later for more input. Thanks Ethan.

ethanwhite: Sounds good. I'll go ahead and merge this in and we can modify that language later if we come up with something better.

PR #1392, opened 2019-10-22
Fix hash value

apoorvaeternity:
Currently, when using hash value for installation of a committed dataset it results in error due to the hash value being read incorrectly. This is a fix for that.

Regression to #1379


henrykironde:
@apoorvaeternity thanks for this

PR #1394, opened 2019-10-28
Return log values in commit_log

henrykironde:
commit_log values are used in other projects


henrykironde:
@apoorvaeternity, we added the commit return to enable using the values in other tools.

ethanwhite:
LGTM!

PR #1398, opened 2019-11-21
[wip]: Add an attribute for the total number of records

henrykironde:
If the total number of records is known, the counting
of the number of line is skipped

### DIFF
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -88,17 +88,25 @@ class Engine(object):
     def add_to_table(self, data_source):
         """Adds data to a table from one or more lines specified
         in engine.table.source."""
+
+        # if the number of records are known avoid counting
+        real_line_length = None
+        if self.table.number_of_records:
+            real_line_length = self.table.number_of_records
+
         if self.table.columns[-1][1][0][:3] == "ct-":
             # cross-tab data
-            real_line_length = self.get_ct_line_length(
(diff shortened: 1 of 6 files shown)


retriever/lib/engine.py
```diff
@@ -951,6 +969,20 @@ def extract_fixed_width(self, line):
...
+    decrement = True
+    while decrement:
+        decrement = False
+        try:
+            csv.field_size_limit(maxInt)
+        except OverflowError:
+            maxInt = int(maxInt / 10)
+            decrement = True
+    return maxInt
+
```
ethanwhite: I found this function difficult to understand. Can we: 1) add a bit more documentation; and 2) see if there is a more intuitive way to write this. The following feels a bit more intuitive to me since we set `decrement` to `True` and only change it when `csv.field_size_limit` works, updating the value used in the loop when it errors.

```python
def set_csv_field_size():
    """Set the CSV size limit based on the available resources"""
    maxInt = sys.maxsize
    decrement = True
    while decrement:
        try:
            csv.field_size_limit(maxInt)
            decrement = False
        except OverflowError:
            maxInt = int(maxInt / 10)
    return maxInt
```

coolalexzb:
Is there any more change needed in this PR? I resolved conflicts after merging master branch and found that the code in master branch is almost right. @henrykironde

PR #1411, opened 2020-01-25
Converting JSON to CSV. Closes #1334

### LINKED ISSUE #1334: Allow consuming JSON data
Currently we only support ingesting delimited tabular data. It is increasingly common for tabular style data to be distributed in JSON files and it would be nice to also be able to consume this. We would probably just convert it to CSV as a starting point and then process it using our standard pipeline.

There are a few not particularly active packages for doing this, but the code to do it is so simple enough that since none of the packages seem to be widely adopted we might be better off just writing and maintaining it ourselves.

(no rush on this, just a thought while looking at a cool dataset that's only available in JSON: http://data.unpaywall.org/products/snapshot)

veb7vmehra:


henrykironde:
This Pr was checked out from your master branch which had other old changes, you should always use a different branch not master. 
Always use the master as the initial branch off point.
That means before creating a PR, update master (which is usually the origin) with the master from weecology/retriever. I will update the docs and show you my git config file

veb7vmehra:
@henrykironde So should I delete this PR  and add a new PR for now sir. And also sorry for inconvenience.

henrykironde:
Just create a new pr, leave this one, I will close it later

ethanwhite:
Replaced by #1475

PR #1414, opened 2020-01-27
Converting JSON files to CSV. Closes #1334

### LINKED ISSUE #1334: Allow consuming JSON data

veb7vmehra:


veb7vmehra:
@henrykironde I was going through the log and I noticed that some input data sets like "simple_csv_simple_csv" is nowhere defined in **test_integration.py** . Can you kindly double-check weather specified data set is correct or not. 
Screenshot (207)

henrykironde:
You are using `from flatten_json import flatten` but I do not see the package being installed.
Where is `flatten_json` from

veb7vmehra:
@henrykironde Okay sir I am done with "ModuleNotFound error" but the previous issue is still prevalent. Also can you suggest me any easy manner to run these test on my machine without pushing it.

henrykironde:
You do not have worry about pushing, that is okey.
Otherwise for `Json` related tests you can run` pytest -k "json2csv" -v . This will run tests related to the match"json2csv"

veb7vmehra:
@henrykironde Thanks for the suggestion. Also can you kindly give your insight on that input issue mentioned above.

henrykironde:
what did you get when you run` pytest -k "json2csv" -v`

veb7vmehra:
@henrykironde Actually it isn't working properly on my machine. I was trying to make it work for some time but firstly it gives the following error:
Screenshot (208)
This gives me the clue that the test case is accessing the wrong path so I uninstalled the library and then tried to run the test and then this error occurred:
Screenshot (209)
I guess we can focus on the log for the time being if you say so.

henrykironde:
That may be because of windows 32 achitecture

veb7vmehra:
@henrykironde I am done with that thing but this issue is still in the log so please guide ahead.

> @henrykironde I was going through the log and I noticed that some input data sets like "simple_csv_simple_csv" is nowhere defined in **test_integration.py** . Can you kindly double-check weather specified data set is correct or not.
> Screenshot (207)
> 
> Also sir there is another error of ModuleNotFound which just arrived in my log and I have no exact clue of how to resolve that so can you please help me with that?

ethanwhite:
Replaced by #1475. Thanks for all your work on this @veb7vmehra!

Issue #1407, opened 2020-01-19
Add support for Kaggle Datasets


DumbMachine:
Kaggle is home to numerous datasets and pre-trained models. Adding support to easily use the kaggle API would be beneficial to download and upload dataset. 
To interact with Kaggle API, user requires two things:
- Kaggle account
- Kaggle access_token stored at ~/.kaggle/kaggle.json or the access_token can be exported to environment `export KAGGLE_USERNAME=datadinosaur
export KAGGLE_KEY=xxxxxxxxxxxxxx`.

Once the access of above is ensured, the following commands can be used 
```
kaggle competitions {list, files, download, submit, submissions, leaderboard}
kaggle datasets {list, files, download, create, version, init}
kaggle kernels {list, init, push, pull, output, status}
kaggle config {view, set, unset}
```
We only have to take care of the `competitions & datasets` commands for inclusion into retriever, as those commands are used to download Kaggle Competition Datasets and other Datasets uploaded by kaggle users. 

To download the dataset, we would have to call the kaggle cli. This could be done with ease: 
```python
from subprocess import Popen, PIPE

process = Popen(['kaggle',  'datasets', 'download', '-d',  'alessiocorrado99/animals10'], stdout=PIPE, stderr=PIPE)
stdout, stderr = process.communicate()
```
We won't be able to provide support for all database systems as many files on kaggle can be images/videos/other formats (like pretrained vecters). But this shouldn't be a problem concerning the initial release.

TLDR; we check for Kaggle API credentials then execute appropriate kaggle cli commands.

DumbMachine:
Talking about this issue reminds me of a feature I had thought of last GSOC season. Dataset's sizes have been on the rise, usage of cloud services are common for training/inference/analysis. What do you think about a feature, which would help users choose their favored cloud provider and we provide a means to easily fetch and update datasets? This might enable people to use customs datasets and upload/commit changes to dataset and their model  @ethanwhite @henrykironde

henrykironde:
I may be wrong but I think we do have what you mentioned in the `Python` interface but not in the `Command line`. Users can create, fetch datasets from their own repositories using `get_script_upstream` but we have not tested this with any other repo other than the `retriever-recipes repo`. You could add an issue and we improve this.

ethanwhite:
[...]

PR #1426, opened 2020-02-22
[WIP]  Add support for Kaggle Datasets

DumbMachine:
This pull request is for catering to this [issue](https://github.com/weecology/retriever/issues/1407). 
For achieving this functionality I propose the following:
- Instead of adding a field, `source`, to the json files, we can modify the url field of each such json file.
For example, consider the following cases:
- Competition data: The data requested by user is that of  kaggle completion. So the url will be: `kaggle:comptetion:digit-recognizer`.
- User Datset data: The data requested by user is that of  kaggle dataset. So the url will be: `kaggle:dataset:jessemostipak/hotel-booking-demand`.

The advantage of this approach is that if we want to add more specific sources of datasets, something similar to kaggle, catering to will be very easy as we will be following the generalised approach of `<source_name>:<extra_information>:<dataset_identifier>`.


henrykironde:
@DumbMachine, the url seems fine since the api uses that format. In the case of `kaggle`, retriever should know that this is a `kaggle `data package and should treat it differently.

DumbMachine:
The `if` condition checks for the type of data package to be used for downloading.
https://github.com/weecology/retriever/blob/1c3cb4479aea60a498b97813e304be7ae2e20ebd/retriever/lib/engine.py#L476

In the future adding support for other data packages will be as simple as adding elif conditions:
```python
elif if <package_name> == url.split(":")[0]:
```

henrykironde:
That may work now but I am not sure it is the optimal solution for the future. We may find that out later as we test all the components related to scripts. 
For compressed files `Retriever recipes` has a tag that we use to specify the the type of compression , i.e `zip`, `gz`, `tar` `https://github.com/weecology/retriever-recipes/search?q=archived&type=Code` 
Then use the engine's download and download_files_from_archive to decompress that file

DumbMachine:
@henrykironde I did not understand the problem that you have mentioned, all the kaggle related code is in download_files_from_archive. This is because all the data downloaded from kaggle is served as `.zip` file. 
I have tested the above code and the feature works as expected. Commands used for testing were (I have made a pr for the corresponding recipes):
```bash
# For datasets
retriever install csv iris_kaggle
# For comptetitions
retriever install csv titanic
```

DumbMachine:
@henrykironde @ethanwhite I have a few questions:
1. Should this PR contain the tests? 
2. How do we plan to pass the KaggleApi `access_token`? Would a simple json work? I believe we can pass them as environment variables to travis ci as :
```yaml
    environment:
      # Handle tzdata install see: https://serverfault.com/a/975084
      TZ: America/New_York
      DEBIAN_FRONTEND: noninteractive
      ...
      KAGGLE_USERNAME=datadinosaur
      KAGGLE_KEY=xxxxxxxxxxxxxx
```

retriever/lib/engine.py
```diff
@@ -521,34 +551,56 @@ def download_files_from_archive(
...
                 os.makedirs(archive_dir)
 
-        if not file_names:
-            self.download_file(url, archive_name)
-            if archive_type in ('tar', 'tar.gz'):
-                file_names = self.extract_tar(archive_full_path, archive_dir,
-                                              archive_type)
-            elif archive_type == 'zip':
+        if url.split(":")[0] ==  "kaggle":
+            from kaggle.api.kaggle_api_extended import KaggleApi
```
henrykironde: This kaggle part can be a static function like download_kaggle()

retriever/lib/Retriever.code-workspace
```diff
@@ -0,0 +1,11 @@
+{
+	"folders": [
+		{
```
henrykironde: Why do we need this file?

henrykironde:
We shall have tests included and they should be designed as you proposed. 
We shall set Kaggle specific ENVs in GItHub
```yaml
  environment:
      # Handle tzdata install see: https://serverfault.com/a/975084
      KAGGLE_TZ: America/New_York
      KAGGLE_DEBIAN_FRONTEND: noninteractive
      ...
      KAGGLE_USERNAME=datadinosaur
      KAGGLE_KEY=xxxxxxxxxxxxxx
```
and In the test we can skip this if the ENV are not present
So if I set These locally, I should be able test them.

```python
# unit test
# test_retriever
def test_download_from_Kagle():
    IF KAGGLE_USERNAME or KAGGLE_KEY or Whatever env:
        Test download function 

```
Later we can add regression tests:
We shall add the script from Kaggle as on of those scripts to be processed in test_regression and calculate it's hash

retriever/lib/engine.py
```diff
@@ -525,6 +495,35 @@ def download_file(self, url, filename):
...
 
+    def download_from_kaggle(
+        self,
+        data_source,
+        dataset_name,
+        archive_dir,
+        archive_full_path,
+    ):
+        """Download files from Kaggle into the raw data directory"""
+        from kaggle.api.kaggle_api_extended import KaggleApi
```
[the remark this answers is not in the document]
apoorvaeternity: @DumbMachine I think we can put this on top if it doesn't break anything since we have already added `kaggle` in `requirements.txt`.

retriever/lib/engine.py
```diff
@@ -481,7 +480,7 @@ def download_file(self, url, filename):
                 requests.get(
                     url,
                     allow_redirects=True,
-                    stream=True,
+                    s0tream=True,
```
apoorvaeternity: @DumbMachine This change seems unintended.

DumbMachine:
@henrykironde the tests are failing on the CI as we have not yet set the ENV vars in the build definition. Should I use my access keys to pass the tests or will we create new ones for retriever? (as in a kaggle account for retriever)

henrykironde:
Please follow this style where you first test for presence of the Env if it is there and set

```python
# unit test
# test_retriever
def test_download_from_Kagle():
    IF KAGGLE_USERNAME or KAGGLE_KEY or Whatever env:
        Test download function 
```

henrykironde:
We need to avoid checking the url, it is not a good approach. we shall need a key word just like  "retriever": "true" to identify Kaggle data packages

DumbMachine:
> We need to avoid checking the url, it is not a good approach. we shall need a key word just like "retriever": "true" to identify Kaggle data packages

Ok, that makes sense. So each kaggle related recipe must have `"kaggle": "true"` ?

retriever/lib/engine.py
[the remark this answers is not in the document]
DumbMachine: Ok, I will make the change

test/test_retriever.py
```diff
@@ -215,6 +215,42 @@ def test_drop_statement():
...
+    files = test_engine.download_from_kaggle(
+        data_source="dataset",
+        dataset_name="uciml/iris",
+        archive_dir=raw_dir_files,
+        archive_full_path=os.path.join(raw_dir_files, "iris")
+    )
+    assert ['Iris.csv', 'database.sqlite'] == files
+
+def test_download_from_kaggle_unknown():
+    """Test the downloading of dataset from kaggle, of a erranous dataset name"""
```
apoorvaeternity: `"""Test the downloading of a dataset from kaggle using an erroneous dataset name"""`

test/test_retriever.py
```diff
@@ -215,6 +215,42 @@ def test_drop_statement():
...
+    setup_functions()
+    try:
+        files = test_engine.download_from_kaggle(
+            data_source="competition",
+            dataset_name="titanic",
+            archive_dir=raw_dir_files,
+            archive_full_path=os.path.join(raw_dir_files, "non_existant_dataset")
```
apoorvaeternity: `non_existent_dataset`

retriever/lib/install.py
```diff
@@ -64,6 +63,7 @@ def install_csv(dataset,
         'force': force,
         'hash_value': hash_value
     }
+    print("thea args are", args)
```
apoorvaeternity: I don't think this `print` statement is required.

DumbMachine:
> We need to avoid checking the url, it is not a good approach. we shall need a key word just like "retriever": "true" to identify Kaggle data packages

@henrykironde could you tell which function will I have to make the changes in to allow for this?

retriever/lib/engine.py
```diff
@@ -521,34 +550,44 @@ def download_files_from_archive(
...
-            self.download_file(url, archive_name)
-            if archive_type in ('tar', 'tar.gz'):
-                file_names = self.extract_tar(archive_full_path, archive_dir,
-                                              archive_type)
-            elif archive_type == 'zip':
-                file_names = self.extract_zip(archive_full_path, archive_dir)
-            elif archive_type == 'gz':
-                file_names = self.extract_gz(archive_full_path, archive_dir)
-            return file_names
+        if url.split(":")[0] ==  "kaggle":
```
henrykironde: The engine has an attribute of script. 
From which we get all the values like url and paths used in the downloading function.
We can check if that attribute of script has `kaggle` and if `kaggle is true`

DumbMachine:
@henrykironde I propose a change that we should remove `kaggle ` from the requirements.txt as on every 1st import kaggle looks for the `access tokens`. On failure to find it will raise an error.
Let make the KaggleApi a conditional import, so that it is imported only when request for a kaggle data is made. If the module is then not found, we could handle this with an informative error message.

DumbMachine:
@henrykironde The kaggle related tests have passed and I have made the change to check `kaggle` field of `<recipe>.json`

henrykironde:
We shall check it tomorrow. EST time

retriever/lib/engine_tools.py
```diff
@@ -131,7 +131,6 @@ def json2csv(input_file, output_file=None, header_values=None, encoding=ENCODING
     subprocess.call(['rm', '-r', input_file])
     return output_file
 
-
```
henrykironde: _Deleted lines_
I do  not recommend changing this here. 
If there are some lines that need to be removed, we can create a new PR separate from this since the changes are not closely coupled to this Kaggle PR

DumbMachine:
I'm working on this now.

retriever/lib/engine_tools.py
DumbMachine: I had a vscode extension running that deleted excess spaces. I have removed the changes

DumbMachine:
I have added the new tests, updated `force` kaggle download each time and checked if kaggle stores any dataset related data in cache (didn't find any).
Following is a screenshot of testing passing locally.
image

DumbMachine:
@henrykironde let me know if anything is required here.

test/test_retriever.py
```diff
@@ -52,6 +53,12 @@
 tar_gz_url = os.path.normpath(achive_url.format(file_path='sample_tar.tar.gz'))
 gz_url = os.path.normpath(achive_url.format(file_path='sample.gz'))
 
+kaggle_datasets = [
+    # test_name, data_source, dataset_name, dataset_name, repath, expected
```
henrykironde: `dataset_name, dataset_name`
We should change the names here. Both can't be the same.

henrykironde:
@DumbMachine lets add kaggle to https://github.com/weecology/retriever/blob/master/Dockerfile#L39

DumbMachine:
@henrykironde We also have to add in a `kaggle.json` for the kaggle access_token. What do you want done here?

DumbMachine:
travis seems to fail because of postgis and appveyor is failing because of the lack of access_token (kaggle.json) in the container. To pass these tests, I can add condition to first check for access_token   and skip the test if keys are not present.

Issue #1403, opened 2020-01-15
Improve retriever autocreate 


henrykironde:
Some features of `retriever autocreate` do need some improvements

I have some changes already, feel free to take over and  create a new PR with the needed changes 
ref : https://github.com/henrykironde/retriever/pull/127/files

henrykironde:
More issues discovered:

- The json file created should have `_` and not `-`.

- The resource `name`  (`resource_dict["name"]`) should be cleaned up and no "dots", space and not `-`

pathak-mayurdeep:
I have cleaned up the resource name and path. 
But regarding other issue with json file name with `-`, I'm little confused. Should I clean up the file name at `def write_out_scripts` or did you meant something else? @henrykironde

pathak-mayurdeep:
@henrykironde I added a line that should remove any `white-space` and `-` in the json file name.
I am not sure if this is what you meant.

henrykironde:
@pathak-mayurdeep, create a PR and I will test it.

pathak-mayurdeep:
Ok... I have created a clean branch from master and corrected all the changes. I am creating another PR.

PR #1427, opened 2020-02-24
Improve retriever autocreate #1403

coolalexzb:
I replace dot, -, space with underscore, generating a new resource_dict["name"]. I also replace all the consecutive underscores with single underscore after the first replacement.  If Ido not fully understand your meaning, please let me know. 
Besides, I want to know if I have method to test locally. Thank you!


henrykironde:
Looks fine @coolalexzb thanks

PR #1430, opened 2020-02-25
Update PostgreSQL config file settings and path.

henrykironde:
PostgreSQL new versions are moving from `.pgpass` to `pgpass.conf`
Windows default location in `AppData\Roaming\postgres`


henrykironde:
Could any Windows user confirm the Postgres pgpass setup

PR #1432, opened 2020-02-27
 Improve output when connection is not made #1281 

ashishpriyadarshiCIC:
Solves issue [#1247](https://github.com/weecology/retriever/issues/1247)


ashishpriyadarshiCIC:
@ethanwhite @henrykironde Can you please take a look at this PR?

henrykironde:
@ashishpriyadarshiCIC thanks for these changes. Could you briefly tell me how you tested these changes or paste the commands you used?
It would be helpful if you paste the output here.

ashishpriyadarshiCIC:
Here it is :-
```
ashish@ashish-HP-ENVY-15-Notebook-PC:~$ retriever install mysql iris 
Connection with MySQL Failed! Check file ~/.my.cnf or try to restart MySQl Server
Installation Failed
ashish@ashish-HP-ENVY-15-Notebook-PC:~$ 
```

retriever/engines/mysql.py
```diff
@@ -122,6 +120,11 @@ def get_connection(self):
 
         args['client_flag'] = client.LOCAL_FILES
         self.get_input()
-        return dbapi.connect(charset=self.lookup_encoding(),
-                             read_default_file='~/.my.cnf',
-                             **args)
+        try:
+            return dbapi.connect(charset=self.lookup_encoding(),
+                             read_default_file='~/.my.cnf', **args)
+        except dbapi.err.OperationalError as _:
```
apoorvaeternity: I think there is no need of `as _` in the above statement. @henrykironde do we also want to print the actual error?

retriever/engines/postgres.py
```diff
@@ -258,13 +258,19 @@ def get_connection(self):
         import psycopg2 as dbapi
 
         self.get_input()
-        conn = dbapi.connect(
+        try:
+        	conn = dbapi.connect(
```
apoorvaeternity: Indentation seems a bit off at some places.

retriever/engines/mysql.py
ashishpriyadarshiCIC: @apoorvaeternity @henrykironde  Should I remove the first exception?

retriever/engines/postgres.py
ashishpriyadarshiCIC: Fixed it

retriever/engines/postgres.py
```diff
@@ -258,13 +258,19 @@ def get_connection(self):
...
+            conn = dbapi.connect(
             host=self.opts["host"],
             port=int(self.opts["port"]),
             user=self.opts["user"],
             password=self.opts["password"],
-            database=self.opts["database"],
-        )
+            database=self.opts["database"],)
+        except dbapi.OperationalError as _: 
+            print("Connection with PostgreSQL Failed! Check file ~/.pgpass or try to restart PostgreSQl Server")
```
henrykironde: I think we should not print out our own message. dbapi.OperationalError will provide the error message
I think something like this would be enough/(same can apply for mysql)
     
```python
 except dbapi.err.OperationalError as error:
            print(error)
```

henrykironde:
Thanks @ashishpriyadarshiCIC this looks good to me .

henrykironde:
It should be one commit

ashishpriyadarshiCIC:
@henrykironde  I squashed my last 9 commits but what should I do with the old commits from last year?

henrykironde:
Thanks @ashishpriyadarshiCIC for this I will squash the other commits here.

Issue #1434, opened 2020-02-27
Add bump version


henrykironde:
When we make a release, bump version should automatically update the places where we should up date the version numbers

This is a new feature and easy to add if you look at some examples online.

Places to update :https://github.com/weecology/retriever/commit/a82905671aa59407f51ff5aa2264539917fe4d11

ashishpriyadarshiCIC:
@henrykironde  I would like to work on it

henrykironde:
Please go ahead @ashishpriyadarshiCIC

coolalexzb:
I think it easy to change bump version in setup.py, _version.py and version.txt automatically if we follow the creating release steps metioned in developer documentation.

But, how does file retriever_installer.iss generate? Is it generated manually before? If so, is it better to add a new function to update retriever_installer.iss in setup.py?

Another question is that when I want make change in setup.py, it seems that I cannot use the command like `python setup.py 2.4.2` to pass a new version to the __version__ in setup.py.

Thanks for your guide @henrykironde

mauwazahmed:
we can change the major,minor and patch in the version
[bumpversion]
current_version = 0.0.0
commit = True
tag = False
parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+)(?P<build>\d+))?
serialize =
    {major}.{minor}.{patch}-{release}{build}
    {major}.{minor}.{patch}

[bumpversion:part:release]
optional_value = prod
first_value = dev
values =
    dev
    prod

[bumpversion:part:build]

[bumpversion:file:VERSION]

[bumpversion:file:./api/__version__.py]

[bumpversion:file:./api/swagger.yaml]
search = version: {current_version}
replace = {new_version}

mauwazahmed:
we run it as
$ bumpversion major/minor/patch; cat version
and the output results in an increment in the mentioned element

henrykironde:
Thanks @mauwazahmed for the input on this

mauwazahmed:
You're welcomed Henry!! Wish you safety and goodwill.

PR #1435, opened 2020-02-28
adds bump-version script #1434

pathak-mayurdeep:
Wrote a python script to automatically update version numbers.

image

image

image

image

image

### DIFF
--- a/bump-version.py
+++ b/bump-version.py
@@ -1,0 +1,35 @@
+from fileinput import FileInput
+import sys
+import re
+
+if len (sys.argv) != 2 :
+    print("Usage: python bump-version.py 3.2.1")
+    sys.exit (1)
+arg = sys.argv[1]
+
+file_list =[ 
+    "retriever/_version.py",
+    "retriever_installer.iss",
+    "setup.py",
+    
+]
+
+
+with FileInput(files=file_list, inplace=True) as f:
+    for line in f:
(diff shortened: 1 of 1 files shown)


henrykironde:
Thanks @pathak-mayurdeep for the feature, it looks like it is what we want . There is however a tool that handles this, I wonder if you could compare the two and see.
https://medium.com/@williamhayes/versioning-using-bumpversion-4d13c914e9b8
https://pypi.org/project/bumpversion/

pathak-mayurdeep:
I have done some research on this tool, and this is what I found.
It supports addition of git tags by adding `--tag` 
Allows easy increments of parts of the version numbers, where parts can be configured as required.

So we can either go with this tool and try to configure it for retriever, or we could create these features on our own. @henrykironde how should I proceed?

apoorvaeternity:
+1 to using something like https://pypi.org/project/bump2version/ and configuring it for retriever. It's battle tested and is being used by several others. It is also a part of development and won't be used by our end users so I don't see any benefit of reinventing the wheels here.
@pathak-mayurdeep @henrykironde thoughts?

pathak-mayurdeep:
@apoorvaeternity  Yeah, I agree. I'll try configuring it for retriever then. Thanks.

henrykironde:
Closing ref #1460

PR #1436, opened 2020-02-29
#1419 Implement citation lookup in Python interface for new test

coolalexzb:
Hi, I merged commits in #1428 in this pr.

PR #1437, opened 2020-03-02
Update format style in print

coolalexzb:
I updated the print statements mentioned in #1436 
E.g 
`print("\nDataset:  {}".format(dataset.name))`
to 
`print(f"\nDataset:  {dataset.name}")`

### DIFF
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -68,7 +68,7 @@ def main():
             else:
                 citations = get_script_citation(args.dataset)
                 for citation in citations:
-                    print(f"Citation:   {citation}")
+                    print(f"Citation:  {citation}")
             return
 
         if args.command == 'license':
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -68,9 +68,9 @@ def main():
             else:
                 scripts = name_matches(script_list, args.dataset)
                 for dataset in scripts:
-                    print("\nDataset:  {}".format(dataset.name))
-                    print("Citation:   {}".format(dataset.citation))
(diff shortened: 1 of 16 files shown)


henrykironde:
After a review, the best way would be to change every format to `fstring`.
- You can commit each file in a new commit to make it easy for review.
- The script directory can be left out since it may deal with backward compatibility with the old versions and the `retriever-recipes`.

coolalexzb:
Actually, I checked all of the files of the project except files in script directory. All the print statements like `print("\nDataset: {}".format(dataset.name))` have been replaced. Do you mean that print statements like `print("Citation : %s", citation) `and `print("Citation : " + citation)` also need to be repalced?

henrykironde:
Try to find other occurences using `grep -iR "format(" ./retriever`

coolalexzb:
I forgot to pull the latest code before making a change. I will resolve conflicts after the mid term exams on Thursday. Sorry for the late submission.

coolalexzb:
I make a change but it seems that there exists something wrong in travis CI build(permission problem in docker)

coolalexzb:
Sorry for the late resubmit. I did not notice a little bug before.

apoorvaeternity:
@coolalexzb It looks like there is some issue with your local git configuration because when you commit it doesn't point to a GitHub user. I think your commits are missing email.

coolalexzb:
I commited with an email address. Can you check again? Thank you! @apoorvaeternity

apoorvaeternity:
@coolalexzb Yep. The latest commit is correct.

coolalexzb:
Ref #1462

PR #1439, opened 2020-03-03
Retriever auto create spatial datasets

henrykironde:


henrykironde:
This is ready for a review however the last downgrade of docker-compose does not resolve the problem. At this point I will contact the folks at travis to see if they can help

ethanwhite:
This looks good to me (though I don't know much about the gdal stuff). Test are passing after a restart. Should I go ahead and merge?

henrykironde:
Thanks @ethanwhite the gdal related work is for reading spatial raw files. I know were are putting that in a try block, that is because we are not adding gdal to the requirements.
We shall add tests for those helper functions later in the year. A merge would be great

PR #1445, opened 2020-03-11
Repair Postgres bulk insert and check hash_value

henrykironde:


retriever/engines/postgres.py
```diff
@@ -108,20 +108,20 @@ def insert_data_from_file(self, filename):
...
 WITH DELIMITER ','
 CSV HEADER;"""
             try:
                 self.execute("BEGIN")
                 self.execute(statement)
                 self.execute("COMMIT")
+                print("Bulk insert on .. ", self.table_name())
                 return True
-            except BaseException:
+            except Exception as e:
```
apoorvaeternity: @henrykironde I don't think we need to assign the exception to a variable if we are not printing it.

henrykironde:
@apoorvaeternity done

henrykironde:
I have put a clean version #1451

Issue #903, opened 2017-04-27
Gracefully handle failed downloads


ethanwhite:
It is not uncommon for a data source to go down (e.g. #902) or for a download to fail for some reason (e.g., #863). We should catch these, not cache the data that comes down (which is sometimes a corrupt file and sometimes a 404 html page), and report to the user that the source appears to be down and that they should try again and if it still fails later let us know.

kvnamipara:
@ethanwhite Can't we have all our data set file in one place to solve the problem of data source going down. Like we can move all our data set files to service providers like AWS and we try to update it whenever the dataset files got updated?

ethanwhite:
@kvnamipara It would certainly be nice if it was that easy. Unfortunately many of the datasets we work with don't have clear open licences and many others update regularly so we'd be dealing with synchronizing with the central source. The philosophy of the project is to work with data as it is, where it is, instead of cleaning it up and moving it around. This has downsides, but it also lets us work with lots of data that we couldn't otherwise.

ShivamNegi:
The download functionality is being implemented in the `retriever/lib/engine.py` (https://github.com/weecology/retriever/blob/master/retriever/lib/engine.py#L396). The functionality is being implemented using `urllib` library. We could use `urllib2` which has better support for error handling. There is a similar question in stack overflow (http://stackoverflow.com/questions/987876/how-to-know-if-urllib-urlretrieve-succeeds).
About the cases when there is ungraceful program termination like force kill, power failure. I think we will need to put checks before using the cached copies of downloaded scripts.

henrykironde:
@ShivamNegi, we also have consider python 3

ShivamNegi:
Thanks! @henrykironde I missed that. So implementing error handling with `urllib` is the way to go? I am looking for other options.

adhaamehab:
I think [Requests](https://github.com/requests/requests) will make sense here.
@henrykironde @ethanwhite  Should I go with it or you have other thoughts ?

henrykironde:
Thank you @adhaamehab, Let not include more dependencies unless there is no way to solve the problem.

adhaamehab:
Alright, I'll work on this issue this weekend.

adhaamehab:
@henrykironde  @ethanwhite  how should we handle a fault download at 
https://github.com/weecology/retriever/blob/master/retriever/lib/engine.py#L407
should we raise an Error or what?

henrykironde:
[...]

PR #1446, opened 2020-03-14
[WIP]: Catch 404 error

henrykironde:


henrykironde:
_**Things to think about:**_

Some of our tests have fake urls.
The `urlretrieve` also need to check for the same.
 related to #903

PR #1451, opened 2020-03-19
Use mariadb-client instead of mysql client

henrykironde:


henrykironde:
closing with ref: #1468

PR #1457, opened 2020-03-25
Retriever autocreate documentation

ashishpriyadarshiCIC:
Docs didn't have any documentation for `retriever autocreate` so I added some documentation as discussed with @henrykironde and @apoorvaeternity .

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -28,7 +28,33 @@ The ``retriever-recipes`` project can be installed from Github using the followi
 Script Creation
 ---------------
 
-To create a new script, try ``retriever-recipes new_json``, which starts the CLI tool for new script creation.
+To create a new script, there are 2 methods :- 
+
+1. Use retriever autocreate to automatically create a script template. Specify the type of data using -dt, the default data type is tabular. Download the files to a folder. In case of tabular data, the files should be CSV files. Autocreate can create a script template for each file using -f or use -d to create a single script template for all files in the directory.
+
+::
+
+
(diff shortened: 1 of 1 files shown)


docs/recipes.rst
```diff
@@ -28,7 +28,33 @@ The ``retriever-recipes`` project can be installed from Github using the followi
 Script Creation
 ---------------
 
-To create a new script, try ``retriever-recipes new_json``, which starts the CLI tool for new script creation.
+To create a new script, there are 2 methods :- 
+
+1. Automated creation of a clean script using ``retriever autocreate`` followed by the path to the data files. This will create a script in which only the fields from the Basic script will be required except table name.
```
henrykironde: 1. Use retriever autocreate to automatically create a script template. 
Specify the type of data using -dt, the default data type is tabular.
Download the files to a folder. In case of tabular data, the files should be CSV files. 
`Autocreate` can create a script template for each file using `-f` or use -d to create a single script template for all files in the directory.

henrykironde:
Could you clean up the commits. This is something that we went through with you. Let me know if you need more clarification. Practice makes perfect

ashishpriyadarshiCIC:
@henrykironde Sorry, I just fell asleep while working. I just woke up now. I'll clean it up

docs/recipes.rst
ashishpriyadarshiCIC: Made the change

henrykironde: I guess you may not have pushed the new changes, I do not see any change here

ashishpriyadarshiCIC: Yep, I've pushed it now

PR #1459, opened 2020-03-29
Create support for other file formats

mauwazahmed:
the following piece of code attempts to convert sqlite,excel,netcdf,hdf, json and cml formats to csv


henrykironde:
Thanks for this initiative @mauwazahmed. This enable some folks to have an insight of the way forward on this project idea.

PR #1470, opened 2020-04-13
added environment file for conda

asarmakeeva:
to turn on that environment the user should run
`conda env create -f environment.yml`

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,4 @@
+name: retriever
+dependencies:
+  - pip:
+      - git+git://github.com/weecology/retriever.git


ethanwhite:
Thanks for this @asarmakeeva. I think having an `environment.yml` file is definitely a nice addition for setting up local development environments using `conda`. I think for it to be maximally useful we should also include all of the dependencies for the retriever so that they can be installed through automatically through `conda`. The full set of dependencies should be listed in `setup.py`

https://github.com/weecology/retriever/blob/e649be2779695f81a4d30260a0b757001062febe/setup.py#L40-L46

An example of how these can be added is included in another one of our projects:

https://github.com/weecology/DeepForest/blob/master/environment.yml

henrykironde:
Thanks @asarmakeeva for the addition, we shall close this and hopefully try to get back to this issue in the future

PR #1474, opened 2020-05-25
Fix encoding bug on windows.

henrykironde:
Read with encoding when OS  NT(Windows)
Add a sample test with encoding but without defining the delimiter.


henrykironde:
A new commit should be able to confirm the patch and have the tests passing.

henrykironde:
Since we need this urgently I will go ahead and merge it.

PR #1480, opened 2020-06-15
xls2csv

ashishpriyadarshiCIC:
xls to CSV conversion function

### DIFF
--- a/retriever/lib/engine_tools.py
+++ b/retriever/lib/engine_tools.py
@@ -167,7 +167,7 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
 def xlsxcsv(book):
     workbook = xlrd.open_workbook(book)
     dire = book.rstrip(".xlsx")
-    os.makedirs(dire,exist_ok=True)
+    os.makedirs(dire, exist_ok=True)
     os.chdir(dire)
     res = len(workbook.sheet_names())
     for sheet in range(res):
@@ -177,6 +177,7 @@ def xlsxcsv(book):
             df.loc[len(df)] = worksheet.row_values(1)
         table_name = workbook.sheet_names()
         df.to_csv(f"{table_name[sheet]}.csv", index_label='index')
+
 
 def getmd5(data, data_type='lines', encoding='utf-8'):
     """Get MD5 of a data source."""
--- a/retriever/lib/engine_tools.py
+++ b/retriever/lib/engine_tools.py
@@ -17,6 +17,8 @@ from retriever.lib.defaults import HOME_DIR, ENCODING
 import xml.etree.ElementTree as ET
 import os
 import csv
+import xlrd
+import pandas as pd
 
 warnings.filterwarnings("ignore")
 from retriever.lib.tools import open_fr, open_csvw, open_fw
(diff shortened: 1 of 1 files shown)


retriever/lib/engine_tools.py
```diff
@@ -162,6 +164,25 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
     return outputfile
 
 
+def xlsxcsv(book):
+    workbook = xlrd.open_workbook(book)
+    dire = book.rstrip(".xlsx")
+    if not os.path.exists(dire):
+        os.makedirs(dire)
+    else:
```
apoorvaeternity, line 172: Is this `else` required?

retriever/lib/engine_tools.py
```diff
@@ -162,6 +164,25 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
...
+        pass
+    os.chdir(dire)
+    res = len(workbook.sheet_names())
+    for sheet in range(0, res):
```
apoorvaeternity, line 176: I think we can omit the `0` and just use `range(res)`.

retriever/lib/engine_tools.py
```diff
@@ -162,6 +164,25 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
...
+        worksheet = workbook.sheet_by_index(sheet)
+        df = pd.DataFrame(columns=list(worksheet.row_values(0)))
+        for index in range(worksheet.nrows):
+            df.loc[len(df)] = worksheet.row_values(1)
+        table_name = workbook.sheet_names()
+        df.to_csv(table_name[sheet] + '.csv', index_label='index')
```
apoorvaeternity, line 182: We can make use of python `f-strings` here.

retriever/lib/engine_tools.py
```diff
@@ -162,6 +164,25 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
...
+    os.chdir("..")
```
apoorvaeternity, line 183: Is there a cleaner way to do this? I am not sure if this will always be compatible across different OS - it maybe. If we can pass the absolute path in `os.chdir()` somehow that would be great.

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: I was using it just to remove path already exists error, I'll remove it by adding `exist_ok=True`

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: :+1:

apoorvaeternity:
@ashishpriyadarshiCIC Any updates on this?

retriever/lib/engine_tools.py
```diff
@@ -162,6 +164,21 @@ def xml2csv(input_file, outputfile=None, header_values=None, row_tag="row"):
     return outputfile
 
 
+def xlsxcsv(book):
```
apoorvaeternity, line 167: The name should be `xslx2csv` to be consistent with other functions. Also, worth adding some comments to the code here.

ashishpriyadarshiCIC:
Hi @apoorvaeternity, I just shifted all the work on Xlsx to CSV conversion function to #1506

henrykironde:
Ref: #1506

PR #1415, opened 2020-01-28
[WIP] adds support for Frictionless datapackage weecology#1404

pathak-mayurdeep:

PR #1420, opened 2020-01-31
[WIP] Support for Frictionless datapackage

pathak-mayurdeep:
This is a separate branch where I have added some new datapackage to `datapackages.yml` and made some changes to debug the path issue while installing.

Issue #1404, opened 2020-01-16
Add  datasets from data hub awesome collection


henrykironde:
## Collections

* [Bibliographic data](https://datahub.io/awesome/bibliographic-data)

	 *List of datapackages created*
	 - [ ] add

* [Climate Change](https://datahub.io/awesome/climate-change)

	 *List of datapackages created*
	 - [ ] add

* [Demographics (population)](https://datahub.io/awesome/demographics)

	 *List of datapackages created*
	 - [ ] add

* [Economic Data and Indicators](https://datahub.io/awesome/economic-data)

	 *List of datapackages created*
	 - [ ] add

* [Education](https://datahub.io/awesome/education)

	 *List of datapackages created*
	 - [ ] add

* [Football](https://datahub.io/awesome/football)

	 *List of datapackages created*
	 - [ ] add

* [GeoJSON](https://datahub.io/awesome/geojson)

	 *List of datapackages created*
	 - [ ] add

* [Health Care Data](https://datahub.io/awesome/healthcare-data)

	 *List of datapackages created*
	 - [ ] add

* [Inflation](https://datahub.io/awesome/inflation)

	 *List of datapackages created*
	 - [ ] add

* [Linked Open Data](https://datahub.io/awesome/linked-open-data)

	 *List of datapackages created*
	 - [ ] add

* [Logistics](https://datahub.io/awesome/logistics-data)

	 *List of datapackages created*
	 - [ ] add

* [Machine Learning / Statistical](https://datahub.io/awesome/machine-learning-data)

	 *List of datapackages created*
	 - [ ] add

* [Movies and TV](https://datahub.io/awesome/movies-and-tv)

	 *List of datapackages created*
	 - [ ] add

* [Open Corporates](https://datahub.io/awesome/opencorporates)

	 *List of datapackages created*
	 - [ ] add

* [Property Prices](https://datahub.io/awesome/property-prices)

	 *List of datapackages created*
	 - [ ] add

* [Reference Data](https://datahub.io/awesome/reference-data)

	 *List of datapackages created*
	 - [ ] add

* [Stock Market Data](https://datahub.io/awesome/stock-market-data)

	 *List of datapackages created*
	 - [ ] add

* [War and Peace](https://datahub.io/awesome/war-and-peace)

	 *List of datapackages created*
	 - [ ] add

* [Wealth, Income and Inequality](https://datahub.io/awesome/wealth-income-and-inequality)

	 *List of datapackages created*
	 - [ ] add

[...]

Issue #1172, opened 2018-07-23
Use shared table schema


henrykironde:
Use a shared table schema in frictionless data json files:
https://discuss.okfn.org/t/shared-table-schema/6969

harshitbansal05:
@henrykironde I was thinking to work on it, but could not understand properly from the discussion in the link. Could you please elaborate?

henrykironde:
Hi, @harshitbansal05 I think we should not work on this for not. I will have to revise the issue or I will close it and open a new one later.

henrykironde:
Closing due to change in the frictionless table schema. 
We also have a pr that handles using frictionless data packages.

PR #1489, opened 2020-07-08
Add frictionless data packages support

### LINKED ISSUE #968: Add ability to use existing data packages
We use Frictionless data's data package standard. One of the advantages of this is that we should be able to use any existing packages directly. This would let us easily support a large number of new datasets. Functionality would include:

* Store a list of data package locations, probably in a single file
* For each location download the data package
* Automatically place the csv files in the correct locations
* Automatically place the the json file in the `scripts` directory and make any necessary modifications to it

Here are a bunch of existing data packages that this can be tested on and that we could quickly incorporate:

https://github.com/datasets

henrykironde:
Closes #1415 #1420 
fixes #968


henrykironde:
Thanks to all the contributors of #1420, #1415, #980. The contributions have been added to this PR.

References.  #981, #1172, #1404, #968

PR #1491, opened 2020-07-10
Kaggle support

DumbMachine:
#1465 had gotten stale, opened a new one.

PR #1492, opened 2020-07-13
[WIP]: Test kaggle

henrykironde:


henrykironde:
@DumbMachine use this for reference to look for your mistake in the current pr #1491. The tests are fine in this PR but will fail due to a bad variable name "l".

PR #1494, opened 2020-07-14
Use of functions in scripts to test added functions

ashishpriyadarshiCIC:


henrykironde:
This is great @ashishpriyadarshiCIC. We can leave this open and we go back to the individual Pr where we are adding this support and add if statements corresponding to that PR or engine. That is to say, if we are adding the json support, we add that if statement and functionally to the PR. Do not forget to look for a script to add.  This way, we can merge as we go.

henrykironde:
Function added to PR #1504, #1505, #1506, #1507

PR #1507, opened 2020-07-30
Test script json

ashishpriyadarshiCIC:


retriever/lib/engine_tools.py
```diff
@@ -101,37 +103,93 @@ def reset_retriever(scope="all", ask_permission=True):
...
+                        dict(i for i in dictionary.items()
+                             if i[0] in header_values)).values)
+            else:
+                rows.extend(
+                    [dict(i for i in dictionary.items() if i[0] in header_values)])
+
+        elif dictionary.get(row_key):
+            if normalize:
+                rows.extend(json_normalize(dictionary[row_key]).values)
+            else:
```
henrykironde, line 172: @ashishpriyadarshiCIC we forgot something, This should return two values not one 
`rows, header_values = walker(`
same applies to all the usages below

ashishpriyadarshiCIC:
@apoorvaeternity, Tests are passing but we have not added the new scripts in retriever recipies

ashishpriyadarshiCIC:
@apoorvaeternity @henrykironde Ready for review

version.txt
```diff
@@ -43,6 +44,7 @@ harvard_forest.json,2.0.0
 home_ranges.json,1.1.2
 intertidal_abund_me.py,1.5.3
 iris.json,1.3.1
+json_test.json,1.2.1
```
henrykironde, line 47: This script should start from 1.0.0. And since it is derevide from portal, and it is used for tests, we can move it to the `test/raw_data/scripts/`  directory

retriever/lib/engine_tools.py
```diff
@@ -101,37 +104,102 @@ def reset_retriever(scope="all", ask_permission=True):
...
+    raw_data = json.loads(file_out.read(), object_pairs_hook=OrderedDict)
+
+    raw_data, header_values = walker(raw_data,
+                                     row_key=row_key,
+                                     header_values=header_values,
+                                     rows=[],
+                                     normalize=False)
+
+    if isinstance(raw_data[0], dict):
+        raw_data = [list(row.values()) for row in raw_data]
```
apoorvaeternity, line 135: Why are we type casting `row.values()` to `list` here - isn't row.values() already a list?

retriever/lib/engine_tools.py
```diff
@@ -101,37 +104,102 @@ def reset_retriever(scope="all", ask_permission=True):
...
+    """
+    Extract rows of data from json datasets
+    """
+    #  Handles the simple case, where row_key and column_key are not required
+    if not (row_key or header_values):
+        if isinstance(dictionary, dict):
+            rows = pd.DataFrame([dictionary]).values
+            header_values = dictionary.keys()
+            return rows, header_values
+        elif isinstance(dictionary, list):
```
apoorvaeternity, line 156: if `dictionary` can also be of type list maybe we should use some other name in place of `dictionary` here as it can be confusing.

retriever/lib/engine_tools.py
```diff
@@ -101,37 +104,102 @@ def reset_retriever(scope="all", ask_permission=True):
...
+            rows = pd.DataFrame(dictionary, columns=header_values).values
+            header_values = [chr(i) for i in range(ord('a'), ord('a') + len(rows[0]))]
```
apoorvaeternity, line 158: I don't understand the code here.

ashishpriyadarshiCIC: That part of code is to create Header values if they don't exist with alphabets.

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: This case is used in case raw_data is a list of dictionaries so it will get every element as row and then retrieve it's values to convert into a list.
Example :
``` j
[
  {
    "User": "Alex",
    "id": "US1",
    "Age": "25",
    "kt": "2.0",
    "qt": "1.00"
  },
  {
    "User": "Tom",
    "id": "US2",
    "Age": "20",
    "kt": "0.0",
    "qt": "1.0"
  }
]
```

apoorvaeternity: No that's fine. I just meant why we are using `list(row.values())` instead of just using `row.values()` in the list comprehension.

retriever/lib/engine_tools.py
apoorvaeternity: something like `a , b, c, d` right? we should add a comment here mentioning that.
wondering if
```
from string import ascii_lowercase
num_columns = len(rows[0])
header_values = list(ascii_lowercase)[:num_columns]
```
would give a similar result.

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: Done

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: As raw_data should be a list of lists in the end

version.txt
ashishpriyadarshiCIC: Done

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: Added a comment for it

retriever/lib/engine.py
```diff
@@ -581,6 +582,13 @@ def excel_to_csv(self, src_path, path_to_csv, excel_info=None, encoding=ENCODING
         if self.find_file(src_path) and excel_info:
             excel_csv(src_path, path_to_csv, excel_info, encoding)
 
+    def process_json2csv(self, src_path, path_to_csv, x, encoding=ENCODING):
```
henrykironde, line 585: We need to rename the `x `parameter with a more suitable name

retriever/lib/engine_tools.py
ashishpriyadarshiCIC: @apoorvaeternity ascii_lowercase  works only for 26 columns and it will show error after that. Could you help me in making it work?
`header_values = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z']`
What we want 
`header_values = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 'aa', 'ab', 'ac', .....]`

apoorvaeternity: Well in that case you can create a generator using `iterator.product()`. Something like this should do it: https://stackoverflow.com/a/42176541

ashishpriyadarshiCIC: @apoorvaeternity Thanks for the suggestion, fixed that issue now

PR #1527, opened 2020-12-16
Update create CREATE EXTENSION postgis

henrykironde:
New versions of postgres use
CREATE EXTENSION postgis_raster

### DIFF
--- a/docs/spatial_dbms.rst
+++ b/docs/spatial_dbms.rst
@@ -1,12 +1,17 @@
-=======================
-Spatial database set up
-=======================
+======================
+Spatial database setup
+======================
 
 Supporting installation of spatial data into `Postgres DBMS`.
 
 1. **Install Postgres**
 
-  To install `Postgres` from scratch please ref to `Postgres installation`_.
+  For Mac the easiest way to get started with PostgreSQL is with `Postgres.app`_.
+
+  For Debain and Ubuntu, install `PostgresSQL and PostGis` please ref to `Postgres installation`_.
+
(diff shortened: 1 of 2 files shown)


docs/spatial_dbms.rst
```diff
@@ -6,7 +6,7 @@ Supporting installation of spatial data into `Postgres DBMS`.
 
 1. **Install Postgres**
 
-  To install `Postgres` from scratch please ref to `Postgres installation`_.
+  To install `Postgres` from scratch please ref to `Postgres installation`_ or use a much simple installation method with `Postgres.app`_.
```
ethanwhite, line 9: Postgres.app is just for macos and it looks like the old installation link was just for Linux. Should we include a recommended link for each OS for installation?

docs/spatial_dbms.rst
```diff
@@ -30,21 +30,24 @@ Supporting installation of spatial data into `Postgres DBMS`.
   This is done by using either `Postgres CLI` or `GUI(PgAdmin)` and run
   **For psql CLI**
 
-.. code-block:: sh
+.. note:: New versions of Postgres use `CREATE EXTENSION postgis_raster;`
```
ethanwhite, line 33: I'd recommend separate instructions by version to make this clear. So one for after the change and one for before the change. Alternatively I think it would be OK to have a minimum postgres version that we support that is one after this change was made.

ethanwhite:
Looks like the failed test was just a download hiccup. This looks great! Merging.

PR #1531, opened 2021-01-26
Fixed kaggle function and modified error messages

### LINKED ISSUE #1529: Download from Kaggle function needs to be updated
The function [download_from_kaggle](https://github.com/weecology/retriever/blob/fe5148b66a44fd94ba714a5cf8f9cda4f35e3249/retriever/lib/engine.py#L512) uses the Kaggle API to download datasets. This functions need to be updated to work well.

Kadam-Tushar:
Fixes #1529

- Corrected the way of checking `kaggle` field from scripts
- Enhanced the error messages when the user is unable to download data from Kaggle (datasets + competitions) 
- Refactored function `download_from_kaggle`

### DIFF
--- a/retriever/lib/engine.py
+++ b/retriever/lib/engine.py
@@ -533,34 +533,30 @@ class Engine():
 
         api = KaggleApi()
         api.authenticate()
+        archive_full_path = archive_full_path + ".zip"
 
         if data_source == "dataset":
-            archive_full_path = archive_full_path + ".zip"
             try:
                 api.dataset_download_files(dataset=dataset_name,
                                            path=archive_dir,
                                            quiet=False,
                                            force=True)
                 file_names = self.extract_zip(archive_full_path, archive_dir)
-            except ApiException:
-                print(f"The dataset '{dataset_name}' isn't currently available "
-                      f"in the Retriever.\nRun 'retriever ls' to see a "
-                      f"list of currently available datasets.")
+            except ApiException as e:
+                print(f"Could not download '{dataset_name}' from kaggle datasets.")
+                print("Response :", e.body.decode("unicode_escape"))
                 return []
 
         else:
-            archive_full_path = archive_full_path.replace("kaggle:competition:",
-                                                          "") + ".zip"
             try:
                 api.competition_download_files(competition=dataset_name,
                                                path=archive_dir,
                                                quiet=False,
                                                force=True)
                 file_names = self.extract_zip(archive_full_path, archive_dir)
-            except ApiException:
-                print(f"The dataset '{dataset_name}' isn't currently available "
-                      f"in the Retriever.\nRun 'retriever ls' to see a "
-                      f"list of currently available datasets.")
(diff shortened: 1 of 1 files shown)


henrykironde:
That is a new error, let me check try to find out why. At the moment, I have restarted the build

retriever/lib/engine.py
```diff
@@ -533,34 +533,30 @@ def download_from_kaggle(
...
-                print(f"The dataset '{dataset_name}' isn't currently available "
-                      f"in the Retriever.\nRun 'retriever ls' to see a "
-                      f"list of currently available datasets.")
+            except ApiException as e:
+                print(f"Could not download '{dataset_name}' from kaggle datasets.")
+                print("Response :", e.body.decode("unicode_escape"))
                 return []
 
         else:
-            archive_full_path = archive_full_path.replace("kaggle:competition:",
```
henrykironde, line 552: So we are saying and confirming that the path for datasets and competition are the same, correct?

henrykironde:
@Kadam-Tushar thanks for contributing this PR. Could we also add some comments on the commands that you used to test this.

Kadam-Tushar:
@henrykironde  To test the `download_from_kaggle` function we can do:
1.  Get some Kaggle scripts for e.g one can use [these](https://github.com/weecology/retriever-recipes/pull/88/files) script.
2.  Put that script in ` ~/.retriever/scripts`  this folder.
3.  One can see those added scripts after running `retriever ls`
4.  To download and install those scripts use the command `retriever install {database_name} {script_name} `
E.g ` retriever install csv coronavirus-italy`
It will download and install dataset in csv format to the current directory

retriever/lib/engine.py
Kadam-Tushar: Yes, for kaggle datasets URL should have  format of `author/dataset`, and for competitions `name-of-competition`.

retriever/lib/engine.py
```diff
@@ -533,34 +533,30 @@ def download_from_kaggle(
...
                                                quiet=False,
                                                force=True)
                 file_names = self.extract_zip(archive_full_path, archive_dir)
-            except ApiException:
-                print(f"The dataset '{dataset_name}' isn't currently available "
-                      f"in the Retriever.\nRun 'retriever ls' to see a "
-                      f"list of currently available datasets.")
+            except ApiException as e:
+                print(f"Could not download '{dataset_name}' from kaggle competitions.")
+                print("Response :", e.body.decode("unicode_escape"))
```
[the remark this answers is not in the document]
Kadam-Tushar: @henrykironde:  To improve the readability of error message
From `ApiException`  Kaggle API returns only byte string (i.e only ASCII characters) so all other Unicode characters will be shown with escape characters. For e.g, if I don't decode these Unicode characters, we will get a response like this.

before

After decoding we can see nice output like this : 

after

henrykironde:
Looks good to me. Thanks @Kadam-Tushar

Issue #1557, opened 2021-03-12
Link Broken of the 'Read The Docs'.


Mohitkumar6122:
Link of the **Read the Docs** section under **Running tests locally**  section in the [https://retriever.readthedocs.io/en/latest/developer.html#setting-up-servers](https://retriever.readthedocs.io/en/latest/developer.html#setting-up-servers) is broken.
<br>
Here is the broken link **MARKED IN RED**. <br>
![Image](https://i.ibb.co/D1J9yhb/Screenshot-2021-03-13-033600.png). 
<br> After visiting link<br>  
![Uploading image.png…](https://i.ibb.co/WWWsj02/Screenshot-2021-03-13-022353.png)

Kindly fix it .
Thanks.

henrykironde:
This is kind of strange, the page loads fine on my end. Anyone else can’t see the docs?

ethanwhite:
Opens fine for me. Maybe just a brief outage?

Mohitkumar6122:
> This is kind of strange, the page loads fine on my end. Anyone else can’t see the docs?

It is still broken for my side. Can you check one more time .

henrykironde:
Can you try to open the link in a private browser

Sohan022:
No, it did not fix. Can I fix this issue?

henrykironde:
Sure

henrykironde:
@Mohitkumar6122 that you for reporting this. Let me know incase it is not solved

PR #1559, opened 2021-03-13
Link broken in docs fixed

Sohan022:
I have fixed the issue 'Link Broken of the Read The Docs'. 
fixed: #1557


henrykironde:
Thanks @Sohan022 for catching that

PR #1577, opened 2021-04-05
Fixes #1572

### LINKED ISSUE #1572: Clean up CLI to use the python interface
In the python interface, we use the function [def datasets(keywords=None, licenses=None):](https://github.com/weecology/retriever/blob/bb06180a030d34eafa2e1ea13b74a3719df827e1/retriever/lib/datasets.py#L5).
This should be able to return what we are doing in the main (CLI interface) at [line](https://github.com/weecology/retriever/blob/main/retriever/__main__.py#L105).
 and the results can be printed in the away that looks good in CLI(terminal)

_Originally posted by @henrykironde in https://github.com/weecology/retriever/issues/1570#issuecomment-810558636_

Aakash3101:
Changed arguments after -v flag in retriever ls command to be necessary
Also added argument completion. Changes in get_opts.py

### DIFF
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -31,9 +31,10 @@ script_list.extend(get_dataset_names_upstream(repo=RETRIEVER_REPOSITORY))
 script_list.extend(get_dataset_names_upstream())
 script_list = sorted(set(script_list))
 
-# set of all possible licenses and keywords
+# set of all possible licenses, keywords and scripts
 licenses_options = set(licenses_list)
 keywords_options = set(keywords_list)
+scripts_options = script_list
 
 parser = argparse.ArgumentParser(prog="retriever")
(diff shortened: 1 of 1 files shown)


retriever/lib/get_opts.py
```diff
@@ -31,9 +31,10 @@
 script_list.extend(get_dataset_names_upstream())
 script_list = sorted(set(script_list))
 
-# set of all possible licenses and keywords
+# set of all possible licenses, keywords and scripts
 licenses_options = set(licenses_list)
 keywords_options = set(keywords_list)
+scripts_options = set(script_list)
```
henrykironde, line 37: keep these sorted sorted(set(script_list))

Aakash3101: Okay @henrykironde

henrykironde:
This looks good to me. Thank you @Aakash3101 👍

PR #1580, opened 2021-04-08
Fix Docs typo-'Postgres' to 'MySqL'

mohamedsto7y:
I think it was intended to type MySQL referring to context.


henrykironde:
Thanks @mohamedsto7y for catching this

PR #1594, opened 2021-06-14
Remove nuclear power plants

Aakash3101:
Nuclear power plants is always changing. (unstable)
Co-authored-by: Kush Kothari <kushkothari2001@gmail.com>


henrykironde:
Thanks @Aakash3101 and @kkothari2001

PR #1597, opened 2021-06-14
Feature to check for already existing data packages

### LINKED ISSUE #1576: Improve script creation platform to detect already existing data packages before we duplicate them
**Easy:**

Before a user/developer spends time on the data, we should be able to tell if we already have that data in the retriever. 
Once the user gets the exact url of the data source. We should run a command to check all the urls in the scripts instance if the url exists.

**How to solve:**
User has a url of a data source, i,e. `new-data-source-url`
Use already available functions like 
[def SCRIPT_LIST(): ](https://github.com/weecology/retriever/blob/bb06180a030d34eafa2e1ea13b74a3719df827e1/retriever/lib/scripts.py#L108)
```
 for script in SCRIPT_LIST()
      check if the script resource url equals new-data-source-url
      if so, register any match
tell the user that we have found some matches in script name
or data not in the retriever yet
```

Sample usage should be something like
`retriever autocreate --check "https://archive.ics.uci.edu/ml/machine-learning-databases/iris/bezdekIris.data" `
Data package Iris exist already

Note: Use [get opts](https://github.com/weecology/retriever/blob/main/retriever/lib/get_opts.py) to add the `--check`  for `retriever autocreate --check`

dassaniansh:
Fixes #1576

### DIFF
--- a/retriever/lib/get_opts.py
+++ b/retriever/lib/get_opts.py
@@ -143,6 +143,7 @@ autocreate_parser.add_argument('-o',
                                help='write scripts out to a designated directory',
                                nargs='?',
                                const='')
+autocreate_parser.add_argument('-c', help='turn files into scripts', action='store_true')
 autocreate_parser.add_argument('--skip-lines',
                                help='skip a set number of lines before processing data',
                                nargs=1,
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ apidocs
 scripts/*.pyc
 .idea*
 *_pycache_
+.vscode/
--- a/retriever/__main__.py
+++ b/retriever/__main__.py
@@ -86,7 +86,6 @@ def main():
             f = open(args.filename, 'w')
             f.write(sample_script)
             f.close()
-
             return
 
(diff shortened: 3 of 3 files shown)


retriever/lib/get_opts.py
```diff
@@ -143,6 +143,7 @@
                                help='write scripts out to a designated directory',
                                nargs='?',
                                const='')
+autocreate_parser.add_argument('-c', help='helps to check for existing data packages', action='store_true')
```
henrykironde, line 146: Thanks @dassaniansh. The tests still fail and it looks like this line is not well formatted. (Too long)
change from 
`helps to check for existing data packages`
to 
`check if data package already exists` 

Make sure the line length is no more than 90

dassaniansh: @henrykironde  sir, Is it fine now or I should update anything else

dassaniansh:
@henrykironde sir, Is it fine no?

henrykironde:
Thanks @dassaniansh for the the PR

PR #1601, opened 2021-07-03
Updated docker-compose.yml and Dockerfile

Aakash3101:


Aakash3101:
So branched from the `socrata` branch to check if there are any cases where the changes in Dockerfile fail. Since all the tests have passed for the `socrata` branch, I will remove those commits and push the dockerfile commit only.

henrykironde:
@Aakash3101 thanks for getting the CI working again.

PR #1604, opened 2021-07-09
Updated Dockerfile

dassaniansh:


Aakash3101:
@dassaniansh This dockerfile is abandoned and not used in any workflow.

dassaniansh:
> @dassaniansh This dockerfile is abandoned and not used in any workflow.

Yeah I merged both dockerfile's into one, Is it fine now?

Aakash3101:
You don't need to make any changes in this Dockerfile, if look at the repository we use the Dockerfile which is present at the root of the project. In the new workflow, I am not using any Dockerfile and docker-compose.yml. So making any changes in them is not going to change the workflow.

henrykironde:
Thanks for the changes @dassaniansh.
The changes in docker/Dockerfile  will enable us automatically register the docker file to the docker registry.
This will make it possible for users to pull the docker image of retriever from docker hub using `docker pull weecology/retriever`

PR #1606, opened 2021-07-16
Added tidycensus dataset

Aakash3101:
Python script for installing the datasets present in the R package `tidycensus` using rpy2.

Some other changes that I made after yesterday's meeting, are to check if the raw data files are present or not. If they are present use those files and install them.

### DIFF
--- a/scripts/tidycensus.py
+++ b/scripts/tidycensus.py
@@ -1,0 +1,207 @@
+# -*- coding: utf-8 -*-
+#retriever
+
+import os
+import re
+import subprocess
+import pandas as pd
+
+try:
+    import rpy2.robjects as ro
+    import rpy2.robjects.packages as rpackages
+    from rpy2.robjects import pandas2ri
+    from rpy2.robjects.conversion import localconverter
+except ImportError:
+    pass
+
+from retriever.lib.models import Table, VectorDataset
+from retriever.lib.templates import Script
+from retriever.lib.defaults import DATA_WRITE_PATH
+
+R_string = "Rscript -e \"install.packages('{package_name}')\""
+exit_status = "had non-zero exit status"
+unable_to_access = "unable to access index for repository"
+packages = ['rgdal', 'units', 'sf', 'tigris', 'tidycensus']
+file_paths = {}
+
+
+def census_api_key():
+    """Checks if the CENSUS_API_KEY is added to the .Renviron file"""
+    r_string = '''
+        home <- Sys.getenv("HOME")
+        renv <- file.path(home, ".Renviron")
+        readRenviron(renv)
+        key <- Sys.getenv("CENSUS_API_KEY")
+        '''
+
+    ro.r(r_string)
+    ro.r['options'](warn=-1)  # suppresses warnings from R console
+    api_key = ro.r('key')[0]
+    if not api_key:
+        print('Add the CENSUS_API_KEY in your .Renviron file')
+        return False
+    else:
+        return True
+
+
+def install_tidycensus():
+    """Installs the R packages required for installing the tidycensus package"""
+    for package in packages:
+        if not rpackages.isinstalled(package):
+            output = subprocess.getoutput(R_string.format(package_name=package))
+            if bool(re.search(exit_status, output)):
+                print(output)
+                print(
+                    "*Please install the system dependencies for {} package*".format(
+                        package))
+                print("Installating R packages failed!!!")
+                return False
+            elif bool(re.search(unable_to_access, output)):
+                print(output)
+                return False
+        print("Package {} is installed".format(package))
+    print("Installing R packages completed!!!")
+    return True
+
+
+class main(Script):
+
+    def __init__(self, **kwargs):
+        Script.__init__(self, **kwargs)
+        self.title = "Load US Census Boundary and Attribute Data as 'tidyverse' and " \
+            "'sf'-Ready Data Frames"
+        self.name = "tidycensus"
+        self.retriever_minimum_version = '3.0.1-dev'
+        self.version = '1.0.0'
+        self.ref = "https://github.com/walkerke/tidycensus"
+        self.citation = ""
+        self.licenses = [{"name": "MIT"}]
+        self.description = "An integrated R interface to the decennial US Census and American Community Survey APIs and" \
+                           "the US Census Bureau's geographic boundary files. Allows R users to return Census and ACS data as" \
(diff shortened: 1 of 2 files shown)


Aakash3101:
Okay, so the tests are failing because when the scripts are read, it reads the import lines, and it could not find the `rpy2` package. So that is the reason why the tests fail. And I have tested that only the import lines are read, the script is actually not executed so we can put the import statements in a try-except block like:
``` python
try:
    import rpy2.robjects as ro
    import rpy2.robjects.packages as rpackages
    from rpy2.robjects import pandas2ri
    from rpy2.robjects.conversion import localconverter
except ImportError:
    pass
```

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+        self.citation = ""
+        self.licenses = [{"name": "MIT"}]
+        self.description = "An integrated R interface to the decennial US Census and American Community Survey APIs and" \
+                           "the US Census Bureau's geographic boundary files. Allows R users to return Census and ACS data as" \
+                           "tidyverse-ready data frames, and optionally returns a listcolumn with feature geometry for all Census" \
+                           "geographies."
+        self.keywords = ["census", "US", "tidycensus", "R", "geographic boundaries"]
+        self.encoding = 'utf-8'
+
+    def census_api_key(api_key):
```
henrykironde, line 46: Add a doc string

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+        library(tidycensus)
+        census_api_key("{}", install = TRUE)
+        home <- Sys.getenv("HOME")
+        renv <- file.path(home, ".Renviron")
+        readRenviron(renv)
+        '''.format(api_key)
+
+        ro.r(r_string)
+
+    def install_tidycensus():
```
henrykironde, line 57: Add a doc string

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+                    print(output)
+                    print(
+                        "*Please install the system dependencies for {} package*".format(
+                            package))
+                    return False
+
+                else:
+                    print("Package {} is installed".format(package))
+            else:
+                print("Package {} is installed".format(package))
```
henrykironde, line 71: These two print statements seem like we could just remove them one.

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+
+        ro.r(r_string)
+
+    def install_tidycensus():
+        for package in packages:
+            if not rpackages.isinstalled(package):
+                output = subprocess.getoutput(R_string.format(package_name=package))
+                if bool(re.search(pattern, output)):
+                    print(output)
+                    print(
```
henrykironde, line 63: Could you comment a sample run when the package is 
1) Not available so it is installed
2) Run it with out an internet connection am curious to see what we get

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+                            package))
+                    return False
+
+                else:
+                    print("Package {} is installed".format(package))
+            else:
+                print("Package {} is installed".format(package))
+        return True
+
+    def download_raw_data(self, datasets):
```
henrykironde, line 74: Add a doc string, and may be a comment on the if statements to explain why we are using the if statement.

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+            'fips_codes.csv', 'mig_recodes.csv', 'pums_variables.csv',
+            'state_laea/state_laea.shp', 'county_laea/county_laea.shp'
+        ]
+        for filepath in filepaths:
+            if os.path.isfile(os.path.join(data_path, filepath)):
+                filename, file_ext = os.path.splitext(os.path.basename(filepath))
+                file_paths[filename] = os.path.join(data_path, filepath)
+                download = False
+
+        if (key[0]
```
henrykironde, line 147: can be `if not key[0] and ...`
This part `[x for x in packages if rpackages.isinstalled(x)` looks like it checking if its installed the package, and then we call "self.install_tidycensus()" which also checks the same. 

Could we organize this in a way that we first check and make sure all libraries are installed, and the api keys is present, then we do not have to check several times.

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+                == '') and (len([x for x in packages if rpackages.isinstalled(x)]) > 0):
+            result = self.install_tidycensus()
+            print("Installing tidycensus...")
+            if result:
+                print("Installation completed!!!")
+                print(
+                    '*The API key provided to you from the Census formated in quotes. A key can be acquired at http://api.census.gov/data/key_signup.html*'
```
henrykironde, line 154: Avoid very long lines, you have some tools that you could use to do some refactoring like yapf. But be care full of applying it to scripts. Especially the the very first lines should not be changed.
```python
# -*- coding: utf-8 -*-
#retriever
```

scripts/tidycensus.py
```diff
@@ -0,0 +1,220 @@
...
+                exit()
+            self.download_raw_data(datasets)
+
+        elif (key[0] == '') and (len([x for x in packages if rpackages.isinstalled(x)])
+                                 == 0):
+            print(
+                '*The API key provided to you from the Census formated in quotes. A key can be acquired at http://api.census.gov/data/key_signup.html*'
+            )
+            print("Enter your census api key here : ", end="")
+            api_key = input()
```
henrykironde, line 170: We should not ask people to enter any information, they should use config files for keys

henrykironde (commented):
Over all, seems good. I have made some comments.
Please include some example commands that you have run and screenshots  where applicable.

henrykironde:
@Aakash3101, let us work on the clean up together, I have got some new designs we can follow.

Aakash3101:
Sure

Aakash3101:
Installing tidycensus without internet:

Screenshot from 2021-08-10 23-55-46

---

Installing tidycensus with internet:

Screenshot from 2021-08-10 23-58-54

Aakash3101:
@henrykironde could you share your changes in the `install_tidycensus` function

Aakash3101:
@henrykironde I am unable to make changes in the script for it to follow the `retriever download tidycensus --path <PATH>` command. The raw data is getting saved in the `~/.retriever/raw_data/tidycensus` directory and not at the specified `PATH`.

I think @kkothari2001 would be able to help me here, and he would also achieve his task from this.

PR #1607, opened 2021-07-18
Clip raster files for installation

dassaniansh:
Use extent to clip raster files
Converting bil files to tif files and clip them


retriever/engines/postgres.py
```diff
@@ -1,7 +1,7 @@
 import os
 import sys
 import subprocess
-
+from osgeo import gdal
```
Aakash3101, line 4: @dassaniansh, so your tests are failing because `osgeo` is not installed, if you haven't written any tests for the new changes, put this import in a try-except block to avoid the failing tests for the time being.

Or you could add the `osgeo` module to the `python-package.yml`

henrykironde: For now we can add a try block, but later we need to install gdal and make sure we set the paths for both `python-package.yml` (Linux) and appveyor.yml (Windows)

retriever/engines/postgres.py
```diff
@@ -166,6 +166,53 @@ def insert_raster(self, path=None, srid=4326):
...
+                yMin = self.opts["bbox"][1]
+                xMax = self.opts["bbox"][2]
+                yMax = self.opts["bbox"][3]
+            else:
+                xMin = self.script.tables[self.table.name].extent["xMin"]
+                yMin = self.script.tables[self.table.name].extent["yMin"]
+                xMax = self.script.tables[self.table.name].extent["xMax"]
+                yMax = self.script.tables[self.table.name].extent["yMax"]
+
+            xMin = 0 if xMin < 0 else xMin
```
henrykironde, line 182: @dassaniansh I will need some explanation  on this function from Line 182  to 200

henrykironde (commented):
Most of the work seems good, however, we made some mistakes in clipping the data. We can discuss this when we meet. 
We shall need to create new test data for this PR. ( I think I have some ready)

**Steps to test:**
I am also creating a replica of Bioclim with only the bio1 table( test-bio-clip).  We shall run the script to test this and render it in Qgis for validation. We shall go through it together. 

- The first run should be without any extent, ie, `retriever install postgres test-bio-clip`
- The second run should include the extent, in this case, am using 75, 65, 125, 45
- The third run should include vague extents like 5005, 00065, -00125, 45000

After performing a manual check on a clipped version,  we should be getting something as shown below. The green part is the clip image and it maps right onto the bigger image.

henrykironde:
Test cases:
biofilm-bio1 no extent
biofilm-bio1-extent( extent in resource)
biofilm-bio1-extent2( extent in global property)

biofilm-bio1-crop-extent( extent in resource)  provide custom extent
biofilm-bio1-crop-extent2( extent in global property) provide custom extent

henrykironde:
Closing in favor of #1614

PR #1610, opened 2021-07-22
Testgdal2

henrykironde:


test/raw_data_gis/scripts/test_raster_bio1clip.json
```diff
@@ -0,0 +1,49 @@
...
+            "other_paths": "",
+            "path": "bio1clip.tif",
+            "scale": null,
+            "statistics": {
+                "maximum": 0.0,
+                "mean": 0.0,
+                "minimum": 0.0,
+                "stddev": -1.0
+            },
+            "url": "https://github.com/henrykironde/retriever/blob/testgdal2/test/raw_data_gis/test_raster_bio1clip.zip"
```
Aakash3101, line 34: I think this url is at fault here, if you try to make a get request at this url, it returns the html code of the webpage. The correct url would be : https://github.com/henrykironde/retriever/raw/testgdal2/test/raw_data_gis/test_raster_bio1clip.zip

test/raw_data_gis/scripts/test_raster_bio1clip.json
```diff
@@ -0,0 +1,49 @@
...
+            "band_name": "bio1clip_1",
+            "color_table": null,
+            "extensions": [
+                "tif"
+            ],
+            "format": "raster",
+            "name": "bio1clip",
+            "no_data_value": -9999.0,
+            "other_paths": "",
+            "path": "bio1clip.tif",
```
Aakash3101, line 26: This might be what causes the error in finding the file. So I did some commands using `unzip`, to see how files are stored in the archive. And I compared the `test_raster_bio1clip.zip` to `test_raster_bio1.zip` (which is present in the current version of retriever).

terminal

In the image, you can clearly see that the `tif` file is present in the folder `test_raster_bio1clip` in the `test_raster_bio1clip.zip`, while for the `test_raster_bio1.zip`, the file is present in the root of the archive.

So either you change the archive, or change the path in the script

Aakash3101: @henrykironde

PR #1621, opened 2021-09-27
Created docker-publish.yml

dassaniansh:


docker/Dockerfile
```diff
@@ -1,25 +1,26 @@
...
 
 MAINTAINER Weecology "https://github.com/weecology/retriever"
 
 # Installing Gdal and configuring it
-RUN add-apt-repository ppa:ubuntugis/ppa &&\
-    apt-get update && apt-get install -y --no-install-recommends apt-utils &&\
+RUN export CPLUS_INCLUDE_PATH=/usr/include/gdal
+RUN export C_INCLUDE_PATH=/usr/include/gdal
+    
+RUN add-apt-repository ppa:ubuntugis/ubuntugis-unstable
```
Aakash3101, line 10: The log for the docker action tells that `add-apt-repository: not found`. You should try again by putting this line after `apt-get update` command. May be that is why it is not found as `apt-utils` are not present before this line.

PR #1622, opened 2021-09-27
Added try block for GDAL

dassaniansh:


henrykironde:
Thanks at @dassaniansh

retriever/engines/postgres.py
```diff
@@ -1,7 +1,11 @@
 import os
 import sys
 import subprocess
-from osgeo import gdal
+try:
+    from osgeo import gdal
+except:
+    gdal = None
+    print("Install GDAL")
```
Aakash3101, line 8: @dassaniansh This will constantly print `Install GDAL` every time you use retriever if GDAL is not installed. It would be better if you put this print statement in the function, so that when the user tries to run the function without the dependencies installed, they would be alerted about the missing dependencies (GDAL).

Here is an example I tested,
```
$ retriever ls -v airports                                                                                                                                               
Install GDAL
1. Dataset containing information on all airports on ouraiports.com
Dataset Name: airports
Keywords: ['airports', 'aeroplanes', 'airplanes', 'frequencies', 'runways', 'navaids', 'countries', 'regions']
Description: Dataset containing information on all airports on ourairports.com
Licenses: None
Citation: OurAirports.com, Megginson Technologies Ltd.
```

henrykironde: @Aakash3101 nice catch 👍🏿 we shall redesign this part as you have recommended.

PR #1629, opened 2021-11-01
Added kaggle to setup.py

kkothari2001:
Kaggle wasn't in the setup.py which meant the user had to install it with the pip command.

This PR will directly install kaggle when retriever is installed.


henrykironde:
Thank you @kkothari2001

PR #1631, opened 2021-11-19
Updated the docs

Aakash3101:
Removed functions from ``__init__.py`` file , (``create_socrata_dataset``, ``update_socrata_contents``, ``update_socrata_script``)
and (``create_rdataset``, ``update_rdataset_contents``, ``update_rdataset_script``). These were irrelevant socrata and rdataset functions in the python interface for the users.

Updated ``test_retriever.py`` to not use python interface for these functions.

Removed these functions from docs as well.


henrykironde:
Thanks @Aakash3101 👍🏿

PR #1635, opened 2022-01-21
Use this token when uploading cov reports

henrykironde:


henrykironde:
[codecov](https://codecov.io/gh/weecology/retriever)

PR #1649, opened 2022-04-26
Clean up tests

henrykironde:


henrykironde:
Test fail because of reduced test coverage, in this case it is okey

PR #1664, opened 2023-01-30
debug display_all_rdataset_names

bloemenk:
Bug of wrong result whit a single string as argument was solved by the adding an extra elif option in the method. This elif checks if the argument of the function is of the type 'string'

### DIFF
--- a/retriever/lib/rdatasets.py
+++ b/retriever/lib/rdatasets.py
@@ -165,6 +165,17 @@ def display_all_rdataset_names(package_name=None):
         print("List of all the packages present in Rdatasets\n")
         packages = [(package, True) for package in rdatasets.keys()]
         lscolumns.printls(packages)
+    
+    elif isinstance(package_name,str):
+        try:
+            dataset_names = rdatasets[package_name].keys()
+            for dataset in dataset_names:
+                script_name = f"rdataset-{package_name}-{dataset}"
+                print(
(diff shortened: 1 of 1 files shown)


bloemenk:
Hi, I would really like to succeed this pull request. Therefor I would like to ask you some advise on how to solve the value error that appears during the build. I am not sure if it has a correlation with the code I worked on.

henrykironde:
Can you add some previous output and current output after the patch.

Nageshbansal:
Hey @bloemenk, let us know if you're still working on this patch, as there is already PR #1667 for this same bug.

henrykironde:
@pyther-hub can you let us know why your solution differs from this initial suggested solution here? @bloemenk, could you also look at @pyther-hub's solution and let us know what you think?

We shall combine and merge both solutions for authorship purposes if validated by the issue creator, @Nageshbansal.

pyther-hub:
@henrykironde much difference is not there I have converted the string value into a list with one element and @bloemenk has removed the for loop if str is the input and copied the code for each element which was written under the for loop 
[the part](https://github.com/weecology/retriever/pull/1664/commits/bae395b6d41404e94970951cb9556eddb84546f0#diff-9fec9131882c37a1c23c4cb5c6891aa36c8a977c2ee0ef62cbc7b9810d08cfe6R169-R178)

PR #1669, opened 2023-04-09
Test on Python 3.11

ethanwhite:


ethanwhite:
The failure here is just code cov

Issue #1671, opened 2023-10-30
Installation from source fails due to missing configuration


dikwickley:
Python version: 3.11.6
When building from source by following steps mentioned in README (https://github.com/weecology/retriever#to-install-from-source)
```
// after cloning and running
pip install . -U
```
Trying to run `retriever ls` gives the following error
```
(env) aniket@air retriever % retriever ls    
Traceback (most recent call last):
  File "/Users/aniket/Projects/gsoc/retriever/env/bin/retriever", line 5, in <module>
    from retriever.__main__ import main
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/__init__.py", line 7, in <module>
    from retriever.lib.engine_tools import set_proxy, create_home_dir
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/__init__.py", line 4, in <module>
    from .datasets import datasets
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/datasets.py", line 1, in <module>
    from retriever.lib.scripts import SCRIPT_LIST, get_script, get_dataset_names_upstream
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/scripts.py", line 506, in <module>
    global_script_list = StoredScripts()
                         ^^^^^^^^^^^^^^^
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/scripts.py", line 495, in __init__
    self._shared_scripts = SCRIPT_LIST()
                           ^^^^^^^^^^^^^
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/scripts.py", line 115, in SCRIPT_LIST
    return reload_scripts()
           ^^^^^^^^^^^^^^^^
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/scripts.py", line 59, in reload_scripts
    if not check_retriever_minimum_version(read_script):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aniket/Projects/gsoc/retriever/env/lib/python3.11/site-packages/retriever/lib/scripts.py", line 32, in check_retriever_minimum_version
[...]

PR #1674, opened 2023-11-06
Added check for empty version string

dikwickley:
fixes: #1671 

Just a simple check so that the version regex of setuptools > 66 is not empty.

@ethanwhite

### DIFF
--- a/retriever/lib/scripts.py
+++ b/retriever/lib/scripts.py
@@ -29,7 +29,7 @@ def check_retriever_minimum_version(module):
     m = module.name
 
     if hasattr(module, "retriever_minimum_version"):
-        if not parse_version(VERSION) >= parse_version("{}".format(mod_ver)):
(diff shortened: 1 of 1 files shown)


ethanwhite (approved):
Thanks @dikwickley! LGTM!

Will leave this for @henrykironde to merge so he can also have a quick look.

retriever/lib/scripts.py
```diff
@@ -29,7 +29,7 @@ def check_retriever_minimum_version(module):
     m = module.name
 
     if hasattr(module, "retriever_minimum_version"):
-        if not parse_version(VERSION) >= parse_version("{}".format(mod_ver)):
+        if mod_ver != "" and not parse_version(VERSION) >= parse_version("{}".format(mod_ver)):
```
henrykironde, line 32: if mod_ver.strip() and not

henrykironde (changes requested):

PR #1686, opened 2025-04-06
Added automatic index creation to SqLite engine

Khush2040:
This PR enhances the SQLite engine by adding automatic index creation after data is inserted into the `species` table. This is done in the `insert_data_from_file` function to improve query performance.


Khush2040:
Hi maintainers, I've added automatic index creation to the SQLite engine as discussed. Please let me know if any changes are needed!

henrykironde:
@Khush2040 Thank you for your contribution! This is a good idea in principle, but since each table is different, creating an index for idx_species_name won’t be applicable in a general-purpose context. One suggestion would be to start by investigating why the current setup is failing. I also recommend removing the AppVeyor builds and relying solely on GitHub Actions for CI. Once builds are back and running we can create indexed tables.

PR #1697, opened 2026-03-21
Add Python 3.12 compatibility test

Pritika2512:
This pull request adds a small compatibility test to verify that the project runs on supported Python versions, including Python 3.12.

The test asserts that the runtime Python version is >= 3.8, which matches the supported versions defined by the project.

This helps maintain compatibility with newer Python releases and ensures that future updates do not unintentionally introduce version-related issues.

### DIFF
--- a/test/test_python312_compatibility.py
+++ b/test/test_python312_compatibility.py
@@ -1,0 +1,7 @@
+import sys
+
+def test_python_version_compatibility():
+    """
+    Ensure the project runs on supported Python versions.
+    """
+    assert sys.version_info >= (3, 8)


henrykironde:
Thanks @Pritika2512! Compatibility tests are best implemented through CI platforms like GitHub Actions. You can refer to a good example here: https://github.com/weecology/DeepForest/blob/main/.github/workflows/ci_tests.yml

Also, this isn’t the correct project for your GSoC contribution, read the idea lists fully. Please take a closer look at the actual source repository: https://github.com/weecology/DeepForest/

